PR #1, opened 2019-06-14
[M1058] Adding circleci config 

gridcell:


gridcell:
[Setup circle config with new API repo](https://trello.com/c/vbubkrDY/1058-setup-circle-config-with-new-api-repo)

PR #3, opened 2019-06-27
[M1065] Cast parser decimal values to string

gridcell:


gridcell:
[SampleEvent uniqueness discrepancy](https://trello.com/c/lbqNsR7U/1065-sampleevent-uniqueness-discrepancy)

src/api/submission/parser.py
```diff
@@ -14,6 +14,12 @@
     "get_site_id",
 )
 
+def _cast_decimal_to_str(val):
+    if val is None:
+        return
```
saanobhaai: Should this be `return None`?
Should we catch decimal conversion exceptions?

gridcell: It does return None, but I've changed it to explicitly return `None`.  I'm not sure if there's anything to capture since it's converting to string.  Also, it would be captured further upstream in a try/except

PR #11, opened 2019-07-29
[M1044] Split datetime into year, month and day in field reports

gridcell:


gridcell:
[For all Export CSVs - separate 'date' into separate day, month and year columns](https://trello.com/c/qD8ABOVp/1044-for-all-export-csvs-separate-date-into-separate-day-month-and-year-columns)

PR #16, opened 2019-09-11
[M1133] remove exposure value from sites field report

gridcell:


gridcell:
[In Collect > Sites > Export site list, 'Reef exposure value' incorrectly exported. Remove this column.](https://trello.com/c/fZVN5O49/1133-in-collect-sites-export-site-list-reef-exposure-value-incorrectly-exported-remove-this-column)

PR #17, opened 2019-09-12
[M1122] Ensure sample events and sample units don't get orphaned

saanobhaai:


saanobhaai:
[Can't delete MR referenced by records, but those records don't exist](https://trello.com/c/Otoca35R/1122-cant-delete-mr-referenced-by-records-but-those-records-dont-exist)

PR #19, opened 2019-10-01
handle special characters in search filters

saanobhaai:


saanobhaai:
[DataError](https://trello.com/c/OIaWcnOd/1157-dataerror)

PR #20, opened 2019-10-01
add habitat complexity score value to export

saanobhaai:


saanobhaai:
[habitat complexity export CSV - change what values are exported in flat CSV](https://trello.com/c/qnv8Yjia/1148-habitat-complexity-export-csv-change-what-values-are-exported-in-flat-csv)

PR #23, opened 2019-10-03
[M1169] Change LIT validation to be within 50% of transect length

saanobhaai:


saanobhaai:
[update LIT length warning logic](https://trello.com/c/ZRCwsT2W/1169-update-lit-length-warning-logic)

PR #26, opened 2019-10-08
[M1132] Ingest templates and PIT & Fish Belt serializers

gridcell:


gridcell:
[draft observation schema](https://trello.com/c/ECwCFYxn/1132-draft-observation-schema)

src/api/ingest/serializers.py
```diff
@@ -0,0 +1,511 @@
...
+    _reverse_choices = {}
+
+    # PROJECT RELATED CHOICES
+    project_choices = {
+        "data__sample_event__site": None,
+        "data__sample_event__management": None,
+    }
+
+    # FIELDS
+    id = serializers.UUIDField(format="hex_verbose")
```
saanobhaai: Haven't seen us need to use `hex_verbose` before... Is this a not the default any more?

src/api/ingest/serializers.py
```diff
@@ -0,0 +1,511 @@
...
+        child=serializers.CharField(), allow_empty=False
+    )
+
+    class Meta:
+        list_serializer_class = CollectRecordCSVListSerializer
+
+    def __init__(self, data=None, instance=None, project_choices=None, **kwargs):
+        if instance is not None:
+            raise NotImplementedError("instance argument not implemented")
+        if isinstance(data, Mapping):
```
saanobhaai: Where would `Mapping` be used?

src/api/ingest/serializers.py
```diff
@@ -0,0 +1,511 @@
...
+    def formatted_errors(self):
+        errors = self.errors
+        return self.format_error(errors)
+
+
+class BenthicPITCSVSerializer(CollectRecordCSVSerializer):
+    protocol = "benthicpit"
+    observations_field = "data__obs_benthic_pits"
+    ordering_field = "data__obs_benthic_pits__interval"
+    header_map = CollectRecordCSVSerializer.header_map
```
saanobhaai: Can't this be `self.header_map`?

src/api/ingest/serializers.py
```diff
@@ -0,0 +1,511 @@
...
+    )
+    data__sample_event__notes = serializers.CharField(required=False, allow_blank=True)
+
+    # data__observers = ProjectProfileSerializer(many=True)
+    data__observers = serializers.ListField(
+        child=serializers.CharField(), allow_empty=False
+    )
+
+    class Meta:
+        list_serializer_class = CollectRecordCSVListSerializer
```
saanobhaai: interesting... so how is `many_init` forced to be always true? If I called a subclass of this without `many=True` what would happen?

src/api/ingest/serializers.py
```diff
@@ -0,0 +1,511 @@
...
+        required=False,
+        allow_null=True,
+        allow_blank=True,
+    )
+    data__sample_event__tide = serializers.ChoiceField(
+        choices=tide_choices, required=False, allow_null=True, allow_blank=True
+    )
+    data__sample_event__notes = serializers.CharField(required=False, allow_blank=True)
+
+    # data__observers = ProjectProfileSerializer(many=True)
```
saanobhaai: interesting -- but if the `child` approach works, delete comment

src/api/ingest/serializers.py
```diff
@@ -0,0 +1,511 @@
...
+            [
+                (field_name, field.get_value(self._original_data))
+                for field_name, field in self.fields.items()
+                if (field.get_value(self._original_data) == empty)
+                and not field.read_only
+            ]
+        )
+
+    def validate(self, data):
+        # Validate common Transect level fields
```
saanobhaai: Why doesn't this call `super()`?

src/api/ingest/serializers.py
gridcell: just being explicit, there are 4 options it can be

src/api/ingest/serializers.py
gridcell: If you only passed in 1 record

src/api/ingest/serializers.py
gridcell: self isn't defined yet, header_map is first inheriting `header_map` from `CollectRecordCSVSerializer` and then extending on the next line

src/api/ingest/README.md
```diff
@@ -0,0 +1,75 @@
...
+- `project_choices`: pre-fetched choices that require a project id, example: ProjectProfile choices.
+- `error_row_offset`: An offset to add to $row_number value in formatted validation errors.  Default: 1
+
+
+### Other
+
+- Records created by serializer are batched, NO signals will be triggered.
+
+
+## Example Using Ingest Serializer
```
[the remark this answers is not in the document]
saanobhaai: Never minda -- separate ticket

src/api/ingest/serializers.py
gridcell: Added to README that it must be True

PR #27, opened 2019-10-09
[M1175] set default ordering of observation models

gridcell:


gridcell:
[benthic PIT interval order error](https://trello.com/c/CGp74z4D/1175-benthic-pit-interval-error)

PR #29, opened 2019-10-09
[M1177] get SECRET_KEY from env

saanobhaai:


saanobhaai:
[get SECRET_KEY from env](https://trello.com/c/fDNGJYQX/1177-get-secretkey-from-env)

PR #30, opened 2019-10-15
[M1180] Add management name to sample units endpoint

gridcell:


gridcell:
[add MR as column to Submitted table](https://trello.com/c/ocdz7ZQR/1180-add-mr-as-column-to-submitted-table)

PR #31, opened 2019-10-18
[M1182] DJ command to ingest Fishbelt and Pit csv files

gridcell:


gridcell:
[create commands (and tests) to run ingest for BenthicPITCSVSerializer and FishBeltCSVSerializer - maybe include wipe of previous CRs as option (bypassing signals to avoid ArchiveRecord writes)?](https://trello.com/c/k4l6isyp/1182-create-commands-and-tests-to-run-ingest-for-benthicpitcsvserializer-and-fishbeltcsvserializer-maybe-include-wipe-of-previous-crs)

PR #32, opened 2019-10-22
[M1191] Collect record ingest endpoint

gridcell:


gridcell:
[create API endpoints for creating collect records via csv](https://trello.com/c/g6jjmT5R/1191-create-api-endpoints-for-creating-collect-records-via-csv)

src/api/ingest/benthicpit_serializer.py
```diff
@@ -5,13 +5,17 @@
 from .serializers import CollectRecordCSVSerializer, build_choices
 
 
+__all__ = ["BenthicPITCSVSerializer"]
```
saanobhaai: Why is this necessary again? Without it, if you did `from benthicpit_serializer import *` wouldn't you get the same thing?

src/api/ingest/serializers.py
```diff
@@ -12,6 +12,12 @@
 from ..resources.choices import ChoiceViewSet
 
 
+__all__ = [
```
saanobhaai: Again I may benefit from having this explained... how does this file know about the classes in its sister files?

src/api/ingest/serializers.py
```diff
@@ -393,9 +405,12 @@ def create(self, validated_data):
 
         return output
 
+    def reverse_kv_lookup(self, lookup):
```
saanobhaai: Should this be in `utils`? It's static.

src/api/resources/collect_record.py
```diff
@@ -238,3 +230,40 @@ def missing(self, request, *args, **kwargs):
...
+
+        protocol = request.data.get("protocol")
+        uploaded_file = request.FILES.get("file")
+        profile = request.user.profile
+        dryrun = truthy(request.data.get("dryrun"))
+
+        if protocol is None:
+            return Response("Missing protocol", status=400)
+
+        if uploaded_file is None:
```
saanobhaai: Don't know how it would be possible, but should we similarly protect against missing `profile.id`?

src/api/resources/collect_record.py
```diff
@@ -238,3 +230,40 @@ def missing(self, request, *args, **kwargs):
...
+        detail=False, methods=["POST"], permission_classes=[ProjectDataAdminPermission]
+    )
+    def ingest(self, request, project_pk, *args, **kwargs):
+        supported_content_types = (
+            "application/csv",
+            "application/x-csv",
+            "text/csv",
+            "text/comma-separated-values",
+            "text/x-comma-separated-values",
+            "text/tab-separated-values",
```
saanobhaai: Are we sure we want this? This would mean no tabs in notes fields, basically, right?

src/api/resources/collect_record.py
```diff
@@ -238,3 +230,40 @@ def missing(self, request, *args, **kwargs):
...
+
+        if uploaded_file is None:
+            return Response("Missing file", status=400)
+
+        content_type = uploaded_file.content_type
+        if content_type not in supported_content_types:
+            return Response("File type not supported", status=400)
+
+        decoded_file = uploaded_file.read().decode("utf-8").splitlines()
+        records, errors = ingest_fishbelt(
```
saanobhaai: Unless I'm missing something which function to call should depend on `protocol` (which wouldn't be passed in function)

src/api/resources/collect_record.py
```diff
@@ -238,3 +230,40 @@ def missing(self, request, *args, **kwargs):
...
+            decoded_file, project_pk, profile.id, protocol, dryrun
+        )
+
+        if errors:
+            return Response(errors, status=400)
+        return Response(CollectRecordSerializer(records, many=True).data)
```
saanobhaai: Does `.data` mean validated data, or is there a `validated _data` attribute? Seems like that would be the point of sending back the data, right, rather than just whatever they uploaded?

src/tools/management/commands/ingest.py
```diff
@@ -0,0 +1,118 @@
...
+
+        if _ingest is None:
+            raise NotImplementedError()
+
+        try:
+            with transaction.atomic():
+                sid = transaction.savepoint()
+                if clear_existing:
+                    self.clear_collect_records(project, protocol)
+                records, errors = _ingest(datafile, project, profile, protocol, dry_run)
```
saanobhaai: No `protocol` on the relevant functions

src/api/ingest/utils.py
```diff
@@ -0,0 +1,105 @@
...
+
+    return project_choices
+
+
+def _create_context(profile_id, request=None):
+    if request is None:
+        profile = Profile.objects.get_or_none(id=profile_id)
+        if profile is None:
+            raise ValueError("[{}] Profile does not exist.".format(profile_id))
+        token = tokenutils.create_token("google-oauth2|109519544860798433542")
```
saanobhaai: Remind me why the hardcoded token again?

src/api/ingest/benthicpit_serializer.py
gridcell: the way it currently is yes

src/api/resources/collect_record.py
gridcell: I didn't think the user would get this far if they didn't have a `profile.id`, Permission would stop them

src/api/resources/collect_record.py
gridcell: It's returning back the serialized CollectRecord instances

src/tools/management/commands/ingest.py
gridcell: _ingest is just an abstract variable Line 80, the specific ingest method is fetched based on protocol

PR #33, opened 2019-10-29
[M1193] add size_display to sample units endpoint

gridcell:


gridcell:
[Make size column sort (in transects tables)](https://trello.com/c/mtAOkOYB/1193-make-size-column-sort-in-transects-tables)

src/api/resources/sample_units/sample_unit_methods.py
```diff
@@ -2,7 +2,8 @@
 
 import re
 import django_filters
-from django.db.models import Case, CharField, Q, Value, When
+from django.db.models import Case, CharField, Q, Value, When, CharField
```
saanobhaai: `CharField` is repeated

PR #34, opened 2019-11-05
[M1161] change UpdateMixin to use filter_queryset instead of get_queryset

gridcell:


gridcell:
[Refactor projects to be offline first](https://trello.com/c/G8X8Pzxp/1161-refactor-projects-to-be-offline-first)

PR #35, opened 2019-11-05
[M1208] when parsing CR data, reef_slope shouldn't be empty string

gridcell:


gridcell:
[Empty string Reef Slope creates an error validation with the label `.`](https://trello.com/c/hifRVmwE/1208-empty-string-reef-slope-creates-an-error-validation-with-the-label)

PR #36, opened 2019-11-06
[M1186] Fix admin fish genus export and project delete

saanobhaai:


saanobhaai:
[fishgenus admin export to csv 500 error](https://trello.com/c/bEgfelrC/1186-fishgenus-admin-export-to-csv-500-error)

PR #38, opened 2019-11-07
[M1213] check for observers that don't match

gridcell:


gridcell:
[Ingest: Observers are being set with null, when there's no email match](https://trello.com/c/zSE5oiKs/1213-ingest-observers-are-being-set-with-null-when-theres-no-email-match)

PR #39, opened 2019-11-08
[M1215] Change Project endpoint to show by profile by default

gridcell:


gridcell:
[Make projects endpoint by default only show projects by profile](https://trello.com/c/zkrouvLu/1215-make-projects-endpoint-by-default-only-show-projects-by-profile)

PR #40, opened 2019-11-12
[M855b] Migrate MR rules to BooleanFields, no nulls allowed

saanobhaai:


saanobhaai:
[reconcile legacy MR rules](https://trello.com/c/MO5tBjRO/855-reconcile-legacy-mr-rules)

PR #42, opened 2019-11-19
[M1200] Add size-dependent width options to transect width choices

gridcell:


gridcell:
[add size-dependent width options to transect width choices](https://trello.com/c/DW5Rxwy0/1200-add-size-dependent-width-options-to-transect-width-choices)

PR #46, opened 2020-01-22
Add missing property for mermaid choices tables

tienhoah:


tienhoah:
[reference fish and benthic pages: add Export button to generate csv (as we do for Sample Units)](https://trello.com/c/jNrNQUZD/871-reference-fish-and-benthic-pages-add-export-button-to-generate-csv-as-we-do-for-sample-units)

PR #47, opened 2020-01-22
redo commands to use new fabric

saanobhaai:


saanobhaai:
[redo fabric commands](https://trello.com/c/lNYU3GhO/1259-redo-fabric-commands)

PR #48, opened 2020-02-09
[M2142] Future date validation

gridcell:


gridcell:
[future date api validation](https://trello.com/c/0eWyQA7t/1242-future-date-api-validation)

PR #51, opened 2020-03-06
[M1298] change transect length min validation default value

saanobhaai:


saanobhaai:
[make benthic pit, lit, and hb transect length warning minimum 10 (currently 25)](https://trello.com/c/sz3BSZwk/1298-make-benthic-pit-lit-and-hb-transect-length-warning-minimum-10-currently-25)

PR #54, opened 2020-03-11
[M1262] no test sites/MRs, and ingest encoding tweak

saanobhaai:


saanobhaai:
[exclude sites/mrs from api querysets that are from projects with test=true](https://trello.com/c/Nd4iQBPv/1262-exclude-sites-mrs-from-api-querysets-that-are-from-projects-with-testtrue)

PR #55, opened 2020-03-11
[M1295] support AGRRA fish size bins

saanobhaai:


saanobhaai:
[AGRRA mixed bin choice](https://trello.com/c/bJG5jY2A/1295-agrra-mixed-bin-choice)

src/api/models/mermaid.py
```diff
@@ -1276,10 +1276,10 @@ class Meta:
 
 
 class FishSizeBin(BaseChoiceModel):
-    val = models.PositiveSmallIntegerField()
+    val = models.CharField(max_length=100)
```
gridcell: missing migration for this change.  Also, does this need a migration?

saanobhaai: Not getting you -- it's above, no? Is this missing something?
https://github.com/data-mermaid/mermaid-api/pull/55/files#diff-9f592f5966e57bd922e7bed82a3889b3R11

PR #66, opened 2020-04-29
[M1127] attribute/region validations

gridcell:


gridcell:
[non-region attribute validations](https://trello.com/c/EYLu6e8t/1127-non-region-attribute-validations)

gridcell:
[non-region attribute validations](https://trello.com/c/EYLu6e8t/1127-non-region-attribute-validations)

src/api/models/mermaid.py
```diff
@@ -1007,6 +1007,17 @@ class FishAttribute(BaseAttributeModel):
     class Meta:
         db_table = 'fish_attribute'
 
+    def _get_subclass(self):
```
saanobhaai, line 1010: Isn't this a duplicate of `_get_taxon`?

gridcell: overlooked

PR #67, opened 2020-04-29
[M993] Quoted table search

gridcell:


gridcell:
[Smarter search](https://trello.com/c/Q9sBRQZW/993-smarter-search)

PR #70, opened 2020-05-04
[M1345b] optimize fishbelt se query

saanobhaai:


saanobhaai:
[fix aggregated endpoint timeouts (504)](https://trello.com/c/M64cnxML/1345-fix-aggregated-endpoint-timeouts-504)

PR #72, opened 2020-05-25
[M1183] Bleaching ingest

gridcell:


gridcell:
[create bleaching ingestion templates/serializers/commands](https://trello.com/c/AIlnGlWn/1183-create-bleaching-ingestion-templates-serializers-commands)

src/api/ingest/benthicpit_serializer.py
```diff
@@ -25,14 +25,15 @@ def benthic_attributes_choices():
...
-    observations_field = "data__obs_benthic_pits"
+    observations_fields = ["data__obs_benthic_pits"]
     ordering_field = "data__obs_benthic_pits__interval"
     additional_group_fields = CollectRecordCSVSerializer.additional_group_fields.copy()
     additional_group_fields.append("data__benthic_transect__label")
 
     header_map = CollectRecordCSVSerializer.header_map.copy()
     header_map.update(
         {
+            "Observation interval *": "data__obs_benthic_pits__interval",
```
saanobhaai, line 36: Was this really just overlooked previously? There's no similar field missing from fish belt?

src/api/ingest/utils.py
```diff
@@ -4,18 +4,22 @@
...
+    BleachingCSVSerializer,
+    FishBeltCSVSerializer,
+)
 from api.models import (
     BENTHICLIT_PROTOCOL,
     BENTHICPIT_PROTOCOL,
     BLEACHINGQC_PROTOCOL,
     FISHBELT_PROTOCOL,
     HABITATCOMPLEXITY_PROTOCOL,
+    CollectRecord,
```
saanobhaai, line 18: Why is this needed?

src/api/ingest/benthicpit_serializer.py
gridcell: The observation interval was on the base serializer and I moved it out.  Yeah I think it was overlooked

src/api/ingest/utils.py
gridcell: Formatting change via `isort`

PR #73, opened 2020-05-25
[M1346] Only update record if validation results are different

gridcell:


gridcell:
[sites listing still slow after editing](https://trello.com/c/MMSMhyma/1346-sites-listing-still-slow-after-editing)

PR #75, opened 2020-05-26
[M1355] Add api version to response header

gridcell:


gridcell:
[send API version in every response header](https://trello.com/c/MR9KvZlA/1355-send-api-version-in-every-response-header)

ci_cd/version.sh
```diff
@@ -0,0 +1,9 @@
+#! /bin/bash
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+COMMIT=`git log --format="%H" -n 1`
+echo "Commit id: $COMMIT"
+VERSION=`git describe --contains "$COMMIT" || git rev-parse --abbrev-ref HEAD`
```
saanobhaai, line 6: I'm a little unclear on this... This is simpler; would it potentially miss something or fail?
`git describe --tags --abbrev=0`

PR #81, opened 2020-06-02
[M1364] Ingest bulk validation/submit

gridcell:


gridcell:
[Add flags to ingest to bulk validate/submit](https://trello.com/c/3RTUGxue/1364-add-flags-to-ingest-to-bulk-validate-submit)

PR #82, opened 2020-06-05
Add a new MailChimp field to identify contacts created from API

jgnasser:


jgnasser:
[revisit mailchimp options](https://trello.com/c/MmHCrGIS/1234-revisit-mailchimp-options)

PR #83, opened 2020-06-05
[M1354] Add extra column mapping for ingestion errors

gridcell:


gridcell:
[handle incorrect ingestion schema headers](https://trello.com/c/WAFMvbcv/1338-handle-incorrect-ingestion-schema-headers)

PR #84, opened 2020-06-08
[M1359] Use size bin min_val when available for fish size validation

gridcell:


gridcell:
[Adjust fish maximum size warnings wrt bin size](https://trello.com/c/BfUcVKJe/1359-adjust-fish-maximum-size-warnings-wrt-bin-size)

PR #85, opened 2020-06-09
Update MailChimp profile for existing subscribers

jgnasser:


jgnasser:
[revisit mailchimp options](https://trello.com/c/MmHCrGIS/1234-revisit-mailchimp-options)

src/api/auth_backends.py
```diff
@@ -103,11 +103,12 @@ def _validate_profile(self, payload):
                         client = MailChimp(
                             mc_api=settings.MC_API_KEY, mc_user=settings.MC_USER
                         )
-                        client.lists.members.create(
+                        client.lists.members.create_or_update(
                             settings.MC_LIST_ID,
+                            hashlib.md5(profile.email.encode("utf-8")).hexdigest(),
```
saanobhaai, line 108: Do you need `.encode("utf-8")`?

jgnasser: Not 100% sure but based on what I saw it's needed for Python 3 to make the hash work. But probably if we are sure profile.email is utf-8 it may be unnecessary

saanobhaai: I think it's worth testing locally -- I took a lot of these out when we moved to python 3

PR #88, opened 2020-06-18
[M1357] Refactor Sample Event and Sample Units

gridcell:


gridcell:
[refactor SE/SU schema](https://trello.com/c/8naL7bbQ/1357-refactor-se-su-schema)

PR #90, opened 2020-07-03
[M1375] handle fish attr regions that are not querysets in validation

saanobhaai:


saanobhaai:
[CollectRecords validate 500](https://trello.com/c/DIQGXHe1/1375-collectrecords-validate-500)

PR #91, opened 2020-07-05
Update fab tasks

vpop:
This will align the documentation with the changes done in the b07b9bbee30d734832734ca7deea0ac5929b27f2 commit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,11 +80,11 @@ Then the containers will be started.
 
 With a database already created and persisted in an S3 bucket via 
 ```sh
-$ fab db_backup
+$ fab dbbackup
 ``` 
 ,
 ```sh
-$ fab db_restore
+$ fab dbrestore
 ``` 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -96,7 +96,7 @@ to create its schema.
 A shortcut for the above steps, once S3 is set up, is available via:
 
 ```
-$ fab fresh_install:[env]
+$ fab fresh-install:[env]
```
saanobhaai, line 99: I think we might as well take out the hyphen here too, for consistency (something we've done on other projects).

vpop: Sure, makes sense. Should I also remove the `fresh-install` alias from the [task](https://github.com/data-mermaid/mermaid-api/blob/master/fabfile.py#L118)?

saanobhaai: Sure, not needed any more I think

PR #96, opened 2020-07-15
[M1379] Add access restriction to management rules

gridcell:


gridcell:
[new "access restrictions" Management Regime rule](https://trello.com/c/2vGxFkUf/1379-new-access-restrictions-management-regime-rule)

src/tools/management/commands/find_bad_mrs.py
```diff
@@ -97,7 +100,8 @@ def handle(self, *args, **options):
                     Q(no_take=True) |
                     Q(size_limits=True) |
                     Q(gear_restriction=True) |
-                    Q(species_restriction=True)
+                    Q(species_restriction=True) |
+                    Q(access_restriction=False)
```
saanobhaai, line 104: should be `access_restriction=True`

src/tools/management/commands/find_bad_mrs.py
```diff
@@ -26,7 +26,8 @@ def __init__(self):
         self.outpath = ''
         self.header = ['name', 'secondary name', 'project', 'project admins', 'year est',
                        'open_access', 'periodic_closure', 'size_limits',
-                       'gear_restriction', 'species_restriction', 'no_take'
+                       'gear_restriction', 'species_restriction', 'no_take',
+                       'access_restriction'
```
saanobhaai, line 30: Also need to add after line 59 below

PR #97, opened 2020-07-15
[M1371] switch to use belt transect names instead of values

gridcell:


gridcell:
[refactor ingestion handling of BeltTransectWidth to use name rather than val](https://trello.com/c/xOkBOhS2/1371-refactor-ingestion-handling-of-belttransectwidth-to-use-name-rather-than-val)

PR #98, opened 2020-07-16
[M1381] incorrect PROPOSED object

saanobhaai:


saanobhaai:
[PROPOSED attribute status error](https://trello.com/c/0YIUNtwF/1381-proposed-attribute-status-error)

PR #99, opened 2020-07-16
PROPOSED hotfix second try

saanobhaai:


saanobhaai:
[PROPOSED attribute status error](https://trello.com/c/F4Mbqlk2/1382-proposed-attribute-status-error)

PR #102, opened 2020-07-20
M88: Compress DB backups and switch to pg_restore

gridcell:


gridcell:
[compress db backups](https://trello.com/c/85KbE84C/88-compress-db-backups)

PR #103, opened 2020-07-31
[M1389] revert clearexisting delete call to raw sql

saanobhaai:


saanobhaai:
[ingestion iteration issues](https://trello.com/c/8BHwNeZa/1389-ingestion-iteration-issues)

PR #105, opened 2020-08-05
[M1386-1] api support for transect length 1 decimal value

tienhoah:


tienhoah:
[Collect - allow PIT transect length to have one decimal, e.g., 10.5m length (currently, integer only)](https://trello.com/c/OFDr06ry/1386-collect-allow-pit-transect-length-to-have-one-decimal-eg-105m-length-currently-integer-only)

PR #107, opened 2020-08-18
M1396: create/update mixin for offline models

gridcell:


gridcell:
[Site creation IntegrityError](https://trello.com/c/VlwJRCTi/1396-site-creation-integrityerror)

PR #108, opened 2020-08-18
M1397: ingest reef slope fix

gridcell:


gridcell:
[Incorrect "reef slope" lookups on ingestion](https://trello.com/c/UyWffhKL/1397-incorrect-reef-slope-lookups-on-ingestion)

PR #112, opened 2020-09-16
M1303: add sample_time range validation

gridcell:


gridcell:
[add validation warning to SU time > 7pm and < 6am](https://trello.com/c/hwCGVz2L/1303-add-validation-warning-to-su-time-7pm-and-6am)

src/api/submission/protocol_validations.py
```diff
@@ -113,7 +116,9 @@ def validate(self):
 
 class SampleUnitValidation(ProtocolValidation):
     DEPTH_RANGE = (1, 30)
+    DATE_TIME_RANGE = (datetime.time(6,0), datetime.time(19,0))
     DEPTH_MSG = str(_("Depth value outside range of {} and {}".format(*DEPTH_RANGE)))
+    DATE_TIME_MSG = str(_("Sample time outside of range {} and {}".format(*DATE_TIME_RANGE)))
```
saanobhaai, line 121: Do we need to wrap the `_()` with `str()` for all these?

gridcell: str() forces it to be evaluated right away rather than lazy load

PR #114, opened 2020-11-11
fix incorrect FishFamily/Genus biomass constants, jsonb agg fields

saanobhaai:


saanobhaai:
[SE/SU bugfixes](https://trello.com/c/JkMkyo7Y/1410-se-su-bugfixes)

src/api/admin/mermaid.py
```diff
@@ -651,7 +651,7 @@ class FishAttributeGroupingAdmin(FishAttributeAdmin):
     def region_list(self, obj):
         if not hasattr(obj, "regions"):
             return []
-        return ", ".join([r.name for r in obj.regions.order_by("name")])
```
gridcell, line 654: does the previous one just need to be changed to `obj.regions.all().order_by("name")`

saanobhaai: `obj.regions` is just a list of `UUID`s

PR #118, opened 2020-12-15
[M1418] add agg fields to fishbelt and benthiclit SU endpoints

saanobhaai:


saanobhaai:
[Fishbelt SU view does not have "Total abundance" (which is in Collect)](https://trello.com/c/PCBFAaBP/1418-fishbelt-su-view-does-not-have-total-abundance-which-is-in-collect)

PR #121, opened 2021-01-30
[M1429] Job, Queue, Worker

gridcell:
Ticket will be split into 2 pull requests first being the queueing system, second (yet to come) will be the summary endpoint caching .

### DIFF
--- a/src/simpleq/jobs.py
+++ b/src/simpleq/jobs.py
@@ -1,0 +1,103 @@
+import codecs
+import uuid
+from datetime import datetime
+from pickle import dumps, loads
+
+
+class Job:
+    """An abstraction for a single unit of work (a job!)."""
+
+    def __init__(self, job_id, callable, *args, **kwargs):
+        """
+        Create a new Job,
+
+        :param obj callable: [optional] A callable to run.
+        """
+        self._id = job_id if job_id is not None else str(uuid.uuid4())
+        self.start_time = None
+        self.stop_time = None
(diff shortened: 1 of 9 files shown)


src/simpleq/queues.py
```diff
@@ -0,0 +1,219 @@
...
+        messages = self.queue.receive_messages(
+            MaxNumberOfMessages=self.BATCH_SIZE,
+            WaitTimeSeconds=self.WAIT_SECONDS,
+            MessageAttributeNames=["id"],
+        )
+        duplicate_job_groups = dict()
+
+        for message in messages:
+            job = Job.from_message(message)
+            if job.id is not None and job.id in duplicate_job_groups:
```
saanobhaai, line 211: As I understand it, based on `content_deduplication` this should never happen, but is good to have in case `content_deduplication` was not true?

src/simpleq/jobs.py
```diff
@@ -0,0 +1,103 @@
...
+    @property
+    def message(self):
+        return dict(
+            MessageAttributes={
+                "id": {
+                    "StringValue": self.id,
+                    "DataType": "String"
+                }
+            },
+            MessageGroupId="mermaid",
```
saanobhaai, line 49: Should this be from a var, and/or should it be appended with `-ENVIRONMENT`?

saanobhaai (changes requested):
Just a couple minor questions

PR #124, opened 2021-02-07
[M1429-2] SQL Models

gridcell:


gridcell:
[improve aggregated endpoint performance](https://trello.com/c/VmhKeFJN/1429-improve-aggregated-endpoint-performance)

src/api/scripts/test_sql_models.py
```diff
@@ -0,0 +1,300 @@
+from api.models import (
```
saanobhaai, line 1: These are great... are you thinking we'll remove the MV versions down the road? If so we should create a ticket for that, so we don't forget. Also we already merged in the SQS code which we're not (yet) using... all fine, just don't want a lot of code hanging around that we're not using.

src/sqltables/query.py
```diff
@@ -0,0 +1,201 @@
...
+                    params = []
+
+                alias = self.join(
+                    SQLTable(self.get_meta().db_table, None, params, self.model_sql)
+                )
+            else:
+                alias = self.join(BaseTable(self.get_meta().db_table, None))
+        return alias
+
+    def setup_joins(  # noqa: C901
```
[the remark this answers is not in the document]
gridcell: telling flake8 to ignore the complexity of the method

src/api/scripts/test_sql_models.py
gridcell: we can chat about this further

PR #125, opened 2021-02-08
Ported fab commands for UI devs For those that don't want to deal with setting up python and fabric locally

gridcell:


Makefile
```diff
@@ -1,3 +1,34 @@
 
-.DEFAULT:
-	fab "$@"
\ No newline at end of file
+down:
+	@docker-compose down
+
+build_image:
```
saanobhaai, line 5: Maybe we should use same syntax as in fabric to decrease potential future confusion? i.e. `buildnocache` for this one
Also, don't we need to execute `./ci_cd/version.sh` here?

Makefile
```diff
@@ -1,3 +1,34 @@
...
+	@docker-compose build --no-cache --pull
+
+up:
+	@docker-compose up -d
+
+dbrestore:
+	@docker exec -it api_service python manage.py dbrestore local
+
+migrate:
+	@docker exec -it api_service python manage.py migrate
```
saanobhaai, line 15: Don't know if it's worth making `api_service` a variable...

Makefile
```diff
@@ -1,3 +1,34 @@
...
+down:
+	@docker-compose down
+
+build_image:
+	@docker-compose build --no-cache --pull
+
+up:
+	@docker-compose up -d
+
+dbrestore:
```
saanobhaai, line 11: wouldn't hurt to do `dbbackup` too maybe?

Makefile
```diff
@@ -1,3 +1,34 @@
...
+	@docker exec -it api_service python manage.py dbrestore local
+
+migrate:
+	@docker exec -it api_service python manage.py migrate
+
+fresh_install:
```
saanobhaai, line 17: as above -- maybe `freshinstall`

mmnoo:
Could these changes come with an update to the readme?

PR #131, opened 2021-02-26
M1212: limit error maximum to 1000 errors

gridcell:


gridcell:
[ingest error feedback throttling](https://trello.com/c/UCwSCnmA/1212-ingest-error-feedback-throttling)

PR #133, opened 2021-03-29
M1497: remove validation error when there is no bleaching percent cover obs

gridcell:


gridcell:
[make bleaching QC % cover minimum observations validation warning, not error](https://trello.com/c/vdtYt2xk/1497-make-bleaching-qc-cover-minimum-observations-validation-warning-not-error)

PR #134, opened 2021-03-29
M1517: set DEBUG to False for dev/prod environments

gridcell:


gridcell:
[Django DEBUG setting should be False for prod and dev environments](https://trello.com/c/91lB8YZZ/1517-django-debug-setting-should-be-false-for-prod-and-dev-environments)

PR #139, opened 2021-04-13
[M1531] Refactor SQL Tables to use Dictionary instead of tuple for sql params

gridcell:
- removed all the unsupported join code


gridcell:
[Refactor sql tables to use dict params instead of tuple](https://trello.com/c/pdicQppr/1531-refactor-sql-tables-to-use-dict-params-instead-of-tuple)

PR #143, opened 2021-04-21
M1523 Revision pull endpoint

gridcell:
Includes:
- revision model and migration
- database triggers to populate revisions table for data that can be edited offline
- tests


gridcell:
[Create /pull endpoint](https://trello.com/c/xi7zPZD2/1523-create-pull-endpoint)

src/api/models/revisions.py
```diff
@@ -0,0 +1,207 @@
...
+        unique_together = ("table_name", "record_id")
+
+    def __str__(self):
+        return f"[{self.revision_num}] {self.table_name} {self.record_id}"
+
+
+# -- TRIGGER SQL --
+
+forward_sql = """
+    CREATE SEQUENCE revision_seq_num START 1;
```
saanobhaai, line 26: Just wondering... once we create this sequence, everything -- all users everywhere, offline, etc. -- depends on it (since it's not per-date/table/project/whatever), and I think adjustments or rollbacks might be difficult. Since we're doing this by trigger anyway, is it worth considering just doing the nextval logic internally, i.e. `max(revision_num) + 1`?

src/api/tests/fixtures/revisions_setup.py
```diff
@@ -0,0 +1,23 @@
...
+    )
+
+    collect_record.save()
+    
+    CollectRecord.objects.create(
+        project=project1, profile=profile1, data=dict()
+    )
+    
+    rev_rec = Revision.objects.get(record_id=collect_record.id)
+    rev_rec.id
```
saanobhaai, line 19: What does this do?

saanobhaai (changes requested):

src/api/resources/sync/views.py
```diff
@@ -0,0 +1,166 @@
...
+        "last_revision_num": -1,
+    }
+
+
+@api_view(http_method_names=["POST"])
+def pull(request):
+    response_data = {}
+    source_types = request.data.keys()
+
+    check_permissions(request, source_types)
```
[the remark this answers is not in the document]
gridcell: if it fails it raises a 403 error

src/api/resources/sync/views.py
```diff
@@ -0,0 +1,166 @@
...
+    params = {"revision_num": data.get("last_revision")}
+
+    if "project" in required_filters:
+        project = data.get("project")
+        if project is None:
+            raise ValueError("Project is missing")
+        params["project"] = project
+
+    if "profile" in required_filters:
+        user = request.user
```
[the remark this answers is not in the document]
gridcell: profile should exist by this point because it's used during the permissions checks

src/api/models/revisions.py
gridcell: I left the sequence but am setting start value after the sequence has been created

```
CREATE SEQUENCE revision_seq_num START 1;
SELECT setval('revision_seq_num', (SELECT MAX(revision_num) FROM revision));
```

src/api/tests/fixtures/revisions_setup.py
gridcell: populates the rev_rec object before collect record gets deleted (a couple lines down) otherwise the rev_rec values will be different.

I added a note to the code of what's going on

src/api/resources/sync/push.py
```diff
@@ -0,0 +1,85 @@
...
+    """
+    if record.get("_deleted") is True:
+        return "DELETE"
+    elif record.get("_last_revision_num") is None:
+        return "POST"
+    else:
+        return "PUT"
+
+
+def has_conflicts(record_id, last_revision_num):
```
saanobhaai, line 23: Does this need to be a separate function? It's only used in one place, is short, and could be misleading in other contexts (i.e. the `>` comparison only means a conflict in the context of a push)

src/api/resources/sync/views.py
```diff
@@ -0,0 +1,348 @@
...
+        elif status_code == 409:
+            msg = "Conflict"
+            data = _get_serialized_record(serializer, record_id)
+        else:
+            msg = ""
+            data = _get_serialized_record(serializer, record_id)
+
+        return {"status_code": status_code, "message": msg, "data": data}
+    except Exception as err:
+        return _error(500, err)
```
saanobhaai, line 225: Is there anywhere else we should be returning a 500? If it's really on the level of a 500, should we attempt to hook into logging so we're notified the same way?

saanobhaai (changes requested):
Just minor comments

PR #146, opened 2021-05-07
[M1548] Add force flag to push endpoint

gridcell:
Option for bypassing conflict during a push


gridcell:
[Add flag to bypass push conflicts](https://trello.com/c/vBmLkUC4/1548-add-flag-to-bypass-push-conflicts)

gridcell:
[Add flag to bypass push conflicts](https://trello.com/c/vBmLkUC4/1548-add-flag-to-bypass-push-conflicts)

src/api/resources/sync/views.py
```diff
@@ -311,30 +343,18 @@ def vw_pull(request):
...
+        for source_type, source_data in request_data.items()
+    }
 
     return Response(response_data)
 
 
 @api_view(http_method_names=["POST"])
 def vw_push(request):
     request_data = request.data or {}
+    force = str(request.query_params.get("force")).lower() in ("1", "t", "true", "y", "yes")
```
saanobhaai, line 357: Don't we have a "truthy" util function somewhere? Also wonder if a `strip()` is in order.

src/api/signals.py
```diff
@@ -300,3 +307,26 @@ def run_cr_management_validation(sender, instance, *args, **kwargs):
...
+@receiver(post_save, sender=FishGenus)
+@receiver(post_delete, sender=FishGenus)
+@receiver(post_save, sender=FishSpecies)
+@receiver(post_delete, sender=FishSpecies)
+@receiver(post_save, sender=BenthicAttribute)
+@receiver(post_delete, sender=BenthicAttribute)
+def bust_revision_cache(sender, instance, *args, **kwargs):
+    if sender is FishSpecies:
+        print(f"sender: {sender}")
+        cache.delete(FISH_SPECIES_SOURCE_TYPE)
```
saanobhaai, line 323: Isn't this the wrong way around? If somebody edits/deletes a species, we need to recalc averages for genus and family, right? I'm wondering if we shouldn't just delete them all if the sender is any of them, to be safe. Will hardly ever happen.

src/app/settings.py
```diff
@@ -303,8 +303,8 @@ def show_toolbar(request):
 
 CACHES = {
     'default': {
-        'BACKEND': 'django.core.cache.backends.locmem.LocMemCache',
-        'LOCATION': 'locmem',
+        'BACKEND': 'django.core.cache.backends.db.DatabaseCache',
```
saanobhaai, line 306: Any reason not to keep it in memory? Too large?

Are there any other places in the code where we were/are using cache that need to get cleaned up or refactored for this?

saanobhaai (changes requested):

src/app/settings.py
gridcell: Doesn't support cross processor

src/api/signals.py
```diff
@@ -318,13 +318,7 @@ def update_with_covariates(sender, instance, *args, **kwargs):
...
 @receiver(post_delete, sender=BenthicAttribute)
 def bust_revision_cache(sender, instance, *args, **kwargs):
-    if sender is FishSpecies:
-        print(f"sender: {sender}")
-        cache.delete(FISH_SPECIES_SOURCE_TYPE)
-    elif sender is FishGenus:
-        cache.delete(FISH_SPECIES_SOURCE_TYPE)
-        cache.delete(FISH_GENERA_SOURCE_TYPE)
-    elif sender is FishFamily:
+    if sender is (FishSpecies, FishGenus, FishFamily):
```
saanobhaai, line 321: shouldn't this be `is in`?

saanobhaai (changes requested):
Just one issue I'm seeing -- then (or if I'm missing something) good to commit

PR #148, opened 2021-05-21
[M1553] Validate the regex before making filtering

gridcell:
Checks the search regex string to see if it compiles before filtering sample unit methods and if fails raises a validation error rather than a 500 error

### DIFF
--- a/src/api/resources/sample_units/sample_unit_methods.py
+++ b/src/api/resources/sample_units/sample_unit_methods.py
@@ -168,7 +168,6 @@ class SampleUnitMethodView(BaseProjectApiViewSet):
     )
 
     def get_queryset(self):
-        print(self.request.query_params)
         qs = self.queryset
 
         protocol_condition = Case(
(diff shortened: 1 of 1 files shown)


gridcell:
[Invalid regular expression 500](https://trello.com/c/meLKBwBm/1553-invalid-regular-expression-500)

src/api/resources/sample_units/sample_unit_methods.py
```diff
@@ -43,6 +43,11 @@ class SearchNonFieldFilter(django_filters.Filter):
     def filter(self, qs, value):
         value = value or ""
 
+        try:
+            re.compile(value)
+        except re.error:
+            raise exceptions.ValidationError("Invalid search")
```
saanobhaai, line 49: So this exception will be handled and passed to response, and not cause a 500?

saanobhaai (changes requested):

src/api/resources/sample_units/sample_unit_methods.py
gridcell: Yeah causes a 400 error now

src/api/resources/sample_units/sample_unit_methods.py
```diff
@@ -163,6 +168,7 @@ class SampleUnitMethodView(BaseProjectApiViewSet):
     )
 
     def get_queryset(self):
+        print(self.request.query_params)
```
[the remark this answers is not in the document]
gridcell: removed

PR #149, opened 2021-05-24
[M1555] Return same revision number if no updates

gridcell:


gridcell:
[When pulling updates (`/pull`), and no updates exist, `last_revision_number` is null](https://trello.com/c/5xDFhXIu/1555-when-pulling-updates-pull-and-no-updates-exist-lastrevisionnumber-is-null)

PR #150, opened 2021-05-26
[M1556] Fix for setting project attribute when checking revision pull permissions

gridcell:
Before project permissions can be set, a `project` property must be set.   Previously this well set using the project id but in the case of `/pull` requests that isn't being passed, just `project`.  This PR handles both cases.

PR also fixes some issues with Permission error handling and reporting


gridcell:
[Projects /pull needs to include "project" property in payload](https://trello.com/c/Mpju9kXn/1556-projects-pull-needs-to-include-project-property-in-payload)

PR #151, opened 2021-06-08
[M1563] ingestion tweaks

saanobhaai:


saanobhaai:
[ingestion: look for opportunities to use ids/values instead of labels](https://trello.com/c/W5D3nOKv/1563-ingestion-look-for-opportunities-to-use-ids-values-instead-of-labels)

gridcell (approved):

PR #152, opened 2021-06-10
[M1562-2]  Use Viewsets instead of classes for fetching revision changes

gridcell:


gridcell:
[Remove project_id requirement from projects in `/pull`](https://trello.com/c/ccXZ8yh0/1562-remove-projectid-requirement-from-projects-in-pull)

src/api/resources/sync/pull.py
```diff
@@ -20,44 +39,75 @@ def _get_records(model_class, filters):
...
+            revision.record_id = model_select."__pk__"
         WHERE
-            {" AND ".join(filters)}
+            {" AND ".join(updates_filters)}
         ORDER BY
             revision.updated_on DESC, revision.revision_num  DESC
     """
 
-    return model_class.objects.raw(sql)
+    updates = model_class.objects.raw(updates_sql)
```
saanobhaai, line 53: Any reason not to turn this into a list of dicts, to be consistent with `deletes`?

src/api/resources/sync/pull.py
```diff
@@ -20,44 +39,75 @@ def _get_records(model_class, filters):
             revision.revision_num as revision_revision_num,
             revision.deleted as revision_deleted
         FROM
-            revision
-        LEFT JOIN
             model_select
+        INNER JOIN
```
saanobhaai, line 43: Just for my own curiosity, I'd be interested in a walkthrough of why you decided we could no longer do this in one query with left join

src/api/resources/sync/views.py
```diff
@@ -181,10 +181,15 @@ def _format_errors(errors):
...
 
 
-def _get_serialized_record(serializer, record_id):
-    record = get_record(model_class=serializer.Meta.model, record_id=record_id)
+def _get_serialized_record(viewset, record_id):
+    serializer = viewset.serializer_class
+    record = get_record(viewset, record_id)
 
-    data = serialize_revisions(serializer, [record])
+    if isinstance(record, dict):
```
saanobhaai, line 188: Why not have `get_records` return the same data type (list of dicts) for updates and deletes, and avoid this?

saanobhaai (changes requested):

src/api/resources/sync/pull.py
gridcell: The model instances get serialized with it's Serializer and we don't have that option for deletes.  I think some of this weill be refactored in M1561 so we can speed up the fishspecies call

PR #154, opened 2021-06-29
[M1532] Filter attributes by regions

gridcell:
- All attribute endpoints can now be filtered by regions (1 to many)


gridcell:
[enable filtering by `regions` on attributes](https://trello.com/c/NKPYU6rY/1532-enable-filtering-by-regions-on-attributes)

PR #155, opened 2021-06-29
[M1533] Add threaded delete to Project DELETE method endpoint

gridcell:


gridcell:
[handle project DELETE requests](https://trello.com/c/j8b9m94t/1533-handle-project-delete-requests)

PR #156, opened 2021-06-29
M1534: Lock down ProjectAuthenticatedUserPermission Block read-only from being able to do find/replace with sites and management regimes

gridcell:


gridcell:
[ensure ProjectAuthenticatedUserPermission requires project membership above level of "read-only"](https://trello.com/c/WVTEfArW/1534-ensure-projectauthenticateduserpermission-requires-project-membership-above-level-of-read-only)

PR #157, opened 2021-07-05
[M1535] Refactor out size_bin from observation records

gridcell:
- includes moving fish belt validation and protocol validation tests to pytest


gridcell:
[remove size_bin from fishbelt observations](https://trello.com/c/8a1x36YJ/1535-remove-sizebin-from-fishbelt-observations)

PR #158, opened 2021-07-06
M1539: Extend CollectRecordOwner permission to edit delete methods

gridcell:
- includes removing show all collect records (don't show collect records to non-owners) 
- added tests for the changes

### DIFF
--- a/src/api/resources/collect_record.py
+++ b/src/api/resources/collect_record.py
@@ -42,7 +42,10 @@ logger = logging.getLogger(__name__)
 class CollectRecordOwner(permissions.BasePermission):
     def has_permission(self, request, view):
         record_ids = request.data.get("ids") or []
-        if not record_ids:
+        pk = view.kwargs.get("pk")
+        if pk:
+            record_ids = [pk]
+        elif not record_ids:
             return True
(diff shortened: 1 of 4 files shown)


gridcell:
[refactor CollectRecord permissions](https://trello.com/c/WCk3B4fA/1539-refactor-collectrecord-permissions)

src/api/resources/collect_record.py
```diff
@@ -42,7 +42,10 @@
 class CollectRecordOwner(permissions.BasePermission):
     def has_permission(self, request, view):
         record_ids = request.data.get("ids") or []
-        if not record_ids:
+        pk = view.kwargs.get("pk")
```
saanobhaai, line 45: This makes me think maybe we should do a quick look at all other permissions classes to catch the same thing?

gridcell: @saanobhaai I checked sites, management and project profile I think we are OK.  Collect Record is bit of an exception because the records belong to the project and a person where the other records belong to the Project.

I also tested editing a ProjectProfile with the read-only and collector roles and was blocked

PR #159, opened 2021-07-08
[1569] Add caching to serializer LazyChoiceField

gridcell:
- fixed ingest tests
- added caching to potential improve performance of fields with large number of choices


gridcell:
[solution for ingests that currently take > 60s](https://trello.com/c/9EKsh8wp/1569-solution-for-ingests-that-currently-take-60s)

src/api/fields.py
```diff
@@ -16,9 +20,35 @@ def choice_strings_to_values(self):
...
+        # a small number of them when deserializing records.
+
+        if data == "" and self.allow_blank:
+            return ""
+
+        try:
+            str_data = str(data)
+            val = None
+            if str_data not in self._cached_choices:
+                val = self.choice_strings_to_values[str(data)]
```
saanobhaai, line 45: `val = self.choice_strings_to_values[str_data]`

src/api/fields.py
```diff
@@ -16,9 +20,35 @@ def choice_strings_to_values(self):
...
+            self._flattened_choices = fields.flatten_choices_dict(self.grouped_choices)
+        return self._flattened_choices
 
     def _set_choices(self, choices):
         self._choices = choices
+        self._flattened_choices = None
+        self._cached_choices = {}
+
+    def to_internal_value(self, data):
+        # Overriding this function add cache the choices that are
```
saanobhaai, line 33: grammar

src/api/fields.py
```diff
@@ -16,9 +20,35 @@ def choice_strings_to_values(self):
         return {str(key): key for key in self.choices}
 
     def _get_choices(self):
-        return fields.flatten_choices_dict(self.grouped_choices)
+        if self._flattened_choices is None:
+            self._flattened_choices = fields.flatten_choices_dict(self.grouped_choices)
+        return self._flattened_choices
```
saanobhaai, line 25: I'm sure this is right, but just wondering if `_flattened_choices` is necessary, if everything is going back to `cached_choices`?

saanobhaai (changes requested):

src/api/fields.py
gridcell: the method gets called several times.

PR #160, opened 2021-07-15
[M1551] Refactor tests that aren't using pytest

gridcell:


gridcell:
[refactor existing API tests](https://trello.com/c/shPB5uTq/1551-refactor-existing-api-tests)

src/api/tests/fixtures/collect_records.py
```diff
@@ -240,3 +239,218 @@ def invalid_collect_record_error(
...
+
+@pytest.fixture
+def invalid_benthic_lit_collect_record(
+    benthic_attribute_3,
+    project1,
+    profile1,
+    project_profile1,
+    sample_event1,
+):
+    observations = [
```
saanobhaai, line 287: Would it be considered bad form to DRY up all the pieces in the valid/invalid versions that are the same, but add the two observations for the valid version? Had to think a bit on why this was invalid.

src/api/tests/test_protocol_validations.py
```diff
@@ -1,242 +1,125 @@
-import copy
```
saanobhaai, line 1: I should know this, but -- I gather fixtures are automatically imported/processed/scoped based on a `fixtures` subdir name, so you don't have to import them here?

src/api/tests/test_protocol_validations.py
```diff
@@ -1,242 +1,125 @@
...
-        observations = [
-            dict(attribute=str(self.benthic_attribute1a.id), interval=5),
-            dict(attribute=str(self.benthic_attribute1a.id), interval=10),
-            dict(attribute=str(self.benthic_attribute1a.id), interval=15),
-            dict(attribute=str(self.benthic_attribute1a.id), interval=20),
-            dict(attribute=str(self.benthic_attribute1a.id), interval=25),
-            dict(attribute=str(self.benthic_attribute2c.id), interval=30),
+
+
+def test_fishbelt_protocol_validation_ok(
```
saanobhaai, line 10: Just for my own edification I'd be interested to know more about the move away from class-based tests.

saanobhaai (changes requested):
Mostly just a few questions I'd be interested to know more about, but looks great, and sheesh that was a lot of validation test writing. Might be good to review/document process for adding/editing/running tests as a result of code changes, as well as coverage and priorities, now that we have this solid basis.

fabfile.py
```diff
@@ -97,10 +97,10 @@ def lint(c):
 @task
 def test(c):
     """Run unit tests"""
-    local("docker exec -it api_service pytest --cov-report=html --cov=api --verbose")
+    local("docker exec -it api_service pytest --nomigrations --cov-report=html --cov=api --verbose api/tests")
```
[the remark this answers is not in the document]
gridcell: Wasn't going to, the Makefile currently is only being used by the front end devs so they don't have to install fab and have virtualenvs setup

src/api/tests/test_protocol_validations.py
gridcell: yup

src/api/tests/fixtures/collect_records.py
gridcell: I was on the fence with this and leaned towards readability and keeping it in one place

src/api/tests/test_protocol_validations.py
gridcell: - fixtures
- lots of pytest plugins
- less boilerplate code

PR #163, opened 2021-08-11
[M1519] Vibrant Oceans Threats Covariates

gridcell:


gridcell:

src/api/covariates/__init__.py
```diff
@@ -36,29 +17,75 @@ def update_site_covariates(site):
...
+    vibrant_oceans_threats = VibrantOceansThreatsCovariate()
+    results = vibrant_oceans_threats.fetch([(point.x, point.y)])
+
+    if not results or not results[0]:
+        return
+
+    result = results[0]
+
+    data_date = result["date"]
+    requested_date = result["requested_date"]
```
saanobhaai, line 54: Here and in the aca version, if we get a result back but it has null datestamps, should we bail?

src/api/covariates/base.py
```diff
@@ -0,0 +1,12 @@
...
+
+class CovariateRequestError(Exception):
+    pass
+
+
+class BaseCovariate:
+    radius = 0.025  # in km
+
+    def fetch(self, points: List[Tuple[float, float]]) -> List[dict]:
+        raise NotImplementedError()
```
saanobhaai, line 12: Any other properties/methods that should be defined here? `_sqkm_to_sqm` maybe? `num_threads`? Wonder also about the repeated `ThreadPoolExecutor` usages, and whether that should be DRYed up in this file...

src/api/covariates/vibrant_oceans.py
```diff
@@ -0,0 +1,157 @@
...
+                                ST_GeomFromText('SRID=4326;POINT(%(x)s %(y)s)')::geography,
+                                %(radius)s
+                            ),
+                            allreef.geom::geography
+                        ),
+                        true
+                    ) /
+                    ST_Area(
+                        ST_Buffer(
+                            ST_GeomFromText('SRID=4326;POINT(%(x)s %(y)s)')::geography,
```
saanobhaai, line 102: Maybe format this and the above instances the same as below, on one line

src/api/resources/sample_units/benthiclitmethod.py
```diff
@@ -1,18 +1,14 @@
 from django.db import transaction
-from django.db.models import Sum
```
saanobhaai, line 2: For all the sample_unit files, there are significant number of deleted imports. I assume that's because they are unused -- just checking...

saanobhaai (changes requested):

src/api/reports/formatters.py
```diff
@@ -120,12 +120,12 @@ def to_percent_cover(value, field, row, serializer_instance):
            f"Average soft coral %: {percent_soft_avg_avg}, Average macroalgae %: {percent_algae_avg_avg}"
 
 
-def _to_aca_covariate(value, covariate_key):
+def to_covariate(value, field, row, serializer_instance):
```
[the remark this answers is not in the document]
gridcell: yeah that's the function argument signature for formatter functions

src/api/covariates/__init__.py
```diff
@@ -36,29 +17,75 @@ def update_site_covariates(site):
...
+        covariate.requested_datestamp = requested_date
+        covariate.value = cov
+        covariate.save()
+
+
+def update_site_covariates(site, force=False):
+    site_pk = site.pk
+
+    point = site.location
+    north_pole = (90, 0)
```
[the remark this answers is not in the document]
gridcell: if they are going to be used somewhere else I would say yeah

src/api/resources/sample_units/benthiclitmethod.py
gridcell: yes

src/api/covariates/__init__.py
```diff
@@ -36,29 +17,75 @@ def update_site_covariates(site):
...
+    south_pole = (-90, 0)
+    existing_site = Site.objects.get_or_none(pk=site_pk)
+
+    if force is False and (
+        (
+            existing_site
+            and existing_site.location == point
```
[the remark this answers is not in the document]
gridcell: This wasn't a supported covariate check but a check to see if the covariate has been collected or not.  Don't think it's worth doing that check anymore since we backfill when we add a new covariate to the system

src/api/covariates/vibrant_oceans.py
gridcell: I think the line length gets pretty long then

src/api/covariates/base.py
gridcell: ok to leave this for now?  Thinking if we do another covariate this way than we do it then, in case it moves to new system

src/api/covariates/__init__.py
```diff
@@ -16,7 +38,10 @@ def _update_site_aca_covariates(point, site_pk):
     result = results[0]
 
     data_date = result["date"]
```
saanobhaai, line 40: If we check `requested_date`, shouldn't we give `data_date` the same treatment?

src/api/covariates/vibrant_oceans.py
saanobhaai: OK no big deal -- but I had to read these a couple times, whereas 109 I got in one glance :-)

saanobhaai (changes requested):

PR #164, opened 2021-08-11
[M325] Django command to update profile picture from Auth0

gridcell:


gridcell:
[Back fill profile pictures for all users](https://trello.com/c/JWUidfqA/325-back-fill-profile-pictures-for-all-users)

src/tools/management/commands/update_profile_picture.py
```diff
@@ -0,0 +1,52 @@
...
+        )
+
+        parser.add_argument(
+            "--profile_id",
+            type=str,
+            help="Only update profile for given profile id",
+        )
+
+    def update_profile_picture(self, profile):
+        auth_user = profile.authusers.all()[0]
```
saanobhaai, line 26: What if the first one doesn't have a `picture`? Should we iterate over `profile.authusers.all()` until we find one?

saanobhaai (changes requested):

PR #165, opened 2021-08-11
[M334] Include email in project profile serializer

gridcell:


gridcell:
[Add email as a read-only field to project profile serializer](https://trello.com/c/2zUPSoob/334-add-email-as-a-read-only-field-to-project-profile-serializer)

PR #166, opened 2021-08-12
[1574] make covariate model field nullable

saanobhaai:


saanobhaai:
[make covar value field nullable](https://trello.com/c/DQt7EzPh/1574-make-covar-value-field-nullable)

PR #167, opened 2021-08-12
[M335] New project resource action to add users to project

gridcell:


gridcell:
[Support pending state when non-user (email) is added to a project](https://trello.com/c/eIHhvo3Y/335-support-pending-state-when-non-user-email-is-added-to-a-project)

src/api/resources/project.py
```diff
@@ -407,3 +407,36 @@ def destroy(self, request, *args, **kwargs):
...
+        email = request.data.get("email")
+        role = request.data.get("role")
+        admin_profile = request.user.profile
+
+        if email is None:
+            raise exceptions.ValidationError(
+                detail={"email": "Email is required"}
+            )
+
+        profile = Profile.objects.get_or_none(email=email)
```
saanobhaai, line 426: Can we not just do `get_or_create`?

src/api/utils/email.py
```diff
@@ -57,3 +62,26 @@ def contact_project_admins(project, subject, body, from_email):
     # lots of checking/validation/antispam here
     # email_project_admins(project, subject, body, from_email)
     pass
+
+
+@run_in_thread
```
saanobhaai, line 67: `run_in_thread` is cool, but should we be sending email synchronously and letting user know when the new profile was added but the email didn't successfully send?

saanobhaai (changes requested):

PR #169, opened 2021-08-17
[M1573] Support null Quadrat benthic percent values

gridcell:


gridcell:
[allow nulls for bleaching QC percent cover observations](https://trello.com/c/QlBm2v0h/1573-allow-nulls-for-bleaching-qc-percent-cover-observations)

src/api/submission/validations.py
```diff
@@ -1075,56 +1075,58 @@ def get_observations(cls, data, key=None):
...
+                return self.error(self.identifier, self.LESS_QUADRAT_NUMBER)
 
-            if (
-                self._validate_percent_value_total(values, "validate_percent_values")
-                == ERROR
-            ):
-                return ERROR
+            quadrat_nums.append(quadrat_num)
+
+        if len(quadrat_nums) != len(set(quadrat_nums)):
```
saanobhaai, line 1128: This tests that the number of quadrat numbers is the same -- but couldn't they be the same and still have a duplicate? I'm probably not seeing the whole logic.

src/api/resources/obs_quadrat_benthic_percent.py
```diff
@@ -1,13 +1,28 @@
...
-        extra_kwargs = {}
+        extra_kwargs = {
+            "quadrat_number": {
+                "error_messages": {
+                    "null": "Quadrat number is required"
+                }
+            }
+        }
+        validators = [
+            UniqueTogetherValidator(
```
saanobhaai, line 20: Might be good to talk this through... not quite getting why we need a special serializer validator in addition to the submission validations (and/or model `unique_together`), or how they interact...

src/api/submission/validations.py
```diff
@@ -1075,56 +1075,58 @@ def get_observations(cls, data, key=None):
...
     VALID_NUMBER_MSG = "Percent cover value not between 0 and 100"
+    LESS_QUADRAT_NUMBER = "Quadrat number less than 1"
+    DUPLICATE_QUADRAT_NUMBERS = "Duplicate quadrat numbers"
 
-    def _check_percent_value(self, value, msg_param):
-        try:
-            _ = int(value)
-        except (TypeError, ValueError):
-            return ERROR, self.VALID_NUMBER_MSG
+    def validate_percent_values(self):
```
saanobhaai, line 1085: OK I think I get this, and it's much better...

src/api/submission/validations.py
gridcell: it's checking the length of the list versus the length of the set of quadrat_nums ?

src/api/resources/obs_quadrat_benthic_percent.py
gridcell: Custom error message wasn't working.  Apparently there was a bug related to this in an earlier version but the version we are currently running should have this fix in it.

PR #170, opened 2021-08-18
[M1570] Change base api viewset to use SafeSearchFilter

gridcell:
SafeSearchFilter extends SearchFilter by including validation of regex search terms

### DIFF
--- a/src/api/resources/base.py
+++ b/src/api/resources/base.py
@@ -1,4 +1,6 @@
+import re
 import uuid
+
 from django.utils.translation import ugettext_lazy as _
 from django.db.models.sql.constants import ORDER_PATTERN
 from rest_framework import serializers, viewsets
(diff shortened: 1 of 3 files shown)


gridcell:
[/sites/ search regex 500 error](https://trello.com/c/uE9ETq2V/1570-sites-search-regex-500-error)

src/api/resources/base.py
```diff
@@ -446,6 +448,30 @@ def remove_invalid_fields(self, queryset, fields, view, request):
         ]
 
 
+class SafeSearchFilter(SearchFilter):
+
+    def _check_search_terms(self, search_fields, search_terms):
+        if (
+            not search_fields
+            or not search_terms
+            or "$" not in [sf[0] for sf in search_fields]
```
saanobhaai, line 457: Is this the only thing wrong with the test query? Is there anything else we should be protecting against?

src/api/resources/base.py
```diff
@@ -530,7 +556,7 @@ class BaseApiViewSet(MethodAuthenticationMixin, viewsets.ModelViewSet, UpdatesMi
 
     pagination_class = StandardResultPagination
 
-    filter_backends = (DjangoFilterBackend, RelatedOrderingFilter, SearchFilter)
+    filter_backends = (DjangoFilterBackend, RelatedOrderingFilter, SafeSearchFilter)
```
saanobhaai, line 559: Are there any endpoints to worry about that somehow don't inherit from `BaseApiViewSet`?

src/api/resources/base.py
gridcell: I wanted to address the immediate need it can always built out (more safety)

src/api/resources/base.py
gridcell: SearchFilter not being used anywhere else

PR #171, opened 2021-08-18
[M1570] Change Site and Management endpoints to be authenticated read-only

gridcell:


gridcell:
[make site and mr endpoints require authentication (still read-only)](https://trello.com/c/XOWvsi7h/1575-make-site-and-mr-endpoints-require-authentication-still-read-only)

PR #176, opened 2021-09-08
[M1519c] Speed up sample unit method CSV report generation

gridcell:
- include latitude/longitude in base sql call instead of using location and formatting
- not using qs.values() to prepare query set for report serializer


gridcell:

PR #178, opened 2021-09-23
[M364] Handle empty strings and null values when validating collect records

gridcell:


gridcell:
[Handle sending nulls?](https://trello.com/c/6C926NVW/364-handle-sending-nulls)

src/api/submission/validations.py
```diff
@@ -88,7 +89,7 @@ def validate_region(self):
         region = str(regions[0].pk)
 
         observations = self.get_observations(self.data, key=self.observations_key)
-        attribute_ids = list(set(obs.get(self.attribute_key) for obs in observations))
+        attribute_ids = list({obs.get(self.attribute_key) for obs in observations if obs.get(self.attribute_key) is not None})
```
saanobhaai, line 92: Is this dict technique preferable to the `set()` pattern?

src/api/submission/validations.py
```diff
@@ -153,54 +154,40 @@ class FishAttributeMixin(RegionalAttributesMixin):
...
-            fish_size_bin_lookup = dict()
+            fish_size_bin_lookup = {}
 
         # only validate lengths at species level until we know how to aggregate to genus
-        fish_attrs = (
-            FishSpecies.objects.filter(id__in=fish_attr_ids)
-            .select_related("genus__name")
-            .values("id", "genus__name", "name", "max_length")
-        )
+        fish_attrs = {i["id"]: i for i in FishSpecies.objects.filter(id__in=fish_attr_ids).select_related("genus__name").values("id", "genus__name", "name", "max_length")}
```
saanobhaai, line 170: Looks like we need to reapply Black?

src/api/submission/validations.py
```diff
@@ -670,7 +658,7 @@ def validate_total_length(self):
         benthic_transect = self.data.get("benthic_transect") or {}
         # Convert to cm
         transect_length = (benthic_transect.get("len_surveyed") or 0.0) * 100
-        obs_len = sum([ob.get("length") or 0.0 for ob in obs])
+        obs_len = sum(ob.get("length") or 0.0 for ob in obs)
```
saanobhaai, line 661: This works? Doesn't have to be a list comprehension?

src/api/submission/validations.py
```diff
@@ -853,7 +837,14 @@ def validate_observation_density(self):
 
     def validate_fish_count(self):
         obs = self.data.get("obs_belt_fishes") or []
-        num_fish = sum([obs.get("count") or 0 for obs in obs])
+        counts = []
+        for ob in obs:
+            try:
+                counts.append(int(ob.get("count") or 0))
```
saanobhaai, line 843: Shouldn't we use `cast_int` here?
Also: the previous comprehension approach would still work, wouldn't it? The idea here is to specifically catch invalid values?

src/api/submission/validations.py
```diff
@@ -875,7 +866,7 @@ def validate_fish_family_subset(self):
         if isinstance(fish_family_subset, list) is False:
             return self.ok(self.identifier)
 
-        fish_attribute_ids = set([ob.get("fish_attribute") for ob in obs])
+        fish_attribute_ids = {ob.get("fish_attribute") for ob in obs}
```
saanobhaai, line 869: Again I'm unfamiliar... this is a dict comprehension with keys only? Where do the vals come from?

src/api/tests/fixtures/collect_records.py
```diff
@@ -134,32 +134,32 @@ def valid_collect_record(
 ):
     observations = [
         dict(
-            count=10,
+            count="10",
```
saanobhaai, line 137: By making these strings, we're testing both the casting and the resulting values, is that it?

saanobhaai (changes requested):
Just a couple questions

PR #179, opened 2021-09-28
[M1519d] Pull from different database vibrant ocean fields

gridcell:
- covariate keys and display names have been renamed
- for some of the covariates we are now pulling the "raw" value instead


gridcell:

PR #180, opened 2021-09-28
[M1573b] Updates to bleaching ingest and validation

gridcell:
- change validation message when bleaching percent values are null
- during bleaching ingest, ensure the quadrat numbers are continuous


gridcell:
[allow nulls for bleaching QC percent cover observations](https://trello.com/c/QlBm2v0h/1573-allow-nulls-for-bleaching-qc-percent-cover-observations)

src/api/ingest/bleaching_serializer.py
```diff
@@ -57,6 +57,15 @@ def benthic_attributes_choices():
     ]
 
 
+class BleachingCSVListSerializer(CollectRecordCSVListSerializer):
+    def create_hook(self, grouped_records):
```
saanobhaai, line 61: Couldn't you just override `group_records`, and then avoid `output = self.create_hook(output)` ?

saanobhaai (changes requested):

PR #181, opened 2021-09-28
M1572 email template redesign

alanjleonard:


saanobhaai:
[redesign emails](https://trello.com/c/wjJ54F2h/1572-redesign-emails)

PR #184, opened 2021-09-29
[M1584] Move serializer validation so it's run first during validation

gridcell:
- **BIG CHANGE ALERT:** if the serializer validation fails, no further validations are run and a response is returned

### DIFF
--- a/src/api/submission/protocol_validations.py
+++ b/src/api/submission/protocol_validations.py
@@ -1,4 +1,5 @@
 import datetime
+from logging import raiseExceptions
 
 import dateutil
 from django.utils.translation import ugettext_lazy as _
@@ -25,6 +26,12 @@ from .validations import (
     ValueInRangeValidation,
 )
 from ..utils import cast_float, cast_int
+
+
+class SerializeValidationError(Exception):
+    def __init__(self, message="", errors=None):
+        super().__init__(message)
+        self.errors = errors
 
 
 class ProtocolValidation(object):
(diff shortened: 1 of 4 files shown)


gridcell:
[date format 500](https://trello.com/c/r4Bw6n9P/1584-date-format-500)

src/api/submission/protocol_validations.py
```diff
@@ -243,9 +255,12 @@ class FishBeltProtocolValidation(TransectValidation):
     SAMPLE_UNIT = "fishbelt_transect"
 
     def validate(self):
-        results = [super(FishBeltProtocolValidation, self).validate()]
+        try:
```
saanobhaai, line 258: Would it be better/possible to put this in the `validate` method of both `TransectValidation` and `QuadratValidation`, and then these calls in individual protocol validations wouldn't have to change?:

```
        try:
            return [super().validate()]
        except SerializeValidationError:
            return ERROR
```

src/api/submission/writer.py
```diff
@@ -77,27 +81,26 @@ def _create_sample_event(self, sample_event_data):
...
+        }
+        se = SampleEvent.objects.filter(**query_params)
+        if se.exists():
+            sample_event = se[0]
+            notes = sample_event_data.get("notes") or ""
+            if notes.strip():
+                sample_event.notes += "\n\n{}".format(notes)
+                sample_event.save()
+            return sample_event
+        return serializer.save()
```
saanobhaai, line 103: ok so with this refactor we should be able to delete the whole `_create_sample_event` method, right?

saanobhaai (changes requested):
Two small refactor questions

PR #185, opened 2021-09-29
[M378] Add more logic to Profile's full_name property

gridcell:
- possible returned values in order
1. First and last name
2. Pending using (if no AuthUser exists for the profile)
3. Email pulled from the AuthUser user_id

### DIFF
--- a/src/api/models/base.py
+++ b/src/api/models/base.py
@@ -55,21 +55,15 @@ class Profile(models.Model):
 
     @property
     def full_name(self):
-        name = []
-        if self.first_name:
-            name.append(self.first_name)
-
-        if self.last_name:
-            name.append(self.last_name)
-
-        if len(name) > 0:
-            return ' '.join(name)
+        if self.first_name and self.last_name:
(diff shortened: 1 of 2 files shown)


gridcell:
[Change "non-mermaid users" with profile_name as ( pending user ) instead of "None None"](https://trello.com/c/QR0xQztB/378-change-non-mermaid-users-with-profilename-as-pending-user-instead-of-none-none)

src/api/models/base.py
```diff
@@ -55,21 +55,15 @@ def __str__(self):
...
-        name = []
-        if self.first_name:
-            name.append(self.first_name)
-
-        if self.last_name:
-            name.append(self.last_name)
-
-        if len(name) > 0:
-            return ' '.join(name)
+        if self.first_name and self.last_name:
```
saanobhaai, line 58: I'm thinking this shouldn't be `and` but rather `or`, and then use the existing dev logic. So a person with only first or last name populated would still count.

saanobhaai (changes requested):

PR #188, opened 2021-10-05
[M383] Create new project profile revisions when related models change

gridcell:


gridcell:
[When transferring collect records to another user, the project profile's num_active_sample_units isn't update in the endpoint](https://trello.com/c/XV9EIAt4/383-when-transferring-collect-records-to-another-user-the-project-profiles-numactivesampleunits-isnt-update-in-the-endpoint)

PR #195, opened 2021-11-01
[M1598] Belt Fish observation required validator

gridcell:


gridcell:
[Fishbelt observations required validators](https://trello.com/c/tKQA1uoA/1598-fishbelt-observations-required-validators)

PR #196, opened 2021-11-02
[M1599] Region validator

gridcell:


gridcell:
[Region validator](https://trello.com/c/jKsRcRif/1599-region-validator)

PR #200, opened 2021-12-13
[M527] Include observation id to validator results

gridcell:
"observation_id" is added to validator result context


gridcell:
[Add observation id to observation validation results](https://trello.com/c/MSXcthW7/529-add-observation-id-to-observation-validation-results)

PR #202, opened 2021-12-15
Speed improvements for benthic and fish species pull/endpoints

gridcell:


src/api/resources/fish_species.py
```diff
@@ -67,9 +74,16 @@ class FishSpeciesViewSet(BaseAttributeApiViewSet):
     serializer_class = FishSpeciesSerializer
     queryset = (
         FishSpecies.objects.select_related()
-        .extra(select={"display_name": "fish_genus.name || ' ' || fish_species.name"})
-        .prefetch_related(Prefetch("regions", queryset=Region.objects.all().only("id")))
+        .annotate(
+            regions_=ArrayAggExt(
+                "regions",
+                # filter=Q(regions__isnull=False),
```
saanobhaai, line 80: remove commented

src/api/resources/base.py
```diff
@@ -720,3 +721,27 @@ def partial_update(self, request, pk=None):
 
     def destroy(self, request, pk=None):
         raise MethodNotAllowed("DELETE")
+
+
+class ArrayAggExt(ArrayAgg):
+    template = "ARRAY_REMOVE(%(function)s(%(distinct)s%(expressions)s %(ordering)s), NULL)"
```
saanobhaai, line 727: Just to confirm, I'm looking at the [source](https://github.com/django/django/blob/main/django/contrib/postgres/aggregates/general.py#L45) and wanting to make sure: currently if a fish species is not associated with any region, we get `[]`, right? This new approach will ensure we keep getting that, and not `[None]`, right? But the source code indicates that by passing `default=None` we could get what will be default in Django 5, i.e. `None`. Would the latter not work for us, thereby making the separate `regions_` source unnecessary? I get why we don't want `[None]`, but wouldn't `None` work?

src/api/resources/fish_species.py
```diff
@@ -1,13 +1,20 @@
...
+    BaseAPISerializer,
+    RegionsSerializerMixin,
+)
 from .mixins import CreateOrUpdateSerializerMixin
-from ..models import FishSpecies, Region
+from ..models import FishSpecies
 
 
-class FishSpeciesSerializer(CreateOrUpdateSerializerMixin, BaseAPISerializer):
+class FishSpeciesSerializer(RegionsSerializerMixin, CreateOrUpdateSerializerMixin, BaseAPISerializer):
```
saanobhaai, line 17: I've never been clear on mixin order rules...

src/api/resources/fish_species.py
```diff
@@ -67,9 +74,16 @@ class FishSpeciesViewSet(BaseAttributeApiViewSet):
...
         FishSpecies.objects.select_related()
-        .extra(select={"display_name": "fish_genus.name || ' ' || fish_species.name"})
-        .prefetch_related(Prefetch("regions", queryset=Region.objects.all().only("id")))
+        .annotate(
+            regions_=ArrayAggExt(
+                "regions",
+                # filter=Q(regions__isnull=False),
+                # default=Value([])
+            ),
+            display_name=Concat(F("genus__name"), Value(" "), F("name"))
```
saanobhaai, line 83: We still have to deal with `FishGenus` and `FishFamily` right?

saanobhaai (changes requested):

src/api/resources/base.py
gridcell: yes

src/api/resources/fish_species.py
gridcell: didn't bother, it seemed reasonably quick

PR #207, opened 2022-01-04
[M1616] docker_compose_volumes: add persistent volume to postgres service in docker compose

gridcell:


gridcell:
[Persist postgres volume for local development](https://trello.com/c/AYsopFLN/1616-persist-postgres-volume-for-local-development)

docker-compose.yml
```diff
@@ -1,10 +1,15 @@
...
+volumes:
+  pg_data:
+
 services:
 
   api_db:
     container_name: api_db
     image: mdillon/postgis:11-alpine
+    volumes:
+      - pg_data:/var/lib/postgresql/data
```
saanobhaai, line 12: Not sure I understand this... Is this the default location for a linux postgres install, and if so wouldn't it potentially step on whatever else a developer might have there?

gridcell: Should mount the container directory (and contents) outside in the host

PR #208, opened 2022-01-19
M1618: skip updates summary sample events and sites when not prod env. - includes option to force update - DJ Management commands

gridcell:


gridcell:
[don't run site/SE summary updates unless on prod (or if pass force argument)](https://trello.com/c/NKwwaq4E/1618-dont-run-site-se-summary-updates-unless-on-prod-or-if-pass-force-argument)

PR #209, opened 2022-01-19
[M1619] Make summary commands more effcient

gridcell:
Only update projects that have been updated rather than all


src/api/signals/project_edits.py
```diff
@@ -0,0 +1,48 @@
...
+]
+
+
+def get_all_project_ids():
+    qs = Project.objects.filter(status__in=[Project.OPEN, Project.LOCKED])
+    return [p.pk for p in qs]
+
+
+def update_project_changes(instance, *args, **kwargs):
+    project = get_related_project(instance)
```
saanobhaai, line 22: Scenario: user syncs 100 SUs, each with 100 observations, after long period offline. db hits would be something like this?:
- 4 hits per observation to find project id, 3 per SU, 2 per SE, 1 per site
- 1 hit per each of above to get `EditedProjects`
- 1 hit to insert the project pk
Right? So something like (100 * 100 * 4) + (100 * 3) + (40 * 2) + (40 * 1) + (100 * 100) + 100 + 40 + 40 + 1 = 50601?

I know we decided we couldn't use `Revision` to track what's changed, but what about using the `push` endpoint to extract project ids from the request body instead of doing db calls? Hmm I guess looking there, we'd still have to do the db calls to look up... Well, dang.

saanobhaai (changes requested):
Worth a talk-through...

PR #210, opened 2022-01-20
M538: handle invalid site id in sample_date validator

gridcell:


gridcell:
[empty-string uuid /validate 500](https://trello.com/c/qBksVUFM/538-empty-string-uuid-validate-500)

src/api/submission/validations2/validators/sample_date.py
```diff
@@ -33,6 +36,9 @@ def is_future_sample_date(self, date_str, time_str, site_id):
         tf = TimezoneFinder()
         tz_str = tf.timezone_at(lng=x, lat=y) or "UTC"
         tzinfo = tz.gettz(tz_str)
+
+        date_str = date_str or ""
```
saanobhaai, line 40: Does `parse_datetime` return None with these defaults?

gridcell: if empty string parse_datetime returns `None`

PR #213, opened 2022-02-03
[1626] extend backup for scheduled execution

saanobhaai:


saanobhaai:
[server db backups](https://trello.com/c/ljXar4a9/1626-server-db-backups)

src/api/utils/__init__.py
```diff
@@ -177,3 +176,28 @@ def cast_int(val):
...
+    data, err = proc.communicate(input=std_input)
+
+    if to_file is not None:
+        with open(to_file, "w") as f:
+            f.write("DATA: \n")
+            f.write(str(data))
+            f.write("ERR: \n")
+            f.write(str(err))
+    else:
+        print(data)
```
gridcell, line 202: do you want to print these or return them?

saanobhaai: I dunno, I guess it seems to me these should print. I see this function as not doing something generic that would be captured by something else, but just running code in a different process. So I wouldn't want to assign its invocation to a variable and then print it every time, but rather print to stdout just as anything else would. But I'm probably missing something here.

PR #215, opened 2022-02-08
[M539] Validator fix: when fish attr doesn't exist in family subset

gridcell:


gridcell:
[500 on fish family lookup of `None` fish attribute id](https://trello.com/c/1O9eIych/539-500-on-fish-family-lookup-of-none-fish-attribute-id)

PR #216, opened 2022-02-14
Different approach for update site and sample event summaries

gridcell:


gridcell:
[only update sites/SEs for projects that need it](https://trello.com/c/FA7DitHm/1619-only-update-sites-ses-for-projects-that-need-it)

src/api/models/mermaid.py
```diff
@@ -775,6 +780,10 @@ def __str__(self):
     def profile_name(self):
         return u'{} {}'.format(self.profile.first_name, self.profile.last_name)
 
+    # @property
```
saanobhaai, line 783: No longer needed?

src/api/models/mermaid.py
```diff
@@ -756,6 +757,10 @@ def sample_unit(self):
     def __str__(self):
         return str(_(u'transect method'))
 
+    @property
```
saanobhaai, line 760: Is this needed anymore? I can't see a reference, but maybe it's there.

saanobhaai (changes requested):
Two small questions

src/api/models/mermaid.py
gridcell: Yeah it gets used in the signals...

```
@receiver(post_delete, sender=TransectMethod)
def update_summaries_on_delete_transect_method(sender, instance, *args, **kwargs):
...
```

PR #218, opened 2022-02-25
[M1630] Normalize covariate responses when  no covariates are found for a location

gridcell:


gridcell:
[Don't return value for empty ACA covariate results](https://trello.com/c/A0eMfNhl/1630-dont-return-value-for-empty-aca-covariate-results)

PR #219, opened 2022-02-25
M1634: Change check for existence of new project profile

gridcell:


gridcell:
[500 error when adding profile that already exists in a project](https://trello.com/c/JeOMIfmU/1634-500-error-when-adding-profile-that-already-exists-in-a-project)

src/api/tests/test_project.py
```diff
@@ -0,0 +1,37 @@
...
+
+    response = client.post(
+        url,
+        data={"email": "bill@test.com", "role": ProjectProfile.COLLECTOR},
+        HTTP_AUTHORIZATION=f"Bearer {token1}",
+    )
+    assert response.status_code == 200
+
+
+def test_add_profile_new(
```
saanobhaai, line 23: This (overloading) isn't legal in python, is it?
https://stackoverflow.com/a/5079643

PR #222, opened 2022-03-09
[M1623] Refactor duplication sample unit validation for fishbelt transect

gridcell:


gridcell:
[Refactor duplication sample unit validation](https://trello.com/c/YX0zFbeK/1623-refactor-duplication-sample-unit-validation)

src/api/submission/validations2/validators/fishbelt_transect.py
```diff
@@ -4,60 +4,63 @@
...
         try:
             number = int(number)
             check_uuid(width)
             check_uuid(site)
             check_uuid(management)
-            if relative_depth is not None:
-                check_uuid(relative_depth)
+            float(depth)
+
             if parse_datetime(f"{sample_date} 00:00:00") is None:
```
saanobhaai, line 57: While we're at it, should we maybe change this to use `parse_date` without the 0s, for less chance of returning something other than `None`?

saanobhaai (commented):
Small additional suggestion

PR #225, opened 2022-03-27
Rearrange summary sql joins

gridcell:


src/api/utils/summaries.py
```diff
@@ -1,17 +1,26 @@
 from ..models import (
+    Project,
     SummarySampleEventModel,
     SummarySampleEventSQLModel,
     SummarySiteModel,
     SummarySiteSQLModel,
 )
 
+project_statuses = [Project.OPEN, Project.LOCKED]
```
saanobhaai, line 9: Is this used in this file? Looks like only used in command, and this and the `Project` import could be removed here.

src/api/signals/__init__.py
```diff
@@ -4,12 +4,12 @@
...
 from django.core import serializers
 from django.core.cache import cache
-from django.db.models.signals import post_delete, post_save, pre_delete, pre_save, m2m_changed
+from django.db.models.signals import post_delete, post_save, pre_save, m2m_changed
 from django.dispatch import receiver
 
-from . import revision
-from . import summaries
-from ..covariates import update_site_covariates
+from .revision import * # noqa
```
saanobhaai, line 10: additional space before comment

saanobhaai (changes requested):
Two trivial comments; in addition, I'm assuming the re-addition of threading is a temporary step on the way to simpleq. But all the sql changes look right to me so this is good to merge otherwise.

PR #226, opened 2022-03-29
[M1643] Run summary updates in simpleq worker

gridcell:
- Added another container for elasticmq (SQS clone) for local testing
- Send summary updates that were being called in signals to simpleq jobs

### DIFF
--- a/src/api/signals/__init__.py
+++ b/src/api/signals/__init__.py
@@ -7,8 +7,8 @@ from django.core.cache import cache
 from django.db.models.signals import post_delete, post_save, pre_save, m2m_changed
 from django.dispatch import receiver
 
-from .revision import * # noqa
-from .summaries import * # noqa
+from .revision import *  # noqa
+from .summaries import *  # noqa
 from ..covariates import update_site_covariates_threaded
 from ..models import *
 from ..resources.sync.views import (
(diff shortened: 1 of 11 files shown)


gridcell:
[Run summary site and sample event updates in SimpleQ](https://trello.com/c/QF3DtWuB/1643-run-summary-site-and-sample-event-updates-in-simpleq)

src/simpleq/queues.py
```diff
@@ -65,12 +66,16 @@ def __init__(self, name, sqs_resource=None):
...
         else:
-            self.sqs_resource = boto3.resource(
-                "sqs",
-                aws_access_key_id=os.environ["AWS_ACCESS_KEY_ID"],
-                aws_secret_access_key=os.environ["AWS_SECRET_ACCESS_KEY"],
-                region_name=os.environ["AWS_REGION"],
-            )
+            resource_args = {
+                "service_name": "sqs",
+                "aws_access_key_id": os.environ["AWS_ACCESS_KEY_ID"],
```
saanobhaai, line 71: Should we be getting these from settings, rather than directly from `os.environ`?

src/api/signals/__init__.py
```diff
@@ -8,7 +8,7 @@
 from django.dispatch import receiver
 
 from .revision import * # noqa
-from .summaries import *  # noqa
+from .summaries import * # noqa
```
saanobhaai, line 11: 2 spaces before inline comments is PEP no?

saanobhaai (changes requested):
Two trivial questions; also there's a conflict.

PR #228, opened 2022-04-06
SQL model sample unit performance changes

gridcell:


saanobhaai:
[slow authenticated aggregate queries](https://trello.com/c/bGXPKDPb/1649-slow-authenticated-aggregate-queries)

PR #232, opened 2022-04-14
[M620] Add "label" to Sample Unit Method Endpoint

gridcell:
Label was missing from the sample unit method endpoint and is needed to create consistency between collect record and submitted record tables in UI

- Expose label in the sample unit endpoint response


gridcell:
[Add label to sample unit endpoint](https://trello.com/c/DUgKzBoC/620-abc)

JBurkinshaw (approved):

PR #237, opened 2022-05-03
[M1608-1] New validators for Benthic Photo Quadrat Transect

gridcell:


gridcell:
[Benthic Photo Quadrat Transect validations](https://trello.com/c/KNQQHTsu/1608-benthic-photo-quadrat-transect-validations)

PR #239, opened 2022-05-04
[M1654] Benthic Photo Quadrat Transect submission

gridcell:


src/api/submission/writer.py
```diff
@@ -153,9 +153,15 @@ def get_or_create_fishbelt_transect(self, sample_event_id):
...
 
-    def get_or_create_beltfish(self, fishbelt_transect_id, sample_unit_method_id=None):
+    def get_or_create_beltfish(
+        self, collect_record_id, fishbelt_transect_id, sample_unit_method_id=None
+    ):
         beltfish_data = {"transect": fishbelt_transect_id, "id": sample_unit_method_id}
-        return self.get_or_create(BeltFish, BeltFishSerializer, beltfish_data)
+        record = self.get_or_create(BeltFish, BeltFishSerializer, beltfish_data)
+        record.collect_record_id = collect_record_id
+        record.save()
```
saanobhaai, line 162: Are we adding `collect_record_id` to support SU->CR conversion? I thought we already did that for fishbelt?

Can we not avoid saving twice by setting `collect_record_id` before doing `get_or_create`? I worry about signals and whatnot...

gridcell: Agreed, pushed up a change

src/api/submission/writer.py
```diff
@@ -69,13 +70,18 @@ def validate_data(self, serializer_cls, data):
...
+    def get_or_create(self, model, serializer_cls, data, additional_data=None):
+        pk = data.get("id") or uuid.uuid4()
+        data["id"] = pk
         serializer = self.validate_data(serializer_cls, data)
         try:
             data.pop("id")
             return model.objects.get(**data)
         except model.DoesNotExist:
+            if isinstance(additional_data, dict):
+                data["id"] = pk
```
saanobhaai, line 82: Redundant, no?

saanobhaai (changes requested):

src/api/submission/writer.py
gridcell: No b/c it gets stripped above the except

PR #246, opened 2022-05-20
[M651] New notification model and endpoint

gridcell:
Card: [M651](https://trello.com/c/Mf9DIlIS/651-bell-card-models-and-api-endpoints)


gridcell:
[Bell card models and api endpoints](https://trello.com/c/Mf9DIlIS/651-bell-card-models-and-api-endpoints)

src/api/models/mermaid.py
```diff
@@ -1738,3 +1738,24 @@ class AuditRecord(JSONMixin):
...
+    ERROR = "error"
+    
+    STATUSES = (
+        (INFO, INFO),
+        (WARNING, WARNING),
+        (ERROR, ERROR),
+    )
+
+    title = models.CharField(max_length=200)
+    status = models.CharField(max_length=10, choices=STATUSES)
```
saanobhaai, line 1755: Just making sure we don't want to store ordinal values, to allow for ranking, like we do for project status?

src/api/models/mermaid.py
```diff
@@ -1738,3 +1738,24 @@ class AuditRecord(JSONMixin):
...
+    description = models.TextField(null=True, blank=True)
```
saanobhaai, line 1756: Just wondering how this would translate into the types of cards shown on https://trello.com/c/mWx3Nbhf/357-bell-notifications-bell-cards-for-discussion

- the first example I guess we're not doing in the bell anymore
- the second could use this field; this assumes that the app will detect a version upgrade and refresh itself though. Do we have a card for that?
- the third looks like an amalgamation of several `Notification` instances; I'm not sure this is quite right. But if it was, how would we generate "go to project"?

Basically, I'm wondering if the frontend will need some kind of `context` object; maybe the `data` field on `BaseModel` will serve this purpose.

src/api/models/mermaid.py
```diff
@@ -1738,3 +1738,24 @@ class AuditRecord(JSONMixin):
     event_by = models.ForeignKey("Profile", on_delete=models.SET_NULL, null=True, blank=True)
     model = models.CharField(max_length=100)
     record_id = models.UUIDField(db_index=True)
+
+
+class Notification(BaseModel):
```
saanobhaai, line 1743: So `created_on` in `BaseModel` would be used for the datestamp, right?

saanobhaai (changes requested):

src/api/models/mermaid.py
gridcell: yeah

src/api/models/mermaid.py
gridcell: Think of description as storing the same content as a WYSIWYG.  The description will support "markup"`

PR #249, opened 2022-06-15
[M1670] Hotfix: Site/Sample Event Summary updates

gridcell:
Need to remove summary records from projects that have been converted from open/locked to test


gridcell:
[During summary sample event handle case when projects move from open/locked --> test](https://trello.com/c/gKymxwqT/1670-during-summary-sample-event-handle-case-when-projects-move-from-open-locked-test)

PR #250, opened 2022-06-15
[M1670] Hotfix: Site/Sample Event Summary updates

gridcell:


gridcell:
[During summary sample event handle case when projects move from open/locked --> test](https://trello.com/c/gKymxwqT/1670-during-summary-sample-event-handle-case-when-projects-move-from-open-locked-test)

src/api/utils/summaries.py
```diff
@@ -7,40 +9,46 @@
...
+    update_project_summary_site(project_id, skip_test_project)
+    update_project_summary_sample_event(project_id, skip_test_project)
 
 
 def update_project_summary_site(project_id, skip_test_project=True):
     if (
         skip_test_project
         and Project.objects.filter(pk=project_id, status=Project.TEST).exists()
     ):
+        SummarySiteModel.objects.filter(project_id=project_id).delete()
```
saanobhaai, line 22: Why not just do this once, before doing anything else, and then not repeat it below?

src/tools/management/commands/update_summaries.py
```diff
@@ -23,12 +23,12 @@ def add_arguments(self, parser):
...
-        update_project_summaries(project_id)
+    def update_summaries(self, project_id, skip_test_project):
+        update_project_summaries(project_id, skip_test_project)
 
     def handle(self, *args, **options):
         is_forced = options["force"]
-        include_test_projects = options["test_projects"]
+        skip_test_project = options["test_projects"] is not True
 
         if settings.ENVIRONMENT != "prod" and is_forced is False:
```
saanobhaai, line 33: Can we add `dev` here? That was the original ticket, i.e. making sure dashboard updates happened on dev as well as prod.

saanobhaai (changes requested):

src/api/utils/summaries.py
gridcell: delete has to happen after the sql table query, which doesn't happen in this spot

PR #261, opened 2022-06-28
fix deleting sample events

gridcell:
- includes test for editing photo quadrat


src/api/resources/mixins.py
```diff
@@ -244,4 +244,5 @@ def edit(self, request, project_pk, pk):
             )
             return Response({"id": str(collect_record.pk)})
         except Exception as err:
+            print(err)
```
saanobhaai, line 247: Still needed?

saanobhaai (approved):

src/api/resources/mixins.py
gridcell: wanted to keep that in there for logging

PR #262, opened 2022-06-30
[M711] Fix revision updates when transferring collect records to another user

gridcell:
Card [#711](https://trello.com/c/CviXxiz9/711-update-current-user-crs-after-transferring-them-away)


gridcell:
[update current user CRs after transferring them away](https://trello.com/c/CviXxiz9/711-update-current-user-crs-after-transferring-them-away)

PR #264, opened 2022-07-12
[M1621] Endpoint to dump ingest csv templates

gridcell:
Card [#1621](https://trello.com/c/5cETZqGn/1621-create-endpoints-for-outputting-blank-ingestion-csvs-headers)


gridcell:
[create endpoints for outputting blank ingestion csvs (headers)](https://trello.com/c/5cETZqGn/1621-create-endpoints-for-outputting-blank-ingestion-csvs-headers)

src/api/resources/collect_record.py
```diff
@@ -228,3 +232,27 @@ def ingest(self, request, project_pk, *args, **kwargs):
             return Response(errors, status=400)
 
         return Response(CollectRecordSerializer(records, many=True).data)
+
+    @action(
+        detail=False,
+        methods=["GET"],
+        permission_classes=[ProjectDataAdminPermission],
+        url_path="ingest_schema/(?P<sample_unit>\w+)",
```
saanobhaai, line 240: Should we append `/csv/` so that we're consistent with the way other endpoints do this, and to allow for a json response without the `/csv/` if we decide to do that?

saanobhaai (changes requested):

PR #266, opened 2022-08-07
[M1644] fix SE averaging for biomass trophic group and family breakouts

saanobhaai:


saanobhaai:
[Handle nonexistent trophic groups in SE biomass aggregation as 0s](https://trello.com/c/TIeC9Vgt/1644-handle-nonexistent-trophic-groups-in-se-biomass-aggregation-as-0s)

PR #270, opened 2022-08-17
[M684]  Fix projects page refreshing

gridcell:
- Return list of record ids for a source type that need to be removed if the user gets 403 when calling `/pull`
- Trigger project revision updates when project profiles are deleted (signal)

Card [#684](https://trello.com/c/tfQCgp1p/684-projects-page-refreshing)


gridcell:
[projects page refreshing](https://trello.com/c/tfQCgp1p/684-projects-page-refreshing)

PR #278, opened 2022-09-07
Onboarding cleanup

aspcanada:
- Allow recaptcha testing on local machine w/o secret key
- add db healthcheck and remove unnecessary sleep
- separate env variables from secrets and update README - this will help with identifying which env vars and secrets are to be attached to the docker container when we move to ECS.
- add make build (with cache)


aspcanada:
This branch is already contained in #282

PR #287, opened 2022-09-25
[M1682] Benthic photo quadrat transect v1 validations

gridcell:


gridcell:
[v1 photo quadrat transect UI using v2 validations](https://trello.com/c/fLMo7kWf/1682-v1-photo-quadrat-transect-ui-using-v2-validations)

PR #289, opened 2022-09-29
[M763] Add copy resource to management and site viewsets

gridcell:


gridcell:
[Create copy site api endpoint](https://trello.com/c/nn6SRwCc/763-create-copy-site-api-endpoint)

PR #292, opened 2022-10-07
[M1721] fix Monocanthidae

saanobhaai:


src/api/migrations/0005_auto_20221006_2037.py
```diff
@@ -0,0 +1,25 @@
...
+
+from django.db import migrations
+
+
+def fix_mono(apps, schema_editor):
+    FishFamily = apps.get_model("api", "FishFamily")
+    FishGenus = apps.get_model("api", "FishGenus")
+
+    correct_family = FishFamily.objects.get(name="Monacanthidae")
+    incorrect_family = FishFamily.objects.get(name="Monocanthidae")
```
gridcell, line 11: should handle the case if the migration get's ran more than once (i.e. incorrect_family doesn't exist)

gridcell (approved):

PR #299, opened 2022-10-24
[M769] return any protected SU methods when pushing site/mr delete

saanobhaai:


saanobhaai:
[return SU objects using site/mr when attempting to delete site/mr](https://trello.com/c/WqyltMf5/769-return-su-objects-using-site-mr-when-attempting-to-delete-site-mr)

gridcell (approved):

PR #307, opened 2022-11-16
[M1592] benthic PIT v2 validations

saanobhaai:


saanobhaai:
[Port Benthic PIT validations to v2](https://trello.com/c/3E3k8yPn/1592-port-benthic-pit-validations-to-v2)

PR #312, opened 2022-11-25
[M1593] v2 benthic LIT validations

saanobhaai:


saanobhaai:
[Port Benthic LIT validations to v2](https://trello.com/c/AIzW4MyF/1593-port-benthic-lit-validations-to-v2)

gridcell (approved):

PR #313, opened 2022-11-25
[M1720] Validate value when filtering by projects

gridcell:
Site and management endpoints allow excluding records from projects.   This fix includes validating query parameters are uuid else 400 error


gridcell:
[sites endpoint country query 500](https://trello.com/c/wLsOl9BC/1720-sites-endpoint-country-query-500)

PR #316, opened 2022-12-05
[M848] Don't test for site validity in sample_date validation

saanobhaai:


saanobhaai:
[errors showing up on wrong field.](https://trello.com/c/zDz3Nk8c/848-errors-showing-up-on-wrong-field)

gridcell (approved):

PR #318, opened 2022-12-07
[M1741] switch to v2 validations for ingestion

saanobhaai:


saanobhaai:
[make ingest work with v2 validations](https://trello.com/c/joP7RPfL/1741-make-ingest-work-with-v2-validations)

gridcell (approved):

PR #321, opened 2022-12-13
[M1747] Summary XLSX Report

gridcell:
XLSX report Includes observation, sample unit and sample event summaries  (one per worksheet) for a protocol


gridcell:
[excel route for project obs/su/se](https://trello.com/c/mcxQCiGX/1747-excel-route-for-project-obs-su-se)

src/api/resources/mixins.py
```diff
@@ -260,6 +264,31 @@ def edit(self, request, project_pk, pk):
             return Response(str(err), status=500)
 
 
+class SampleUnitMethodSummaryReport(object):
+    @action(
+        detail=False, methods=["GET"], permission_classes=[ProjectDataAdminPermission]
+    )
+    def xlsx(self, request, project_pk):
+        from ..utils.reports import create_sample_unit_method_summary_report
```
saanobhaai, line 272: Any reason not to put this up top?

src/api/resources/mixins.py
```diff
@@ -260,6 +264,31 @@ def edit(self, request, project_pk, pk):
...
+        try:
+            model = self.get_queryset().model
```
saanobhaai, line 274: Protect against model not having attr `protocol`?

src/api/resources/mixins.py
```diff
@@ -260,6 +264,31 @@ def edit(self, request, project_pk, pk):
...
+            with NamedTemporaryFile(delete=False) as f:
+                create_sample_unit_method_summary_report(
+                    project_pk,
+                    getattr(model, "protocol"),
+                    f.name,
+                    request=request
+                )
+                response = HttpResponse(content_type="application/vnd.openxmlformats-officedocument.spreadsheetml.sheet")
+                response["Content-Length"] = os.fstat(f.fileno()).st_size
+                response["Content-Disposition"] = f'attachment; filename="summary_report_{project_pk}.xlsx"'
```
saanobhaai, line 284: I think filename should be something more liike
`<sanitized_project_name>_<protocol>_<iso_date>.xlsx`
(I feel like we have a utility somewhere for producing  `sanitized_project_name`)

src/api/utils/reports.py
```diff
@@ -0,0 +1,39 @@
+from ..mocks import MockRequest
+from ..models import (
+    BENTHICLIT_PROTOCOL,
```
saanobhaai, line 3: I think we should add modifying this to the README documenting what needs to be done to add a new protocol

src/api/reports/summary_report.py
```diff
@@ -0,0 +1,240 @@
...
+    wb.remove(wb.active)
+    for sheet_name in sheet_names:
+        wb.create_sheet(title=sheet_name)
+
+    return wb
+
+
+def create_belt_fish_report(request, project_pk):
+    worksheet_names = [
+        "Belt Fish Obs",
```
saanobhaai, line 88: I think we should use full words everywhere, i.e. `observations`, `sample units`, and `sample events`

src/api/reports/summary_report.py
```diff
@@ -0,0 +1,240 @@
...
+
+    content = get_viewset_csv_content(BeltFishProjectMethodSEView, project_pk, request)
+    insert_range_and_resize(wb[worksheet_names[2]], "A1", content)
+
+    return wb
+
+
+def create_benthic_pit_report(request, project_pk):
+    worksheet_names = [
+        "Benthic PIT Obs",
```
saanobhaai, line 108: ... but we can leave `PIT` and `LIT` as acronyms...

src/api/reports/summary_report.py
```diff
@@ -0,0 +1,240 @@
...
+        BenthicLITProjectMethodSEView, project_pk, request
+    )
+    insert_range_and_resize(wb[worksheet_names[2]], "A1", content)
+
+    return wb
+
+
+def create_bleaching_qc_report(request, project_pk):
+    worksheet_names = [
+        "BQT Colonies Bleached Obs",
```
saanobhaai, line 160:
```
        "Colonies Bleached observations",
        "Benthic Percent observations",
        "Bleaching sample units",
        "Bleaching sample events",
```

saanobhaai (changes requested):
Verrrry nice...

src/api/resources/mixins.py
gridcell: avoiding circular reference.  Not ideal but I wanted to leverage the existing viewset CSV and since this is a mixin that gets added back to the view set it gets grumpy

src/api/reports/summary_report.py
gridcell: We're limited to ~30-32 chars (don't remember the exact number)

src/api/reports/summary_report.py
gridcell: same comment as above

PR #324, opened 2023-01-10
[M1771] catch invalid site location coordinates

saanobhaai:


saanobhaai:
[GDAL data type error](https://trello.com/c/XbvH2xIW/1771-gdal-data-type-error)

PR #325, opened 2023-01-10
[M1797] add missing PQT fields to parser

saanobhaai:


saanobhaai:
[PQT transect number bug](https://trello.com/c/q6NhcSr5/1797-pqt-transect-number-bug)

PR #330, opened 2023-01-20
Set up autoscaling for the API

aspcanada:


aspcanada:
[Implement auto scaling for Fargate API](https://trello.com/c/TNrbkw75/1811-implement-auto-scaling-for-fargate-api)

PR #331, opened 2023-01-20
only run one deploy action at a time

aspcanada:


aspcanada:
[see about making GH actions wait for existing deployments](https://trello.com/c/LQyXj5tB/1812-see-about-making-gh-actions-wait-for-existing-deployments)

PR #347, opened 2023-02-01
[M1822] handle None in expired_timestamp

saanobhaai:


saanobhaai:
[fishfamilies 500](https://trello.com/c/5zvgz5or/1822-fishfamilies-500)

PR #349, opened 2023-02-03
Queue worker construct

aspcanada:
I've refactored the api.py so that all of the "worker" resources are encapsulated into it's own construct. Please pay attention to the `cdk diff` in the comments after the diff runs.

### 🚨 PLEASE NOTE 🚨
This will tear down the existing SQS Queue so any message in the queue will be lost.

### DIFF
--- a/iac/stacks/constructs/worker.py
+++ b/iac/stacks/constructs/worker.py
@@ -6,16 +6,27 @@ from aws_cdk import (
     aws_cloudwatch_actions as cw_actions,
     aws_sns as sns,
     aws_sns_subscriptions as sns_subs,
+    aws_ecs_patterns as ecs_patterns,
+    aws_ecs as ecs,
+    aws_ec2 as ec2,
+    aws_applicationautoscaling as appscaling,
+    aws_s3 as s3,
 )
 
 from iac.settings.settings import ProjectSettings
 
-class MonitoredQueue(Construct):
+class QueueWorker(Construct):
     def __init__(
         self, 
         scope: Construct, 
         id: str, 
         config: ProjectSettings,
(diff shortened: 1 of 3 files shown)


iac/stacks/constructs/worker.py
```diff
@@ -0,0 +1,131 @@
...
+            evaluation_periods=1,
+            comparison_operator=cw.ComparisonOperator.GREATER_THAN_OR_EQUAL_TO_THRESHOLD,
+            treat_missing_data=cw.TreatMissingData.IGNORE,
+        )
+
+        # CloudWatch Alarm for DLQ
+        dlq_alarm = cw.Alarm(
+            self,
+            "DLQAlarm",
+            metric=queue.metric_approximate_number_of_messages_visible(
```
saanobhaai, line 78: Shouldn't this be `dead_letter_queue`?

iac/stacks/constructs/worker.py
```diff
@@ -0,0 +1,131 @@
...
+                appscaling.ScalingInterval(lower=1, change=+1),
+            ],
+        )
+
+        # allow worker access to public bucket
+        public_bucket.grant_read_write(fargate_service.task_definition.task_role)
+
+        # exports
+        self.queue = queue
+        # self.fargate_service = fargate_service
```
saanobhaai, line 131: delete comment?

iac/stacks/constructs/worker.py
aspcanada: yes, thank you!

iac/stacks/constructs/worker.py
aspcanada: done

PR #350, opened 2023-02-03
parameterize the queue name and changed the name itself

aspcanada:
fix cdk


aspcanada:
[DLQ](https://trello.com/c/FOrNb9Sm/1824-dlq)

PR #354, opened 2023-02-09
Cache summary data for observations, sample unit and sample events

gridcell:
Includes creating new set of models for the different sample units to store cached records from SQLModels for observation, sample unit and sample event summaries.


src/api/resources/sampleunitmethods/__init__.py
```diff
@@ -166,6 +166,6 @@ def csv(self, request, *args, **kwargs):
 
     def get_queryset(self):
         project_id = self.kwargs.get("project_pk")
-        return self.model.objects.all().sql_table(
+        return self.model.objects.filter(
```
saanobhaai, line 169: This is my favorite diff line -- what a lot lies behind that small change...

src/api/utils/summary_cache.py
```diff
@@ -0,0 +1,192 @@
...
+@timing
+def update_summary_cache(project_id, sample_unit=None, skip_test_project=True):
+    if (
+        skip_test_project is True
+        and Project.objects.filter(id=project_id, status=Project.TEST).exists()
+    ):
+        return
+
+    with transaction.atomic():
+        if sample_unit is None or sample_unit == FISHBELT_PROTOCOL:
```
saanobhaai, line 134: Why is `if sample_unit is None` repeated for every if statement? Do we want to run `_update_cache` for every protocol if `sample_unit` is not specified?

saanobhaai (commented):
This is a huge step forward...

src/api/utils/summary_cache.py
gridcell: if sample unit is None all the sample units need to run so I need to check it for each sample unit.  Unless I'm tired and missed something obvious (which could be the case)

PR #358, opened 2023-02-16
[M954] fix MR validations; django security patch bump

saanobhaai:


saanobhaai:
[MR rule validator triggering incorrectly](https://trello.com/c/sLSofShD/954-mr-rule-validator-triggering-incorrectly)

gridcell (approved):

PR #359, opened 2023-02-21
[M1766] make cloudtunnel for pg access

saanobhaai:


saanobhaai:
[pgadmin access](https://trello.com/c/O0na1gQt/1766-pgadmin-access)

Makefile
```diff
@@ -104,15 +104,31 @@ test:
...
-		--region us-east-1 \
+		--profile mermaid \
 		--cluster mermaid-api-infra-common-MermaidApiClusterB0854EC6-xitj9XbqTwap \
 		--task $(taskid) \
 		--container MermaidAPI \
 		--command "/bin/bash" \
 		--interactive
+
+cloudtunnel:
+	$(eval cluster=mermaid-api-infra-common-MermaidApiClusterB0854EC6-xitj9XbqTwap)
```
gridcell, line 124: Are we providing to much infrastructure details here?

cc: @aspcanada

saanobhaai: Yeah I don't like this either, but not sure what to do... Maybe move to another file, out of version control?

aspcanada: 1. I think the resource ID's should be in ENV VARs.
2. Perhaps some of the ports should be masked out too?

gridcell (approved):

PR #361, opened 2023-02-24
[M1762] exclude "Other" from SU/SE agg benthic views

saanobhaai:


saanobhaai:
[exclude observations of attributes with "Other" top-level category from benthic SU and SE aggregate endpoints](https://trello.com/c/23dihXJh/1762-exclude-observations-of-attributes-with-other-top-level-category-from-benthic-su-and-se-aggregate-endpoints)

src/api/models/sql_models/base.py
```diff
@@ -232,6 +232,9 @@ class BaseSQLModel(models.Model):
         string_agg(DISTINCT visibility_name, ', ' ORDER BY visibility_name) AS visibility_name
     """
 
+    # top-level benthic categories to exclude from SU- and SE-level summaries
```
gridcell, line 235: Do you expect that there will be more than one category to exclude?  If not, including `AND benthic_category != 'Other'` in the 3 protocols sql directly is clearer and you wouldn't have to introduce a base class property that isn't being used by the other protocols

gridcell (approved):

PR #362, opened 2023-03-02
[M1836] MR validation: handle id resulting in None from lookup

saanobhaai:


saanobhaai:
[validate 500](https://trello.com/c/BAew6DMX/1836-validate-500)

gridcell (approved):

PR #364, opened 2023-03-09
[M1842] return project info with push subnode 403s

saanobhaai:


saanobhaai:
[include project_id and project_name in 403 subnode responses](https://trello.com/c/gEOXZjhs/1842-include-projectid-and-projectname-in-403-subnode-responses)

src/api/resources/sync/views.py
```diff
@@ -112,6 +113,24 @@ def _get_source(source_type):
     return project_sources.get(source_type) or non_project_sources.get(source_type)
 
 
+def _get_project(data, source_type):
+    projid = None
+    projname = ""
```
gridcell, line 118: If this is getting display at some point, should there be a placeholder name, in the case the project doesn't exist?

saanobhaai: good point

PR #372, opened 2023-03-23
[M988] Add removes to revisions pull responses

gridcell:
Include "removes" in revision pull response.  Removes are used when a user is loses access to a record or related record.


gridcell:
[Number of submitted sample units in Project Card doesn't update after moving one to collecting.](https://trello.com/c/MbqBk6KZ/1015-number-of-submitted-sample-units-in-project-card-doesnt-update-after-moving-one-to-collecting)

src/api/resources/sync/pull.py
```diff
@@ -124,31 +156,40 @@ def get_records(viewset, revision_num=None, project=None, profile=None):
...
     :param viewset: ModelViewSet instance
     :type viewset: rest_framework.serializers.viewsets.ModelViewSet
-    :param revision_num: Revision number, defaults to None
-    :type revision_num: int, optional
-    :param project: Project id, defaults to None
-    :type project: UUID, optional
-    :param profile: Profile id, defaults to None
-    :type profile: UUID, optional
-    :return: Updates and deletes
+    :param profile_id: Profile id
```
saanobhaai, line 159: I think another comment or two for our future selves would help here -- I keep having to remind myself of the difference between the main `profile_id` arg and the `required_params` arg `profile`.

src/api/signals/revision.py
```diff
@@ -39,12 +39,14 @@ def _create_project_profile_revisions(query_kwargs):
 
 
 @receiver(pre_delete, sender=ProjectProfile)
-def update_delete_project_profile_revisions(sender, instance, *args, **kwargs):
+def delete_project_profile_revisions(sender, instance, *args, **kwargs):
```
saanobhaai, line 42: I think maybe it's worth a comment somewhere in this file explaining `related_to_profile_id` and noting that it is relevant to 1) deleting project profiles and 2) transferring CR ownership, for `pull/`s by the relevant user. And noting that 2) is handled in `replace.py` not by a signal.

Also: I'm sure it's here somewhere because it's previous functionality, but I'm not seeing how users *other than* `related_to_profile_id` know a project profile has been deleted?

src/api/tests/test_revision_pull.py
```diff
@@ -21,12 +21,17 @@ def test_get_records(db_setup, project1, project2, project3, profile1, project_p
...
+    updates, deletes, removes = get_records(
         collect_record_viewset,
-        None,
-        collect_record.project_id,
-        collect_record.profile_id,
+        profile1.pk,
+        required_params={
+            "revision_num": None,
+            "project_id": collect_record.project_id,
+            "profile_id": collect_record.profile_id,
```
saanobhaai, line 33: Shouldn't the `required_params` keys be `project` and `profile`? (Maybe naming should be made consistent?)

src/api/tests/test_revision_pull.py
saanobhaai, line 1: I think some comments in this file explaining what's being tested conceptually might help future selves

src/api/tests/test_revision_pull.py
```diff
@@ -146,9 +169,10 @@ def test_removed_from_project(db_setup, profile1, project_profile1):
 
     project_profile1.delete()
 
-    updates, deletes = get_records(
+    updates, deletes, removes = get_records(
```
saanobhaai, line 172: Shouldn't we also test for `len(removes)==1`?

saanobhaai (changes requested):

saanobhaai:
[Duplicate users shown after user added back to project](https://trello.com/c/INTOXDqI/988-duplicate-users-shown-after-user-added-back-to-project)

src/api/signals/revision.py
gridcell: 1. Added to note readme explaining "removes" rather than comments throughout the code

2. Other's receive a "delete" revision b/c the DB trigger creates one

PR #374, opened 2023-03-31
[M987] add CR transfer notis and emails

saanobhaai:


saanobhaai:
[CR transfer notifications](https://trello.com/c/FGJHi9Ki/987-cr-transfer-notifications)

src/templates/emails/crs_transferred_away.html
```diff
@@ -0,0 +1,10 @@
+{% extends "emails/base.html" %}
+{% block title %}Unsubmitted Sample Units transferred{% endblock %}
+{% block body %}
+<h2>Unsubmitted Sample Units transferred</h2>
```
gridcell, line 4: Should we be using "Collecting Sample Units transferred" or something equivalent?  Ties it back to the name on the website

saanobhaai: Let's leave it as "unsubmitted" -- we refer to the actual SU in that stage as such elsewhere as well.

PR #375, opened 2023-04-03
[M1039] add 'stale' validations status

saanobhaai:


saanobhaai:
[SU Validation: submit button is allowed after changing info and saving SU without the need to go through a validation](https://trello.com/c/9Pz8fbx2/1039-su-validation-submit-button-is-allowed-after-changing-info-and-saving-su-without-the-need-to-go-through-a-validation)

gridcell (approved):

PR #377, opened 2023-04-05
[M1046] Revision database trigger Update

gridcell:
`related_to_profile_id` does not need to be set via database trigger, it can only be set by explicitly creating a revision record.


gridcell:
[Error saving changes in the Project Info and Data Sharing Pages](https://trello.com/c/gzKk5M3A/1046-error-saving-changes-in-the-project-info-and-data-sharing-pages)

PR #378, opened 2023-04-13
Async metric logger

gridcell:


gridcell:
[endpoint logging](https://trello.com/c/WcqtWvtv/1765-endpoint-logging)

src/api/middleware.py
```diff
@@ -1,14 +1,17 @@
...
 
 from django.conf import settings
 from django.http import HttpResponse
 from django.utils.deprecation import MiddlewareMixin
 
+from tools.logger import DatabaseLogger
 from .utils.auth0utils import decode
 
 
+metrics_logger = DatabaseLogger(chunk_size=10)
```
saanobhaai, line 12: Might have asked this before -- I'd have expected 100 to be a better choice, but I don't know.

saanobhaai (commented):
Just a few questions

src/tools/models.py
```diff
@@ -0,0 +1,12 @@
+from django.db import models
+
+
+class LogEvent(models.Model):
+    timestamp = models.DateTimeField()
```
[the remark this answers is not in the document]
gridcell: The middleware uses UTC timestamp

src/api/middleware.py
```diff
@@ -63,8 +61,18 @@ def _obfuscate(self, value):
...
+        return any(
+            url_path.startswith(ignore_route)
+            for ignore_route in settings.METRICS_IGNORE_ROUTES
+        )
+
     def __call__(self, request):
         url_path = request.path
+
+        if settings.DISABLE_METRICS or self._ignore_url_path(url_path):
+            return self.get_response(request)
```
saanobhaai, line 74: A tiny point, but maybe the `response` definition currently on 83 should be moved up to 72, and then returned here?

saanobhaai (changes requested):
Just one trivial thing to check before merging

PR #380, opened 2023-04-15
[M1852] Excel reports recognize number values

gridcell:
Add data values were strings so when they are written to excel worksheet the format was being set as General.  This PR casts number strings to float or int.

### DIFF
--- a/src/api/utils/castutils.py
+++ b/src/api/utils/castutils.py
@@ -9,3 +9,20 @@ def to_decimal(val, max_digits, precision):
         ctx.prec = max_digits
         places = Decimal(10) ** (precision * -1)
         return Decimal(val).quantize(places, ROUND_HALF_DOWN)
+
+
+def to_number(string):
+    try:
+        num = float(string)
+        return int(num) if num.is_integer() else num
+    except ValueError:
+        return None
+
+
+def cast_str_value(string):
+    if not isinstance(string, str):
+        return string
+
+    val = to_number(string)
+    return val if val is not None else string
+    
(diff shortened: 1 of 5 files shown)


gridcell:
[format numeric columns in exported xlsx files](https://trello.com/c/J0dVKEfK/1852-format-numeric-columns-in-exported-xlsx-files)

src/api/utils/castutils.py
```diff
@@ -9,3 +9,20 @@ def to_decimal(val, max_digits, precision):
...
+def to_number(string):
+    try:
+        num = float(string)
+        return int(num) if num.is_integer() else num
+    except ValueError:
+        return None
+
+
+def cast_str_value(string):
+    if not isinstance(string, str):
```
saanobhaai, line 23: Thinking we should combine and make consistent this check and those of `is_decimal` above

src/api/utils/castutils.py
```diff
@@ -9,3 +9,20 @@ def to_decimal(val, max_digits, precision):
         ctx.prec = max_digits
         places = Decimal(10) ** (precision * -1)
         return Decimal(val).quantize(places, ROUND_HALF_DOWN)
+
+
+def to_number(string):
+    try:
```
saanobhaai, line 15: Also thinking we should protect against bad values here, in case it's ever used by itself. e.g. if `string` was `None` this would throw a `TypeError`, not a `ValueError`. Maybe all these functions should call a string-checking private function that does the same thing?

saanobhaai (changes requested):
Coupla things

PR #381, opened 2023-04-21
[M1711] Fix old broken project editing tests

gridcell:
Tests were breaking since we moved to using backend tasks for summary processing.  These changes fix those tests.


gridcell:
[create local queue and worker for pytest environment](https://trello.com/c/fFJUweAs/31-create-local-queue-and-worker-for-pytest-environment)

PR #385, opened 2023-04-25
[M168] upgrade boto3 and therefore botocore to remove deprecation message

saanobhaai:
ran tests and did a couple spot checks but otherwise not sure how to QA


saanobhaai:
[Update boto3](https://trello.com/c/OqlmRNkf/168-update-boto3)

PR #387, opened 2023-05-01
[M123] Use worker process for update_summaries command

gridcell:
This moves the process to a worker process so take the load of the web server


gridcell:
[update_summaries refactoring](https://trello.com/c/OYqRMhDf/123-updatesummaries-refactoring)

PR #388, opened 2023-05-02
separate task settings

aspcanada:


aspcanada:
[split sqs and webserver configs](https://trello.com/c/VEV3jq29/456-split-sqs-and-webserver-configs)

iac/settings/dev.py
```diff
@@ -7,9 +7,19 @@
     env_id=DEV_ENV_ID,
     database=DatabaseSettings(name=f"mermaid_{DEV_ENV_ID}", port="5432"),
     api=DjangoSettings(
+        # API
         container_cpu=512,
         container_memory=1024,
         container_count=1,
+
+        # SQS
+        sqs_cpu=512,
```
saanobhaai, line 16: I was thinking this would also be
```
        sqs_cpu=2048,
        sqs_memory=4096,
```
like `prod` so that we could test submitted jobs with the same resources, and it still wouldn't cost more because this wouldn't run very much on dev.

saanobhaai (changes requested):

PR #391, opened 2023-06-01
[M476] Strip out covariates from summary endpoints

gridcell:


gridcell:
[separate covars](https://trello.com/c/z7iWHJRx/476-separate-covars)

PR #393, opened 2023-06-08
[M493] match covars filter field for sites endpoint

saanobhaai:


saanobhaai:
[sites covars query param](https://trello.com/c/fJPZYLSO/493-sites-covars-query-param)

gridcell (approved):

PR #394, opened 2023-06-09
[M497] pop_cr_obsids script for v1 observations

saanobhaai:


saanobhaai:
[add observation ids to all CR observations saved with v1](https://trello.com/c/aL4etEv1/497-add-observation-ids-to-all-cr-observations-saved-with-v1)

PR #397, opened 2023-07-26
[M121] Improvements to numeric field-level validations

saanobhaai:


saanobhaai:
[Add depth/length/other numeric field validations to catch before dry submit](https://trello.com/c/tw7u6x6m/121-add-depth-length-other-numeric-field-validations-to-catch-before-dry-submit)

gridcell (approved):

PR #398, opened 2023-08-03
[M516] handle fish belt SUs with no observations in views

saanobhaai:


saanobhaai:
[fishbelt SUs with no observations should be output in SU/SE views](https://trello.com/c/7aZShQbN/516-fishbelt-sus-with-no-observations-should-be-output-in-su-se-views)

gridcell (approved):

PR #404, opened 2023-09-27
[M368] Create project revision when project related record is edited

gridcell:


gridcell:
[Sort project page by last updated](https://trello.com/c/tSImOpxM/368-sort-project-page-by-last-updated)

src/api/signals/revision.py
```diff
@@ -48,14 +47,8 @@ def update_profile_revisions(sender, instance, *args, **kwargs):
...
 
 
-@receiver(post_save, sender=ProjectProfile)
-def update_project_profile_revisions(sender, instance, *args, **kwargs):
-    Revision.create_from_instance(instance.project)
-
-
 @receiver(pre_delete, sender=ProjectProfile)
 def delete_project_profile_revisions(sender, instance, *args, **kwargs):
-    Revision.create_from_instance(instance.project)
```
saanobhaai, line 58: OK so we can drop all the project revision creation steps because of the db-level trigger, right? I think it's worth making a comment at the top of this file to remind our future selves about the triggers; I had to spend a few minutes remembering how we'd set this up.

src/api/models/mermaid.py
```diff
@@ -430,6 +430,10 @@ class Meta:
     def __str__(self):
         return '%s %s' % (self.site.__str__(), self.sample_date)
 
+    @property
```
saanobhaai, line 433: I think these project properties are readable and simple... Just wondering about `get_related_project` which we use in `TransectMethod` and other places where the instance is dynamic. My instinct is to consolidate into one definition of how to get a project from an instance, but I guess these are more readable?

saanobhaai (changes requested):
Looks great; just a couple of small comments.

PR #405, opened 2023-10-02
[M142] add standard deviations to agg views

saanobhaai:


saanobhaai:
[add standard deviations to agg views](https://trello.com/c/wSHbnUsa/142-add-standard-deviations-to-agg-views)

gridcell (approved):

PR #406, opened 2023-10-11
[M491] remove random id from SU agg views

saanobhaai:


saanobhaai:
[remove `id` from SU agg endpoints](https://trello.com/c/93GNybQt/491-remove-id-from-su-agg-endpoints)

gridcell (approved):

PR #409, opened 2023-10-13
[M574] catch no aca covariates when generating xlsx

saanobhaai:


saanobhaai:
[xlsx export 500](https://trello.com/c/8983Cldu/574-xlsx-export-500)

gridcell (approved):

PR #410, opened 2023-10-16
[M133] add assigned_transect_width_m to agg beltfish obs view

saanobhaai:


saanobhaai:
[utility for creating per-observation fish belt width numbers from "transect width" api field](https://trello.com/c/QYpggfgP/133-utility-for-creating-per-observation-fish-belt-width-numbers-from-transect-width-api-field)

saanobhaai:
@gridcell resolved

PR #413, opened 2023-10-20
[M119] Flatten Dictionary values in csv/excel reports

gridcell:


gridcell:
[plan for dealing with nested arbitrary keys / BI / further calcs in csv/xlsx routes](https://trello.com/c/gkFHSh7k/119-plan-for-dealing-with-nested-arbitrary-keys-bi-further-calcs-in-csv-xlsx-routes)

src/api/reports/csv_report.py
```diff
@@ -54,7 +84,7 @@ def get_csv_response(
     serialized_data = get_data(
         serializer_class, queryset, include_additional_fields=include_additional_fields, show_display_fields=show_display_fields
     )
-
+    serialized_data = list(serialized_data)
```
saanobhaai, line 87: Why not wrap the return value in `list()` within `get_data()`?

src/api/reports/rawcsv.py
```diff
@@ -30,6 +30,23 @@ def _flatten_record(self, record_dict):
     def _apply_formatters(self, flat_record):
         return self._flatten_record(flat_record)
 
+    def data(self, fields, data, *args, **kwargs):
```
saanobhaai, line 33: I may need to talk this one through... Wondering about 
- removing `self._apply_formatters()` from `stream`
- renaming `self.data()` to something like `self.formatted_data` (to separate out the formatting, allowing one in theory to be able to stream non-formatted data/headers)
- removing `self.stream_list` altogether, and adjusting `self.stream` header writing as necessary
but I'll bet there are reasons not to use this approach. Maybe it's a question of whether we need to preserve `self.stream` at all?

src/api/reports/csv_report.py
```diff
@@ -30,12 +30,42 @@ def get_data(
     )
     return serializer.get_serialized_data() or []
 
+ 
+def _flatten_column(column_name, column_records):
+    all_keys = {f"{column_name}_{key}": key for d in column_records for key in d.keys()}
+    return {new_key: [dic.get(key, None) for dic in column_records] for new_key, key in all_keys.items()}
```
saanobhaai, line 36: In theory the `None` arg isn't necessary

src/api/reports/summary_report.py
```diff
@@ -38,66 +41,121 @@
...
-        covar_lookup[site_id] = ["", ""]
         _update_covariate_lookup(covar_lookup, site_id, covariates)
 
     return covar_lookup
 
 
-def _append_aca_covariates(site_id_index, content, view_serializer):
-    """ Append Alan Coral Atlas covariates """
+def _get_site_aca_covariate_columns(site_ids: List[str]) -> Tuple[list, list]:
+    """Alan Coral Atlas covariates
```
saanobhaai, line 83: `Allen`

saanobhaai (changes requested):
Amazing, but you're right does add complexity!

requirements.txt
```diff
@@ -6,6 +6,7 @@ django-cors-headers==3.13.0
 django-cors-middleware==1.5.0
 django-debug-toolbar==3.6.0
 django-maintenance-mode==0.16.3
+django-stubs==4.2.5
```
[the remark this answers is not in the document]
gridcell: needed for python typing

PR #415, opened 2023-10-25
[M119] catch Nones in csv column flattening

saanobhaai:


saanobhaai:
[plan for dealing with nested arbitrary keys / BI / further calcs in csv/xlsx routes](https://trello.com/c/gkFHSh7k/119-plan-for-dealing-with-nested-arbitrary-keys-bi-further-calcs-in-csv-xlsx-routes)

PR #419, opened 2023-11-01
[M126] agg metrics output command

saanobhaai:


saanobhaai:
[user metrics](https://trello.com/c/ViQbZ5D5/126-user-metrics)

src/tools/management/commands/output_user_metrics.py
```diff
@@ -0,0 +1,19 @@
...
+        les = LogEvent.objects.all()
+        agg_les = agg_log_events(les)
+        fieldnames = agg_les[0].keys()
+        today = timezone.now().date().isoformat()
+        filename = f"user_metrics-{today}.csv"
+        with open(filename, "w") as csvfile:
+            writer = csv.DictWriter(csvfile, fieldnames=fieldnames)
+            writer.writeheader()
+            writer.writerows(agg_les)
+        # TODO: save to S3?
```
gridcell, line 19: Did you want to include uploading this to S3?

src/tools/metrics.py
```diff
@@ -0,0 +1,188 @@
...
+EVENT_TYPES_FILTER = (SUMMARY_EVENT_TYPE, PROJECT_EVENT_TYPE, SUBMIT_EVENT_TYPE)
+PROFILE_DEFAULT = {"first_name": "", "last_name": "", "email": ""}
+PROJECT_DEFAULT = {"project_name": "", "project_tags": "", "profiles": {}}
+
+
+def meta(method):
+    def _meta(*args, **kw):
+        args = list(args)
+        log_event = args[0]
+        if hasattr(log_event, "meta"):
```
gridcell, line 43: can't this if/else be replaced by...

`log_event.meta = log_event.meta or {}`

src/tools/metrics.py
```diff
@@ -0,0 +1,188 @@
...
+    "/bleachingqcs/sampleevents/",
+    "/habitatcomplexities/obshabitatcomplexities/",
+    "/habitatcomplexities/sampleunits/",
+    "/habitatcomplexities/sampleevents/",
+    "/benthicpqts/obstransectbenthicpqts/",
+    "/benthicpqts/sampleunits/",
+    "/benthicpqts/sampleevents/",
+    # "/summarysampleevents/",
+)
+SUMBIT_URL_SUFFIX = "/submit/"
```
gridcell, line 28: Spelling: `SUBMIT_URL_SUFFIX`

gridcell (changes requested):

src/tools/management/commands/output_user_metrics.py
saanobhaai: You're right, not hard to implement. I didn't see the point of convoluted env/cdk config when the name of the bucket is in the repo anyway, and the bucket doesn't need to be created or accessed programatically (just a human generating a report). For now.

src/tools/metrics.py
saanobhaai: As an obj I don't think that would quite do it but see what I did try to clean it up.

PR #420, opened 2023-11-03
[M149] Remove only successful jobs from queue

gridcell:


gridcell:
[Catch exceptions when processing jobs in worker](https://trello.com/c/kMHtCzQ7/149-catch-exceptions-when-processing-jobs-in-worker)

PR #427, opened 2023-11-29
[M610] include beltfish_su_family_zeroes in CTE

saanobhaai:


saanobhaai:
[Can't export data on Prod](https://trello.com/c/emPCh1nk/610-cant-export-data-on-prod)

PR #431, opened 2023-12-07
[M573] Remove reliance on Sample Unit and Sample Event ordering in tests

gridcell:


gridcell:
[ensure consistent order for SU agg views](https://trello.com/c/AERx81KU/573-ensure-consistent-order-for-su-agg-views)

PR #437, opened 2024-03-01
M682 Change ECS cluster to EC2 capacity

michaelconnor00:
Closes M682

This PR changes the ECS cluster to use EC2 instances. Currently it uses one `t3.medium`. In a future ticket, we will want to add an autoscale policy based on ECS managed metric. At this time we should decide if the instance should be smaller or larger. 

NOTE: Before this is merged to master, send out a notification that there will be some down time. I am confident that the new ECS containers will be created before the old ones are removed, but there is still some risk of downtime.


michaelconnor00:
Don't merge this yet. Looks like it is trying to delete the old fargate cluster, which will try to delete the prod one when it is merged to dev.

michaelconnor00:
Looks good now. The cluster added in commit `177e8a5` will need to be removed before merging into `master`.

gridcell (approved):

gridcell:
Refactor to change ECS to use EC2's from Fargate setup

PR #446, opened 2024-04-26
Fix FIFO queue naming

michaelconnor00:
Fifo suffix was harded coded to the queue name. This caused problems with changing to non-FIFO queues. See ticket: https://trello.com/c/I4tvwECc/756-summary-jobs-not-getting-added-to-queue?filter=*

Tests pass, see below.


michaelconnor00:
Tests pass
image

src/simpleq/queues.py
```diff
@@ -95,7 +96,10 @@ def queue(self):
         if self._queue:
             return self._queue
 
-        queue_name = f"{self.name}.fifo"
+        if self.USE_FIFO:
```
saanobhaai, line 99: I think this should be
```
        queue_name = self.name
        if self.USE_FIFO:
            queue_name = f"{self.name}.fifo"
```

saanobhaai (changes requested):
Reasonable -- but do we want to make this env-configurable? It seems to me whether we're doing FIFO or not is pretty bound up with the business logic. I guess it doesn't hurt, but I wouldn't want somebody (like me) to forget to set it and have things break. @gridcell what do you think -- should we just use `self.name` or is there value in making it configurable?

PR #458, opened 2024-06-12
fix updating restricted project summary SE records

saanobhaai:


src/api/utils/summary_cache.py
```diff
@@ -270,4 +276,5 @@ def update_summary_cache(project_id, sample_unit=None, skip_test_project=False):
         _update_project_summary_sample_event(project_id, skip_updates)
         timestamp = timezone.now()
         _update_unrestricted_project_summary_sample_events(project_id, timestamp, skip_updates)
+        print(f"skip_updates {skip_updates}")
```
gridcell, line 279: Do you want this print in here?

saanobhaai: Whoops nope

PR #467, opened 2024-07-10
M890

gridcell:


src/api/models/classification.py
```diff
@@ -0,0 +1,147 @@
...
+    labels = models.ManyToManyField(Label, related_name="classifiers")
+
+    class Meta:
+        db_table = "class_classifier"
+
+
+# TODO: Add validation to check that collect record is a photo quadrat
+class Image(BaseModel):
+    collect_record_id = models.UUIDField()
+    observation = models.ForeignKey(
```
saanobhaai, line 53: Shouldn't this be the other way around, i.e. the image belongs to an observation? >=1 observations per image.

src/api/resources/classification/image.py
```diff
@@ -0,0 +1,59 @@
...
+from rest_framework import status
+from rest_framework.response import Response
+
+from ...exceptions import check_uuid
+from ...models import BENTHICPQT_PROTOCOL, CollectRecord, Image, ObsBenthicPhotoQuadrat
+from ...permissions import CollectRecordOwner
+from ..base import BaseAPISerializer, BaseProjectApiViewSet
+
+
+class ImageSerializer(BaseAPISerializer):
```
saanobhaai, line 12: Not sure what the answer is but noting our discussion of what read-only nested info might want to be included here: classification status, list of points with annotations per point, etc.

src/api/utils/classification.py
```diff
@@ -0,0 +1,155 @@
...
+    for exif_key in exif_image.list_all():
+        if exif_key.startswith("gps_") or exif_key in (
+            "datetime_original",
+            "datetime_digitized",
+            "offset_time",
+        ):
+            continue
+        exif_image.delete(exif_key)
+
+    img_file = ContentFile(exif_image.get_file(), name=image_record.image.name)
```
saanobhaai, line 149: In progress I think but I think we should store `sample_event_id` somehow.

src/api/models/classification.py
```diff
@@ -0,0 +1,147 @@
...
+        return FileSystemStorage()
+    else:
+        return S3Storage(bucket_name=settings.IMAGE_PROCESSING_BUCKET)
+
+
+class Label(BaseModel):
+    benthic_attribute = models.ForeignKey(
+        BenthicAttribute, related_name="labels", on_delete=models.CASCADE
+    )
+    growth_form = models.ForeignKey(GrowthForm, related_name="labels", on_delete=models.CASCADE)
```
saanobhaai, line 23: Just remembered that `growth_form` needs to be nullable -- most Labels do not in fact have a growth form.

saanobhaai (changes requested):

PR #470, opened 2024-07-16
[M829] Allow filter projects fields in projects endpoint.

gridcell:


gridcell:
@coderabbitapi review

gridcell:
[add field query to projects/ endpoint](https://trello.com/c/s34PxecI/829-add-field-query-to-projects-endpoint)

PR #471, opened 2024-07-16
M890 Classification models and Image API Endpoint

gridcell:
## Summary by CodeRabbit

- **New Features**
  - Added new models and admin classes to support image classification and annotation.
  - Introduced new endpoints for handling image classification through the `ImageViewSet`.

- **Bug Fixes**
  - Updated permissions handling in `CollectRecordOwner` class.

- **Chores**
  - Updated Dockerfile environment variable declarations.
  - Added new dependencies for Django storages and image processing to `requirements.txt`.
  - Updated `.env.sample` and `.gitignore` for better configuration and exclusion practices.

- **Refactor**
  - Streamlined image processing bucket configuration and handling across various modules.


gridcell:
@coderabbitai review

gridcell:
@coderabbitai review

gridcell:
[Image upload and classification trigger](https://trello.com/c/vfUUTvK9/890-image-upload-and-classification-trigger)

src/app/settings.py
```diff
@@ -347,3 +355,20 @@ def show_toolbar(request):
 USE_FIFO = os.environ.get("USE_FIFO", "True")
 # Override default boto3 url for SQS
 ENDPOINT_URL = None if ENVIRONMENT in ("dev", "prod") else "http://sqs:9324"
+# AWS S3 bucket for public files
```
saanobhaai, line 358: I reorganized buckets into one section above in another branch so duplicates should be removed

src/api/models/classification.py
```diff
@@ -0,0 +1,155 @@
...
+        return FileSystemStorage()
+    else:
+        return S3Storage(bucket_name=settings.IMAGE_PROCESSING_BUCKET)
+
+
+class Label(BaseModel):
+    benthic_attribute = models.ForeignKey(
+        BenthicAttribute, related_name="labels", on_delete=models.CASCADE
+    )
+    growth_form = models.ForeignKey(GrowthForm, related_name="labels", on_delete=models.CASCADE)
```
saanobhaai, line 23: `growth_form` should be nullable

src/api/models/classification.py
```diff
@@ -0,0 +1,155 @@
...
+        return self.name
+
+    @property
+    def name(self):
+        return f"{self.benthic_attribute.name} {self.growth_form.name}"
+
+
+class Classifier(BaseModel):
+    name = models.CharField(max_length=50)
+    version = models.CharField(max_length=11)
```
saanobhaai, line 40: Why `max_length=11`? Seems potentially short.

src/api/models/classification.py
```diff
@@ -0,0 +1,155 @@
...
+    patch_size = models.IntegerField(help_text="Number of pixels")
+    num_points = models.IntegerField(default=25)
```
saanobhaai, line 42: Is `num_points` stored with the classifier in CoralNet? I think of this as a setting for use in applying a classifier to an image, but not a property of the classifier itself.

src/api/models/classification.py
```diff
@@ -0,0 +1,155 @@
...
+    description = models.TextField(max_length=1000, blank=True)
+    labels = models.ManyToManyField(Label, related_name="classifiers")
+
+    class Meta:
+        db_table = "class_classifier"
+
+
+class Image(BaseModel):
+    collect_record_id = models.UUIDField()
```
saanobhaai, line 51: coderabbit suggested indexing this and that seems desirable

saanobhaai, line 51: Should we add a `project_lookup` property? 
Also we talked originally about storing image files (and associated resources) with an obfuscated `project_id` and (obfuscated?) image id in the path. We then were thinking about flattening that structure, but I think we have to group them by project for the sake of the ML folks. For discussion.

src/api/models/classification.py
```diff
@@ -0,0 +1,155 @@
...
+    observation = models.ForeignKey(
+        ObsBenthicPhotoQuadrat, null=True, blank=True, on_delete=models.SET_NULL
+    )
+
+    image = models.ImageField(upload_to="", storage=select_image_storage)
```
saanobhaai, line 56: Let's chat about `upload_to` here and in thumbnail. Also, thinking ahead: the classification task will need to store feature vectors and hopefully an annotations csv with the image, so we'll need to ensure we can get the right path to the task so it stores in the right place.
image

src/api/models/classification.py
```diff
@@ -0,0 +1,155 @@
...
+                condition=models.Q(is_confirmed=True),
+            )
+        ]
+
+    def __str__(self):
+        return f"{self.label} - {self.score} - {self.is_confirmed}"
+
+    @property
+    def is_human_created(self):
+        return self.classifier is None and self.is_confirmed
```
saanobhaai, line 120: Now I'm wondering about a possible future edge case in which we would want to have an annotation that is human-created (`self.classifier is None`) but not confirmed (perhaps because they're not sure).

src/api/utils/classification.py
```diff
@@ -0,0 +1,181 @@
...
+from PIL.ExifTags import TAGS
+
+from ..models import Image
+
+
+def create_unique_image_name(image: ImageFieldFile) -> str:
+    image_name = image.name
+    image_ext = os.path.splitext(image_name)[1]
+    current_utc_time = datetime.datetime.now(datetime.timezone.utc)
+    uid = str(uuid.uuid4()).replace("-", "")
```
saanobhaai, line 24: Why not tie this uuid to that of the `Image` primary key? And do we need to add timestamp?

src/api/utils/classification.py
```diff
@@ -0,0 +1,181 @@
...
+
+    return file_hash.hexdigest()
+
+
+def create_thumbnail(image: ImageFieldFile) -> ContentFile:
+    img = PILImage.open(image)
+    size = (100, 100)
+    img.thumbnail(size, PILImage.LANCZOS)
+
+    base, ext = os.path.splitext(image.name)
```
saanobhaai, line 46: This is the name of the image after doing `create_unique_image_name` right?

src/api/utils/classification.py
```diff
@@ -0,0 +1,181 @@
...
+
+    for chunk in image.chunks():
+        file_hash.update(chunk)
+
+    return file_hash.hexdigest()
+
+
+def create_thumbnail(image: ImageFieldFile) -> ContentFile:
+    img = PILImage.open(image)
+    size = (100, 100)
```
saanobhaai, line 43: Just checking that these settings (with respect to aspect ratio, resolution, etc.) are more or less equivalent to CoralNet's?

src/api/signals/classification.py
```diff
@@ -0,0 +1,48 @@
...
+        instance.data = instance.data or {}
+        if "original_image_name" not in instance.data:
+            instance.data["original_image_name"] = original_image_name
+
+        image_name = cls_utils.create_unique_image_name(instance.image)
+        instance.name = image_name
+        instance.image.name = image_name
+
+        cls_utils.correct_image_orientation(instance)
+        cls_utils.update_exif(instance)
```
saanobhaai, line 22: For discussion; latest thinking is we strip location and timestamp.

src/api/resources/classification/image.py
```diff
@@ -0,0 +1,69 @@
...
+
+from ...exceptions import check_uuid
+from ...models import BENTHICPQT_PROTOCOL, CollectRecord, Image, ObsBenthicPhotoQuadrat
+from ...permissions import CollectRecordOwner
+from ...utils import truthy
+from ...utils.classification import classify_image
+from ..base import BaseAPISerializer, BaseProjectApiViewSet
+
+
+class ImageSerializer(BaseAPISerializer):
```
saanobhaai, line 14: So we'll handle potential nested serializers needed by frontend in a future ticket? We should reconvene with Parm and Julia; seems to me the CR should store an object that would make use of a list of annotations for each point in a list of points per image.

saanobhaai (changes requested):
Looking good; some items for discussion mostly.

src/api/models/classification.py
```diff
@@ -0,0 +1,155 @@
...
+        db_table = "class_point"
+
+
+class Annotation(BaseModel):
+    point = models.ForeignKey(Point, on_delete=models.CASCADE, editable=False)
+    label = models.ForeignKey(Label, related_name="annotations", on_delete=models.CASCADE)
+    classifier = models.ForeignKey(
+        Classifier, null=True, on_delete=models.CASCADE, related_name="annotations"
+    )
+    score = models.PositiveSmallIntegerField(default=0)
```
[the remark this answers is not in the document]
gridcell: We can set this as nullable but ...

```
Like a [PositiveIntegerField](https://docs.djangoproject.com/en/5.0/ref/models/fields/#django.db.models.PositiveIntegerField), but only allows values under a certain (database-dependent) point. Values from 0 to 32767 are safe in all databases supported by Django.
```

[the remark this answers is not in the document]
saanobhaai: You mean the min/max constraint? We just need to handle 0-100. Or does this means we'd have to use a different field type because `PositiveSmallIntegerField` doesn't permit NULLs? If so maybe it's not worth changing -- we just need to know that a null `classifier` is what's important, and use that to distinguish; a 0 `score` could in theory be human, or not.

requirements.txt
```diff
@@ -29,4 +29,7 @@ requests==2.28.1
 rest_condition==1.0.3
 simpleflake==0.1.5
 timezonefinder==6.1.1
-whitenoise==6.2.0
\ No newline at end of file
+whitenoise==6.2.0
+django-storages[s3]==1.14.3
```
saanobhaai, line 33: Why not put in alpha order to match everything else?

iac/stacks/api.py
```diff
@@ -298,6 +301,17 @@ def __init__(
...
-        image_processing_bucket.grant_write(image_worker.task_definition.task_role)
-        image_processing_bucket.grant_read_write(service.task_definition.task_role)
+        if config.env_id == "dev":
+            # Create classification image bucket
+            image_processing_bucket = s3.Bucket(
+                self,
+                id="MermaidImageProcessingBackupBucket",
+                bucket_name=image_bucket_name,
+                removal_policy=RemovalPolicy.RETAIN,
+                public_read_access=False,
```
saanobhaai, line 311: I was thinking this would need to be public so frontend can hit image urls directly; are you thinking differently?

src/api/models/classification.py
```diff
@@ -0,0 +1,172 @@
...
+    def collect_record(self):
+        return CollectRecord.objects.get_or_none(id=self.collect_record_id)
+
+    @property
+    def project(self):
+        cr = self.collect_record
+        if cr:
+            return cr.project
+
+        obs = self.observation
```
saanobhaai, line 88: This could be `None` so should we protect against that?

src/api/models/classification.py
saanobhaai: Don't we still need to set this?

src/api/signals/classification.py
```diff
@@ -0,0 +1,47 @@
...
+        original_image_name = instance.image.name
+        instance.data = instance.data or {}
+        if "original_image_name" not in instance.data:
+            instance.data["original_image_name"] = original_image_name
+
+        image_name = cls_utils.create_unique_image_name(instance)
+        instance.name = image_name
+        instance.image.name = image_name
+
+        cls_utils.correct_image_orientation(instance)
```
saanobhaai, line 21: If any of this fails for any reason, should we prevent the save so that we don't get an image with exif data?

src/app/settings.py
```diff
@@ -347,3 +349,15 @@ def show_toolbar(request):
...
 ENDPOINT_URL = None if ENVIRONMENT in ("dev", "prod") else "http://sqs:9324"
+
+STORAGES = {
+    "images": {
+        "BACKEND": "storages.backends.s3.S3Storage",
+        "OPTIONS": {
+            "bucket_name": IMAGE_PROCESSING_BUCKET,
+            "object_parameters": {},
+            "file_overwrite": False,
+            "location": "",  # A path prefix that will be prepended to all uploads.
```
saanobhaai, line 360: shouldn't this be set to `mermaid/`?

src/api/resources/classification/image.py
```diff
@@ -0,0 +1,68 @@
...
+        extra_actions = super().get_extra_actions()
+        filtered_actions = [
+            action for action in extra_actions if action.url_name not in ["missing", "updates"]
+        ]
+        return filtered_actions
+
+    def create(self, request, *args, **kwargs):
+        image = request.data.get("image")
+        collect_record_id = request.data.get("collect_record_id")
+        trigger_classification = truthy(request.data.get("classify", False))
```
saanobhaai, line 45: Just wondering about this... As things stand this means the possibility of an image without generated annotations, and no way of adding them other than by deleting and creating. Any reason not to simply always trigger classification, as simpler with an easy path for adding later?

saanobhaai (changes requested):

iac/stacks/api.py
gridcell: They should be provided a signed url by django storages

src/api/resources/classification/image.py
gridcell: How about defaulting it to classify with the option not to?

saanobhaai: sure that works

iac/stacks/api.py
saanobhaai: Hmm but if it's a Public Data bucket that any ML researcher can use without association with MERMAID - ?

saanobhaai (approved):

PR #476, opened 2024-08-02
[M950] Classification Status Endpoint

gridcell:


gridcell:
@coderabbitai review

gridcell:
[Image classification status endpoint](https://trello.com/c/6QYKvMiR/950-image-classification-status-endpoint)

src/api/resources/classification/classification_status.py
```diff
@@ -0,0 +1,63 @@
...
+        show_all = "showall" in self.request.query_params
+        qs = ClassificationStatus.objects.all()
+
+        if show_all is True:
+            return qs
+
+        latest_status_subquery = (
+            ClassificationStatus.objects.filter(image=OuterRef("image"))
+            .order_by("-created_on")
+            .values("id")[:1]
```
saanobhaai, line 47: Nitpick, but to me `[:1]` is less readable than `[0]`. Is that in order to avoid a `KeyError`, if the query returned nothing? But wouldn't we get one anyway with the brackets, and shouldn't we protect against that?

src/api/resources/classification/classification_status.py
```diff
@@ -0,0 +1,63 @@
...
+class ClassificationStatusSerializer(BaseAPISerializer):
+    class Meta:
+        model = ClassificationStatus
+        fields = ["id", "image", "status", "message", "data"]
+
+
+class ClassificationStatusFilterSet(BaseAPIFilterSet):
+    collect_record = filters.BooleanFilter(field_name="collect_record_id", method="filter_covars")
+
+    def filter_covars(self, queryset, name, value):
```
saanobhaai, line 19: The name `filter_covars` is confusing to me. Isn't it something more like `filter_crs`?

saanobhaai (changes requested):

src/api/resources/classification/classification_status.py
gridcell: With the `[:1]`, you get an empty list back if there's no match.

saanobhaai: OK yeah I'd thought that we'd need an empty queryset, not an empty list; but it looks like from the docs it is an empty queryset, so all good.

saanobhaai (approved):

PR #484, opened 2024-08-26
M984 Add more SQS Permission

michaelconnor00:
Add more permissions so Workers can delete and send messages. See the ticket for reference.


michaelconnor00:
@saanobhaai Do you want to review this or just merge?

saanobhaai (approved):

PR #487, opened 2024-09-05
[M992] Check if BPQT collect record is flagged for classification

gridcell:


gridcell:
[Restricting collect record's that can use image classification](https://trello.com/c/d4Zp8vNP/992-restricting-collect-records-that-can-use-image-classification)

PR #492, opened 2024-09-19
[M1044] add observation count fields to habitat complexity/SSE views

saanobhaai:


saanobhaai:
[add hab comp obs count fields to SU/SE views](https://trello.com/c/HANyIl5H/1044-add-hab-comp-obs-count-fields-to-su-se-views)

gridcell (approved):

PR #499, opened 2024-10-09
[M1081] do not report other BA-based LH proportions when there is a GF match

saanobhaai:


saanobhaai:
[fix PIT obs LH proportion calculations](https://trello.com/c/xXWcRHN6/1081-fix-pit-obs-lh-proportion-calculations)

gridcell (approved):

PR #500, opened 2024-10-10
[M1033] enable GFCR read-only non-proj access; prevent premature proj 403

saanobhaai:


saanobhaai:
[GFCR permissions](https://trello.com/c/be8LslDI/1033-gfcr-permissions)

gridcell (approved):

PR #509, opened 2024-10-28
[M1123]  Fix trying to access annotations before image is processed

gridcell:


gridcell:
['NoneType' object has no attribute 'patch_size'](https://trello.com/c/IvQ4u010/1123-nonetype-object-has-no-attribute-patchsize)

src/api/submission/validations2/validators/classification_benthic_photo_quadrat.py
```diff
@@ -51,31 +54,40 @@ def __call__(self, *args, **kwargs):
...
         annos = Annotation.objects.select_related("point", "point__image").filter(
             is_confirmed=True, point__image_id=image_id
         )
 
-        classifier = annos[0].classifier
+        classifier = Classifier.objects.get_or_none(id=collect_record.get("classifier_id"))
+        if not classifier:
+            classifier = Classifier.latest()
+            if not classifier:
+                return OK, None, [context]
```
saanobhaai, line 72: Why is `context` a list? That looks inconsistent with what's getting returned elsewhere.

src/api/utils/classification.py
```diff
@@ -153,6 +153,8 @@ def correct_image_orientation(image_record: Image):
     except (AttributeError, KeyError, IndexError) as _:
         pass
 
+    # Saving the orientated image back to the image record
```
saanobhaai, line 156: How about renaming `correct_image_orientation` to something like `save_normalized_imagefile` and then we add any additional file editing steps here? Right now it's stripping EXIF as well as correcting image orientation, and that's a major step; there could be others.

saanobhaai (changes requested):

saanobhaai (approved):

PR #511, opened 2024-10-29
[1127] hotfix: ensure image_classification is False not None

saanobhaai:


saanobhaai:
[Can't submit PQT SU after all warnings and errors addressed](https://trello.com/c/A601WXYq/1127-cant-submit-pqt-su-after-all-warnings-and-errors-addressed)

gridcell (approved):

PR #515, opened 2024-11-08
[M1142] Allow CR owner to delete; simplify CR perms

saanobhaai:


saanobhaai:
[CR 403 for Collector](https://trello.com/c/sK98iytC/1142-cr-403-for-collector)

PR #518, opened 2024-11-12
[M1118] change GFCR annual_revenue and verbose names

saanobhaai:


saanobhaai:
[GFCR form changes](https://trello.com/c/tyflN2US/1118-gfcr-form-changes)

PR #521, opened 2024-11-17
[M1134] catch excessive depth precision

saanobhaai:


saanobhaai:
[capture depth decimal places in validator](https://trello.com/c/Rv0g8DCw/1134-capture-depth-decimal-places-in-validator)

PR #527, opened 2024-11-28
[M1133] User metrics

gridcell:
## Summary by CodeRabbit

- **New Features**
	- Introduced a new Django admin interface for managing user metrics.
	- Added functionality for executing daily tasks, including database backups and metrics updates.
	- Created a new model to capture user-related metrics, enhancing data tracking capabilities.

- **Bug Fixes**
	- Improved handling of log events and aggregation of metrics, including counting image uploads.

- **Documentation**
	- Updated help texts for new management commands to clarify their purposes.


gridcell:
@coderabbitai review

gridcell:
[make user_metrics an updated table rather than a csv export](https://trello.com/c/6zPmCyqI/1133-make-usermetrics-an-updated-table-rather-than-a-csv-export)

src/tools/management/commands/update_metrics.py
```diff
@@ -0,0 +1,49 @@
+from datetime import datetime
```
saanobhaai, line 1: We should deal with output_user_metrics.py and the bucket it writes to. Should we delete both? Or consolidate with this command as an option?

src/tools/management/commands/update_metrics.py
```diff
@@ -0,0 +1,49 @@
...
+
```
saanobhaai, line 2: Doing just one day at a time makes sense, but we still need the table to include the aggregations of all log events from the beginning (the way output_user_metrics does). Could be a one-time data migration?

saanobhaai (changes requested):

src/tools/management/commands/update_metrics.py
gridcell: Removed command

src/tools/management/commands/update_metrics.py
gridcell: Manually run on server with a defined date range.

saanobhaai (approved):

PR #537, opened 2025-01-06
[M1216] Update Docker Compose

tienhoah:


Makefile
```diff
@@ -18,13 +18,14 @@
 
 
 API_SERVICE="api_service"
+DOCKER_COMPOSE := $(shell command -v docker-compose >/dev/null 2>&1 && echo docker-compose || echo docker compose)
```
gridcell, line 21: I like this solution

gridcell (approved):

PR #552, opened 2025-01-28
M1252

gridcell:


iac/stacks/constructs/worker.py
```diff
@@ -53,9 +53,9 @@ def __init__(
             # SQS queue's ApproximateNumberOfMessagesVisible metric
             scaling_steps=[
                 # when <=10 messages, scale down
-                appscaling.ScalingInterval(upper=10, change=-1),
+                appscaling.ScalingInterval(upper=100, change=-1),
```
saanobhaai, line 56: Just checking this is desirable for all workers?

src/api/signals/summaries.py
```diff
@@ -21,14 +19,10 @@ def update_summaries_on_delete_transect_method(sender, instance, *args, **kwargs
...
-        True,
-        update_summary_cache,
-        project_id=project.pk,
-        sample_unit=instance.protocol,
-        timestamp=timezone.now(),
-    )
+    try:
+        add_project_to_queue(project.pk)
+    except Exception as e:
+        print(f"Failed to queue summary update for project {project.pk}: {e}")
```
saanobhaai, line 25: Hmmm -- wanting something other than a `print()` when things go wrong. Use logging framework to email me in prod, or output to Cloudwatch?

src/api/utils/summary_cache.py
```diff
@@ -282,6 +284,29 @@ def _update_unrestricted_project_summary_sample_events(
     )
 
 
+def add_project_to_queue(project_id, skip_test_project=False):
+    try:
+        if isinstance(project_id, uuid.UUID) is False:
```
saanobhaai, line 289: use `check_uuid()`?

src/api/utils/summary_cache.py
```diff
@@ -282,6 +284,29 @@ def _update_unrestricted_project_summary_sample_events(
...
+            uuid.UUID(project_id)
+    except (ValueError, TypeError, AttributeError):
+        raise ValueError(f"Invalid project_id: {project_id}")
```
saanobhaai, line 292: So here again, this would generate an exception I'd get emailed about?

src/api/utils/summary_cache.py
```diff
@@ -282,6 +284,29 @@ def _update_unrestricted_project_summary_sample_events(
...
+
+    with connection.cursor() as cursor:
+        if skip_test_project and Project.objects.filter(status=Project.TEST).exists():
```
saanobhaai, line 295: Doesn't the filter need to include the project_id?

src/api/utils/summary_cache.py
```diff
@@ -292,6 +317,7 @@ def update_summary_cache(project_id, sample_unit=None, skip_test_project=False,
         skip_updates = True
 
     try:
+        sample_unit = sample_unit or None
```
saanobhaai, line 320: If the default is `None`, why is this line necessary?

src/tools/management/commands/process_summaries.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def acquire_lock(self):
+        with connection.cursor() as cursor:
+            cursor.execute("SELECT pg_try_advisory_lock(%s);", [self.LOCK_ID])
+            locked = cursor.fetchone()[0]
+        return locked
+
+    def release_lock(self):
+        with connection.cursor() as cursor:
+            cursor.execute("SELECT pg_advisory_unlock(%s);", [self.LOCK_ID])
```
saanobhaai, line 24: Should this return something, in case something needs to know whether it was successful?

src/tools/management/commands/process_summaries.py
```diff
@@ -0,0 +1,58 @@
...
+from api.utils.summary_cache import update_summary_cache
+
+
+class Command(BaseCommand):
+    LOCK_ID = 8080
+    WAIT_SECONDS = 5
+
+    def acquire_lock(self):
+        with connection.cursor() as cursor:
+            cursor.execute("SELECT pg_try_advisory_lock(%s);", [self.LOCK_ID])
```
saanobhaai, line 18: What happens if a lock can't be acquired for some reason? If a lock already exists, I guess this returns `None`; but what if an exception is raised, or some other result happens?

src/tools/management/commands/process_summaries.py
```diff
@@ -0,0 +1,58 @@
...
+            return True
+        except Exception as e:
+            print(f"Error update_summary_cache[Project id: {task.project_id}]: {e}")
+            task.processing = False
+            task.attempts += 1
+            task.save()
+            return False
+
+    def process_tasks(self, tasks):
+        with ThreadPoolExecutor(max_workers=min(os.cpu_count() - 1, 3)) as executor:
```
saanobhaai, line 41: I assume you've thoroughly thought through what `max_workers` should be

src/api/utils/summary_cache.py
```diff
@@ -282,6 +284,29 @@ def _update_unrestricted_project_summary_sample_events(
...
+        VALUES (%s, false, 0, now())
+        ON CONFLICT (project_id, processing)
+        DO NOTHING;
+        """
+
+        cursor.execute(sql, [project_id])
+
+
 @timing
 def update_summary_cache(project_id, sample_unit=None, skip_test_project=False, timestamp=None):
```
saanobhaai, line 311: Do we still need `timestamp` functionality?

src/tools/management/commands/process_summaries.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def release_lock(self):
+        with connection.cursor() as cursor:
+            cursor.execute("SELECT pg_advisory_unlock(%s);", [self.LOCK_ID])
+
+    def _process_tasks(self, task):
+        try:
+            task.processing = True
+            task.save()
+            update_summary_cache(task.project_id)
```
saanobhaai, line 30: Do we need to pass through the test project arg?

saanobhaai (changes requested):

iac/stacks/constructs/worker.py
gridcell: That was one of the changes I made with Mike, to reduce when it would need to autoscale since we don't have the longer running tasks anymore with update_summary.

src/api/signals/summaries.py
gridcell: This would be the equivalent of submit_job not working (currently a Cloudwatch log).  I can change this to notify if you want.

src/api/utils/summary_cache.py
gridcell: I was on the fence if I should use this.  I always associate it with returning a 400 error response but I agree it's likely worth using check_uuid

src/api/utils/summary_cache.py
gridcell: Will change this was a change to support empty string sample_unit to get around the postgres version not support null's as expected with upserts.

src/tools/management/commands/process_summaries.py
gridcell: pg_try_advisory_lock returns back true or false

I've wrapped both acquire and release in try/except to log if there's some DB exception

src/tools/management/commands/process_summaries.py
gridcell: no because that's handled when the task gets added to the table

saanobhaai (approved):

PR #554, opened 2025-01-31
[M1253] Using sql for update summary cache

gridcell:


src/api/utils/summary_cache.py
```diff
@@ -64,6 +64,134 @@
...
+        sql.append(_delete(SummarySampleEventModel, project_id))
+
+    sse_sql = _convert_to_sql(
+        SummarySampleEventSQLModel.objects.all().sql_table(project_id=project_id)
+    )
+    sql.append(_insert(SummarySampleEventModel, sse_sql, suggested_citation))
+
+    with connection.cursor() as cur:
+        cur.execute("".join(sql))
+
```
saanobhaai, line 190: A few extra lines here

src/api/utils/summary_cache.py
```diff
@@ -170,31 +298,13 @@ def _update_bleaching_qc_summary(
     _update_records(bleaching_se, BleachingQCSEModel, created_on, suggested_citation, skip_updates)
 
 
-def _update_project_summary_sample_event(project_id, skip_test_project=True):
```
saanobhaai, line 173: Seems like we should be consistent. This function gets deleted (and the sql equiv added), whereas in other cases the corresponding non-sql function isn't deleted. Should be one or the other.

saanobhaai (changes requested):

PR #573, opened 2025-04-10
[M1215] separate db envs and prepare for pg 17

saanobhaai:
## Summary by CodeRabbit

- **Chores**
  - Upgraded the database service to use a more current PostGIS image, ensuring access to updated features and fixes.
  - Introduced separate development and production database instances with distinct PostgreSQL versions for improved environment management.
  - Corrected the assignment of the AWS secret access key to ensure proper configuration in production environments.


saanobhaai:
[upgrade postgres](https://trello.com/c/CWSTaoYI/1215-upgrade-postgres)

iac/stacks/common.py
```diff
@@ -21,6 +21,36 @@
 
 
 class CommonStack(Stack):
+    def _database(self, id: str, version: rds.PostgresEngineVersion) -> rds.DatabaseInstance:
```
michaelconnor00, line 24: I don't see us adding more than 2 databases, so no need for a function. To keep it DRY, you can put common attributes in a `dict` and deconstruct it (`**common_attrs`) in the `rds.DatabaseInstance` constructor call.

iac/stacks/common.py
```diff
@@ -62,33 +92,12 @@ def __init__(
...
-                ec2.InstanceClass.BURSTABLE3,
-                ec2.InstanceSize.SMALL,
-            ),
-            vpc_subnets=ec2.SubnetSelection(subnet_type=ec2.SubnetType.PRIVATE_ISOLATED),
-            backup_retention=Duration.days(7),
-            deletion_protection=True,
-            removal_policy=RemovalPolicy.SNAPSHOT,
-            credentials=rds.Credentials.from_secret(database_credentials_secret),
+        self.prod_database = self._database(
+            id="ProdDatabase", version=rds.PostgresEngineVersion.VER_13_7
```
michaelconnor00, line 100: With CDK, when you change the name of the `id`, it will common delete the resource. If you look at the diff that ran in the GHA, there is a line that says `[-] AWS::RDS::DBInstance PostgresRdsV2 PostgresRdsV2B4B63A33 destroy`, which means it will delete the production database. If you keep the `id=PostgresRdsV2`, and don't change anything else, it won't get deleted.

michaelconnor00 (changes requested):
Looking good. Just one critical comment about changing the resource `id`.

michaelconnor00:
One more thing, the current RDS instance is a `db.t4g.medium`. I presume that the production env is driving that instance size. I noticed your changes have both databases as SMALL. As part of this PR, I would leave prod unchanged, so leave it as medium. If you want to reduce the size of the DB, then let's do that after this change.

michaelconnor00:
Once deployed, I am presuming you will cleanup the old dev database manually.

iac/stacks/common.py
saanobhaai: Started down this road but the part of the function that generates the secret has to be repeated, and that's more than half the function; thinking it's not worth a refactor?

michaelconnor00: That is fine.

PR #574, opened 2025-04-16
[M1393] Endpoint for combined-export zip

gridcell:
## Summary by CodeRabbit

- **New Features**
  - Summary reports are now grouped and generated by project data policy level, with each group included in a single downloadable zip archive.

- **Improvements**
  - Report emails now provide a single zip file containing all relevant reports, simplifying access to multiple policy-level reports.
  - File deletion and zipping utilities now support handling multiple files at once.
  - Email report filenames no longer include data policy labels.
  - Report downloads now serve the Excel file directly instead of a zip archive, improving access and cleanup.
  - Temporary report files are named with ISO date prefixes for clearer identification.

- **Bug Fixes**
  - Improved file handling to ensure successful deletion and zipping of multiple files.
  - Enhanced cleanup logic to prevent errors when temporary files or paths are unset.

- **Tests**
  - Removed outdated tests related to previous report policy-level logic.


src/api/reports/gfcr.py
```diff
@@ -236,18 +236,21 @@ def create_report(project_ids, request=None, send_email=None):
...
         except Exception:
             logger.exception("Error saving workbook")
             return None
 
-        if send_email:
-            email_report(request.user.profile.email, output_path, "GFCR")
-            delete_file(output_path)
-        else:
-            return output_path
+    if send_email:
```
saanobhaai, line 249: Was the loss of indent here deliberate? Needs `output_path` no?

saanobhaai (changes requested):
Looks good, with one indent comment, but - 
I may be missing something but it seems like it's the same endpoint, no? The hope was to allow users to either do what they're doing (getting a single-xlsx zip) or request an all-three zip.

saanobhaai (approved):

PR #587, opened 2025-05-13
Onboarding - Setup Note Updates

hurrendor:


src/api/utils/__init__.py
hurrendor, line 1: I added these updates from the `master` branch after starting, they probably need to wait to be merged until after the next release because of the random updates they're pulling along with

PR #591, opened 2025-05-22
[M1506] fix for projects/ extent 500

saanobhaai:


saanobhaai:
[projects/ 500 with country or name filters](https://trello.com/c/L16obdiL/1506-projects-500-with-country-or-name-filters)

gridcell (approved):

PR #596, opened 2025-06-02
IAM User creds to generate presigned urls for reports 

ms280690:
WIP, create report se user and add secrets to api, update env vars and presign url function is application code

## Summary by CodeRabbit

- **New Features**
  - Added support for dedicated S3 credentials for report handling, with new environment variables for access and secret keys.
  - Credentials are securely generated and stored for improved security and access management.
  - Report email functionality now explicitly uses these credentials when generating S3 presigned URLs.

- **Chores**
  - Updated sample environment configuration to include new S3 credential variables.


iac/stacks/common.py
```diff
@@ -313,6 +313,48 @@ def __init__(
                 open=True,
             )
 
+        self.report_s3_user = iam.User(
```
saanobhaai, line 316: This looks fine but does creating the IAM user via CDK mean that if we make a change in the future that replaces this, users whose url was created using the old/destroyed IAM but haven't accessed it yet will be unable to access?

src/app/settings.py
```diff
@@ -397,3 +397,9 @@ def show_toolbar(request):
     "AWS_REGION": AWS_REGION,
     "EXTRACTORS_CACHE_DIR": "/tmp/classifier",
 }
+
+# Reporting S3 credentials
+REPORT_S3_ACCESS_KEY_ID = os.environ.get("REPORT_S3_ACCESS_KEY_ID")
+REPORT_S3_SECRET_ACCESS_KEY = os.environ.get("REPORT_S3_SECRET_ACCESS_KEY")
+# if not REPORT_S3_ACCESS_KEY_ID or not REPORT_S3_SECRET_ACCESS_KEY:
```
saanobhaai, line 404: Ignore copilot -- I would just remove the commented lines

iac/stacks/common.py
ms280690: The presigned urls have a lifetime and then expire. The default is 3600 seconds https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/generate_presigned_url.html. If something changes the user can generate another url and should get a new presigned url.

src/app/settings.py
ms280690: done

PR #597, opened 2025-06-24
M1514-az-rebalance

michaelconnor00:
Enable AZ rebalancing on services that require HA.

closes M1514

## Summary by CodeRabbit

- **Chores**
  - Updated internal configuration to enable availability zone rebalancing for certain services.
  - Removed outdated commented configuration lines related to deployment circuit breakers.


michaelconnor00:
As coderabbit pointed out, this is not required as we only have one replica.

PR #598, opened 2025-06-26
[1559] update instead of bulk_create CR xfer to preserve image FKs

saanobhaai:


saanobhaai:
[IC benthic pqt SU transferred doesn't have photos or annotations](https://trello.com/c/F5SGaDmn/1559-ic-benthic-pqt-su-transferred-doesnt-have-photos-or-annotations)

gridcell (approved):

PR #613, opened 2025-09-18
[M1627] Add filters for ProjectSummarySampleEventViewSet

gridcell:
## Summary by CodeRabbit

- New Features
  - Unified read-only project sample event view with an explicit access field; serializer and endpoints now use this view and return results ordered by project ID.
  - Expanded filtering: country, site, management, tags, and JSON-record icontains lookups.

- Performance
  - Added GIN indexes on JSON records to speed JSON-based filtering/searches.

- Chores
  - Database migrations added to create the consolidated view; migration SQL may be resolved at runtime.

- Tests
  - Test setup updated to ensure the consolidated view is created for test runs.

### DIFF
--- a/src/app/settings.py
+++ b/src/app/settings.py
@@ -288,12 +288,12 @@ MC_API_KEY = os.environ.get("MC_API_KEY")
 MC_USER = os.environ.get("MC_USER")
 MC_LIST_ID = os.environ.get("MC_LIST_ID")
 
-
-sentry_sdk.init(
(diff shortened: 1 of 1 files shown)


gridcell:
@coderabbitai review

saanobhaai (changes requested):
Looks good, though maybe we should do a sanity-check comparison of results before and after the merge (against the prod db), with and without authentication.

Couple small comments -- and I think some of the coderabbit suggestions are valid.

PR #634, opened 2026-01-09
[M1746] add profile app settings as separate model/resource

saanobhaai:


saanobhaai:
[API support for Collect and CollectProject settings](https://trello.com/c/d8H20yRj/1746-api-support-for-collect-and-collectproject-settings)

PR #647, opened 2026-02-06
977 image build

ms280690:
https://trello.com/c/J40CNZYz/977-the-cdk-is-rebuilding-new-images-for-every-deploy-even-though-no-code-in-the-images-has-changed
for dev: Only generate the version file when there are changes to the relevant source code
for prod: Only generate the version file when a new tag is published 

## Summary by CodeRabbit

* **Chores**
  * Development deploy now conditionally updates the app version file only when code, dependency, Docker, or configuration changes are detected, reducing unnecessary pipeline updates and noise.
  * Production deploy now records releases using the branch/tag name and adds clearer release logging for improved traceability and auditability of production releases.


michaelconnor00 (changes requested):
I don't understand why you want to only create the VERSION.txt when there are changes? It is the commit sha in the file, so I would suspect you want to always create it so the the relevant commit is in the file. 

The CDK build runs regardless, so the image is going to build everytime. 

I think I need more context around the purpose of the change.

PR #659, opened 2026-03-31
1872 test ci for pr

ms280690:
https://trello.com/c/jssKGLoQ/1872-get-tests-running-in-github-actions
Run tests in the CI

## Summary by CodeRabbit

* **Tests**
  * CI now runs real Python tests with pytest and coverage (threshold enforced), uploads coverage artifacts, uses Python 3.12, and installs DB/geospatial system packages.
  * Added Postgres/PostGIS and local SQS service containers with health checks.
  * Test suite now forces S3 cache misses by default, with opt-in fixtures to simulate cache hits.

* **Configuration**
  * SQS endpoint can be set via an environment variable, with previous behavior as fallback.


.github/workflows/_test.yml
```diff
@@ -4,11 +4,98 @@ on: workflow_call
...
+      DEFAULT_DOMAIN_API: localhost:8080
+      DEFAULT_DOMAIN_COLLECT: localhost:8888
+      SQS_ENDPOINT_URL: http://localhost:9324
+      AWS_ACCESS_KEY_ID: testing
+      AWS_SECRET_ACCESS_KEY: testing
+      AWS_REGION: us-east-1
+
     steps:
-      - uses: actions/checkout@v6.0.2
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```
michaelconnor00: Why are using the hash for version here? `v6` should be enough.

.github/workflows/_test.yml
```diff
@@ -4,11 +4,98 @@ on: workflow_call
...
+              exit 0
+            fi
+            echo "Waiting for SQS... ($i)"
+            sleep 1
+          done
+          echo "SQS did not become ready"
+          exit 1
+
+      - name: Setup Python
+        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
```
michaelconnor00: Why are using the hash for version here? `v6` should be enough.

.github/workflows/_test.yml
```diff
@@ -4,11 +4,98 @@ on: workflow_call
...
+      - name: Install system dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends \
+            libpq-dev \
+            gdal-bin \
+            python3-gdal
+
+      - name: Cache pip packages
+        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
```
michaelconnor00: Why are using the hash for version here? `v5` should be enough.

.github/workflows/_test.yml
```diff
@@ -4,11 +4,98 @@ on: workflow_call
...
+
+      - name: Run tests
+        working-directory: src
+        run: pytest --no-migrations --cov --cov-report=term-missing --cov-report=xml:coverage.xml --cov-fail-under=80 -q
 
-      - name: Run Tests
-        run: echo "No tests"
+      - name: Upload coverage report
+        if: always()
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
```
michaelconnor00: Why are using the hash for version here? `v7` should be enough.

michaelconnor00 (changes requested):
Looks good. I added some comments around action versions. Please respond before I approve this.

.github/workflows/_test.yml
ms280690: https://github.com/data-mermaid/mermaid-api/pull/659#pullrequestreview-4039873314

.github/workflows/_test.yml
ms280690: https://github.com/data-mermaid/mermaid-api/pull/659#pullrequestreview-4039873314

.github/workflows/_test.yml
ms280690: https://github.com/data-mermaid/mermaid-api/pull/659#pullrequestreview-4039873314

.github/workflows/_test.yml
ms280690: https://github.com/data-mermaid/mermaid-api/pull/659#pullrequestreview-4039873314

michaelconnor00 (approved):

.github/workflows/_test.yml
```diff
@@ -4,11 +4,98 @@ on: workflow_call
...
+          key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
+          restore-keys: |
+            ${{ runner.os }}-pip-
+
+      - name: Install Python dependencies
+        run: pip install -r requirements.txt
+
+      - name: Run tests
+        working-directory: src
+        run: pytest --no-migrations --cov --cov-report=term-missing --cov-report=xml:coverage.xml --cov-fail-under=80 -q
```
saanobhaai: Shouldn't we use `api/tests` the way we do in our make command?

.github/workflows/_test.yml
```diff
@@ -4,11 +4,98 @@ on: workflow_call
...
+
+      - name: Upload coverage report
+        if: always()
+        uses: actions/upload-artifact@v7.0.1
+        with:
+          name: coverage-report
+          path: src/coverage.xml
```
saanobhaai: Why xml (as opposed to, say, html), and where does this go? Where and how would the output be used (by me, e.g.) to understand what failed?

src/api/utils/email.py
```diff
@@ -55,6 +55,8 @@ def send_mermaid_email(subject, template, to, context=None, from_email=None, rep
 def _to_in_dev_emails(to):
     to_emails = []
     for to_email in to:
+        if not to_email:
```
saanobhaai, line 58: This seems unrelated -- is this intentional? Why is it needed? I am hesitant to tamper with email logic.

.github/workflows/_test.yml
```diff
@@ -4,11 +4,98 @@ on: workflow_call
...
+          echo "SQS did not become ready"
+          exit 1
+
+      - name: Setup Python
+        uses: actions/setup-python@v6.2.0
+        with:
+          python-version: "3.12"
+
+      - name: Install system dependencies
+        run: |
```
saanobhaai, line 74: A lot of this seems un-DRY with respect to the docker file. If we change in one place we'll need to remember to change in another. But I'm guessing there's not an easy way to do anything about that.

src/api/tests/conftest.py
```diff
@@ -25,3 +27,26 @@ def django_db_setup(django_db_setup, django_db_blocker):
...
+    """Disable S3 cache by default; use `s3_cache_hit` fixture to test cache-hit paths."""
+    if "s3_cache_hit" not in request.fixturenames:
+        with patch("api.utils.cached.exists", return_value=False):
+            yield
+    else:
+        yield
+
+
+@pytest.fixture
+def no_s3_cache():
```
saanobhaai: 1. no_s3_cache is dead code in practice
Since _disable_s3_cache is autouse=True, every test that doesn't use s3_cache_hit already gets the cache disabled. The no_s3_cache fixture does nothing that _disable_s3_cache doesn't already do automatically. It can mislead a developer into thinking an explicit opt-in is needed.

2. no_s3_cache + s3_cache_hit in the same test is undefined
If a test uses both fixtures, _disable_s3_cache sees s3_cache_hit in request.fixturenames and yields without patching — deferring to s3_cache_hit to own the patch. But no_s3_cache still runs its own patch(..., return_value=False). The winner depends on fixture teardown order, which pytest doesn't guarantee between independent fixtures. You'd get a contradictory mock state.

In CI this probably never happens because the fixture set is controlled. Locally, a developer might combine them by accident (e.g., inheriting no_s3_cache from a base class while adding s3_cache_hit explicitly).

Fix: Drop no_s3_cache entirely — it adds confusion without value. If the intent was to have an explicit opt-out marker symmetric with the opt-in, document that autouse handles it, or at least add an assertion in no_s3_cache that s3_cache_hit is not also active.

saanobhaai (changes requested):

ms280690:
@saanobhaai 
1. Use api/tests like make — ✅ Fixed. Added api/tests to the pytest invocation so CI scopes collection
exactly like make test.

2. Why XML, where does it go, how to read failures — ✅ Clarified + improved. Reply for the PR:

Three things, now split clearly: test failures appear directly in the "Run tests" job log (the -q pytest output). Coverage had only xml (machine-readable, for tools like Codecov) — I added html:htmlcov so you can download the coverage-report artifact and open htmlcov/index.html to browse line-by-line coverage, plus term-missing already prints uncovered lines in the log. XML stays for tooling.

3. email.py:58 — intentional? why? — ✅ Keep; here's the why for the PR:

Intentional and required by this PR. Enabling the suite in CI exercises a path that was never hit before: creating a proposed attribute fires a post_save signal that emails SUPERUSER. In CI, SUPERUSER env is unset → settings.SUPERUSER[1] is None → _to_in_dev_emails([None]) did None.endswith(...) → AttributeError, failing the test. The guard skips falsy recipients — a None/empty address can never match a dev-email suffix anyway, so no valid recipient is ever dropped. It's input hygiene on the recipient list, not a behavior change to email sending.

4. _test.yml system deps un-DRY vs Dockerfile — ⏭️  Skip (deliberate). Reply:

Correct, and there's no clean DRY fix without running the suite inside the built Docker image (like make test does via docker exec). That would dedupe the apt/gdal/libpq setup but adds the image build to every CI run (slower). The native-runner duplication is the deliberate trade-off for fast CI. Happy to switch to container-based if you'd prefer DRY over speed — your call.

5. no_s3_cache dead code — ✅ Fixed. Removed it (confirmed zero references; _disable_s3_cache autouse already covers it, and the dual-fixture race you described can't happen if it doesn't exist).

saanobhaai (approved):

PR #663, opened 2026-04-01
Add WAF to ALB

ms280690:
https://trello.com/c/xMIT1njs/1877-add-aws-waf-to-the-application-load-balancer

```
❯ cdk diff mermaid-api-infra-common
start: Building mermaid-api-infra-common Template
success: Built mermaid-api-infra-common Template
start: Publishing mermaid-api-infra-common Template (554812291621-us-east-1-944106cd)
success: Published mermaid-api-infra-common Template (554812291621-us-east-1-944106cd)
Hold on while we create a read-only change set to get a diff with accurate replacement information (use --method=template to use a less accurate but faster template-only diff)

Stack mermaid-api-infra-common
Resources
[+] AWS::WAFv2::WebACL ApiWebAcl ApiWebAcl
[+] AWS::WAFv2::WebACLAssociation ApiWebAclAssociation ApiWebAclAssociation



✨  Number of stacks with differences: 1
```
WAF Count Mode
Count mode is an observability-only mode in AWS WAF. Rules evaluate incoming requests and record matches in CloudWatch metrics, but no traffic is blocked.

Why start in count mode?
1. Zero risk — legitimate API traffic is never disrupted
2. Visibility — CloudWatch metrics show which rules would fire and how often
3. Tuning — identify false positives before enforcing rules

How it's configured

| Rule                                             | Property         | Count value | Block value |
|--------------------------------------------------|------------------|-------------|-------------|
| Managed rule groups (Core, SQLi, Known Bad Inputs) | override_action | count={}    | none={}     |
| Custom rules (Rate limit)                         | action           | count={}    | block={}    |

Note: Managed rule groups use override_action because each rule inside the group already has its own action (block). Setting count={} overrides all of those to count-only. Setting none={} lets the group's built-in actions take effect.

Switching to block mode
After monitoring CloudWatch metrics and confirming no false positives:

## Summary by CodeRabbit

* **New Features**
  * Added regional Web Application Firewall (WAF) protection and associated it with the application load balancer.
  * Enabled CloudWatch visibility for WAF activity.
  * Deployed managed rule groups in counting mode (including common, SQLi, and known-bad inputs) alongside an IP-based rate-count rule set to 2,000 requests per IP.


iac/stacks/common.py
```diff
@@ -272,6 +273,94 @@ def __init__(
...
+        self.web_acl = wafv2.CfnWebACL(
+            self,
+            "ApiWebAcl",
+            name=f"{self.stack_name}-waf",
+            default_action=wafv2.CfnWebACL.DefaultActionProperty(allow={}),
+            scope="REGIONAL",
+            visibility_config=wafv2.CfnWebACL.VisibilityConfigProperty(
+                cloud_watch_metrics_enabled=True,
+                metric_name="MermaidApiWafMetric",
+                sampled_requests_enabled=False,
```
gridcell, line 581: **Warning** · `security` · confidence: high

**sampled_requests_enabled=False disables visibility — use DataProtectionConfig as the privacy-safe middle path**

All five `VisibilityConfigProperty` blocks set `sampled_requests_enabled=False`. This was set intentionally to address a prior review comment about privacy risk from storing unredacted request data, and that concern is valid. However, the result is that CloudWatch metrics show aggregate counts only — no URI, IP, header, or payload evidence is available to distinguish true positives from false positives, which makes it impossible to safely graduate any rule from COUNT to BLOCK mode.

AWS WAF now ships `DataProtectionConfig`, which hashes or substitutes nominated fields (`authorization`, `cookie`, `set-cookie`, `x-api-key`, sensitive body fields) before they appear in sampled requests or logs. This preserves the compliance property while restoring the observability needed to tune rules.

> **Suggestion:** Add a `DataProtectionConfig` at the WebACL level listing fields to hash/substitute, then set `sampled_requests_enabled=True` on at minimum the top-level `VisibilityConfig` and the `RateLimitPerIP` rule. Sampled requests are capped at 100 per 5-minute window and retained for 3 hours — no storage cost.

> **Why it matters:** COUNT mode is only a safe rollout strategy if matched requests can be inspected. Without sampling and without a `DataProtectionConfig`, the team cannot validate rule behavior and COUNT mode will be indefinitely extended, leaving the same attack surface as having no WAF.

<sub>Found by: scaffold, sentinel, forge · CR-001</sub>

iac/stacks/common.py
```diff
@@ -272,6 +273,94 @@ def __init__(
...
+                        ),
+                    ),
+                    visibility_config=wafv2.CfnWebACL.VisibilityConfigProperty(
+                        cloud_watch_metrics_enabled=True,
+                        metric_name="MermaidApiRateLimit",
+                        sampled_requests_enabled=False,
+                    ),
+                ),
+            ],
+        )
```
gridcell, line 649: **Warning** · `security` · confidence: high

**No WAF logging destination configured — only aggregate CloudWatch metric counters**

The WebACL has no `CfnLoggingConfiguration` associated with it. CloudWatch metrics tell you how many requests each rule counted, but provide no request-level detail. Combined with `sampled_requests_enabled=False`, there is currently no way to examine which specific requests triggered a rule. Incident response and COUNT-to-BLOCK graduation both require request-level evidence that neither metrics nor samples currently provide.

> **Suggestion:** Add a `wafv2.CfnLoggingConfiguration` whose `log_destination_configs` points to a CloudWatch Logs log group (name must begin with `aws-waf-logs-`) or a Kinesis Firehose delivery stream targeting S3. Use `redacted_fields` and/or a `logging_filter` to redact `authorization`, `cookie`, `set-cookie`, `x-api-key`, and body fields before they are stored, satisfying the same privacy constraint that motivated disabling sampling.

> **Why it matters:** Without per-request logs, security investigations after an incident and COUNT-mode tuning require guesswork. Adding a `CfnLoggingConfiguration` with field redaction provides the evidence trail while preserving compliance.

<sub>Found by: scaffold, sentinel · CR-002</sub>

iac/stacks/common.py
```diff
@@ -272,6 +273,94 @@ def __init__(
...
+            ),
+            rules=[
+                wafv2.CfnWebACL.RuleProperty(
+                    name="AWSManagedRulesCommonRuleSet",
+                    priority=1,
+                    override_action=wafv2.CfnWebACL.OverrideActionProperty(count={}),
+                    statement=wafv2.CfnWebACL.StatementProperty(
+                        managed_rule_group_statement=wafv2.CfnWebACL.ManagedRuleGroupStatementProperty(
+                            vendor_name="AWS",
+                            name="AWSManagedRulesCommonRuleSet",
```
gridcell, line 591: **Warning** · `security` · confidence: high

**All three managed rule group versions are unpinned — AWS can silently update rules**

All three `ManagedRuleGroupStatementProperty` blocks omit the `version` argument (same issue also applies to lines 310-312 and 326-328). AWS publishes new versions of managed rule groups on a rolling basis without advance notice. While rules are in COUNT mode the impact of a silent update is limited to additional false-positive counts, but once any rule is flipped to BLOCK, an unpinned version update can start blocking legitimate traffic without any change appearing in this repository.

> **Suggestion:** Pin each managed rule group to a specific version, for example `version="Version_1.12"` for `AWSManagedRulesCommonRuleSet`. Check the current stable version via `aws wafv2 list-available-managed-rule-group-versions`. Pair with a lightweight quarterly review to bump pins intentionally.

> **Why it matters:** Unpinned managed rule groups make production WAF posture dependent on AWS release timing. Once rules are in BLOCK mode, a silent update is indistinguishable from an application regression.

<sub>Found by: scaffold, sentinel · CR-003</sub>

iac/stacks/common.py
```diff
@@ -272,6 +273,94 @@ def __init__(
...
+        # WAFv2 WebACL — all rules start in COUNT mode for safe rollout
+        self.web_acl = wafv2.CfnWebACL(
+            self,
+            "ApiWebAcl",
+            name=f"{self.stack_name}-waf",
+            default_action=wafv2.CfnWebACL.DefaultActionProperty(allow={}),
+            scope="REGIONAL",
+            visibility_config=wafv2.CfnWebACL.VisibilityConfigProperty(
+                cloud_watch_metrics_enabled=True,
+                metric_name="MermaidApiWafMetric",
```
gridcell, line 580: **Warning** · `correctness` · confidence: high

**Hardcoded CloudWatch metric names will collide across environments**

`metric_name="MermaidApiWafMetric"` and all per-rule metric name strings (lines 301, 317, 333, 349) are fixed literals. `CommonStack` is shared infrastructure deployed to multiple environments. All environments emit WAF metrics to the same CloudWatch namespace under identical metric names, making it impossible to create environment-specific alarms or dashboards and making aggregate metrics ambiguous.

> **Suggestion:** Prefix every `metric_name` with `self.stack_name`, consistent with the `name=f"{self.stack_name}-waf"` pattern already used on line 280 for the WebACL name: `f"{self.stack_name}-WafMetric"`, `f"{self.stack_name}-CommonRules"`, etc.

> **Why it matters:** Environment-level WAF alerting is unreliable when all environments share the same CloudWatch metric name. A spike in dev will trigger prod alarms and vice versa.

<sub>Found by: forge · CR-004</sub>

iac/stacks/common.py
```diff
@@ -272,6 +273,94 @@ def __init__(
...
+                ),
+                wafv2.CfnWebACL.RuleProperty(
+                    name="RateLimitPerIP",
+                    priority=4,
+                    action=wafv2.CfnWebACL.RuleActionProperty(count={}),
+                    statement=wafv2.CfnWebACL.StatementProperty(
+                        rate_based_statement=wafv2.CfnWebACL.RateBasedStatementProperty(
+                            limit=2000,
+                            aggregate_key_type="IP",
+                        ),
```
gridcell, line 640: **Warning** · `security` · confidence: high

**Rate-based rule uses raw peer IP with no forwarded_ip_config — evasion risk if a proxy is ever added**

`RateLimitPerIP` uses `aggregate_key_type="IP"` with no `forwarded_ip_config`. When the ALB is the direct internet-facing endpoint this is correct — the peer IP is the client IP. However, if CloudFront or another proxy is ever placed in front of the ALB, every request will originate from the proxy's IP, collapsing all users into a single bucket (shared-NAT users hit the limit together) while bad actors behind the proxy evade the per-IP limit entirely. There is no code comment documenting this assumption.

> **Suggestion:** Add a code comment explicitly noting that this rule assumes the ALB is direct internet-facing and that the rule must be updated to `aggregate_key_type="FORWARDED_IP"` with a `ForwardedIPConfigurationProperty` if a CDN or proxy is ever added upstream. This makes the assumption visible to future infrastructure changes.

> **Why it matters:** A rate-limit rule keyed on the wrong field creates a compliance artifact without protection and risks self-inflicted outages for shared-NAT user populations if the network topology changes.

<sub>Found by: sentinel · CR-006</sub>

iac/stacks/common.py
```diff
@@ -272,6 +273,94 @@ def __init__(
...
+                    ),
+                    visibility_config=wafv2.CfnWebACL.VisibilityConfigProperty(
+                        cloud_watch_metrics_enabled=True,
+                        metric_name="MermaidApiRateLimit",
+                        sampled_requests_enabled=False,
+                    ),
+                ),
+            ],
```
gridcell, line 648: **Warning** · `maintainability` · confidence: high

**Deeply inlined rules literal (~65 lines) — extract to a config-driven helper**

The four rules are expressed as ~65 lines of deeply nested `CfnWebACL.RuleProperty` literals with near-identical structure (each managed rule group block differs only in `name`, `priority`, `vendor_name`, and `metric_suffix`). Adding a fifth rule group or changing `override_action` globally requires multiple coordinated edits. Combined with the metric_name collision fix (CR-004) and any future version pinning, each edit touches multiple near-identical blocks.

> **Suggestion:** Extract a private `_managed_rule_group(stack_name, name, vendor_name, priority, metric_suffix)` helper that builds a `RuleProperty` and iterate over a module-level `_MANAGED_RULES` config list. This reduces the rules block to ~10 lines and makes adding or modifying a rule group a single-line config change.

> **Why it matters:** Inlined repetitive CDK literals are a common source of copy-paste drift: a change intended for all rules misses one block, introducing subtle behavioral differences.

<sub>Found by: forge · CR-007</sub>

iac/stacks/common.py
```diff
@@ -272,6 +273,94 @@ def __init__(
...
+                        sampled_requests_enabled=False,
+                    ),
+                ),
+                wafv2.CfnWebACL.RuleProperty(
+                    name="RateLimitPerIP",
+                    priority=4,
+                    action=wafv2.CfnWebACL.RuleActionProperty(count={}),
+                    statement=wafv2.CfnWebACL.StatementProperty(
+                        rate_based_statement=wafv2.CfnWebACL.RateBasedStatementProperty(
+                            limit=2000,
```
gridcell, line 638: **Question** · `docs` · confidence: medium

**Rate-limit threshold of 2000 req/5min — what is the traffic baseline this was derived from?**

The `RateLimitPerIP` rule uses `limit=2000`, which is also AWS WAF's minimum allowed value. There is no comment explaining whether this was derived from observed ALB traffic baselines or chosen as the floor because the team was uncertain. For a scientific API that handles bulk field-device uploads and offline mobile sync (Push/Pull), a per-IP rate of 2000 requests per 5 minutes could be simultaneously too tight for legitimate bulk sync clients and too permissive to deter SQLi probing.

> **Suggestion:** Sample ALB access logs over a 7-day representative window and compute p95/p99 5-minute request rates per source IP. Document the derived threshold in a comment. Consider two separate rate rules at different path scopes (tighter on `/auth/*`, looser on `/push|pull/sync`) if traffic profiles differ significantly by endpoint.

> **Why it matters:** An unjustified threshold cannot be safely adjusted or defended in an audit. Setting `limit=2000` because it is the minimum is a different decision from setting it because observed traffic supports it.

<sub>Found by: scaffold, sentinel · CR-008</sub>

iac/stacks/common.py
```diff
@@ -272,6 +273,94 @@ def __init__(
             certificate_arn=f"arn:aws:acm:us-east-1:{self.account}:certificate/783d7a91-1ebd-4387-9518-e28521086db6",
```
gridcell, line 568: **Info** · `security` · confidence: low

**Pre-existing: manually-created ACM certificate ARN is out of CDK management**

This line references an ACM certificate by a hardcoded ARN with a comment noting it was created manually. This is pre-existing and not introduced by this PR — flagged only because the new WAF code sits immediately adjacent. The concern is operational: a manually-managed certificate can be renewed, rotated, or revoked without CDK noticing, which is a common cause of certificate expiry incidents.

> **Suggestion:** Follow-up ticket to either manage the certificate in CDK using `acm.Certificate` with DNS validation and automatic renewal, or move the ARN to an SSM Parameter and resolve it at synth time so it can be updated without a code change.

> **Why it matters:** Manually-managed production certificates are a common cause of expiry-driven outages and are invisible to infrastructure drift detection.

<sub>Found by: sentinel · CR-011</sub>

gridcell (commented):
## AI Code Review

**Assessment:** The WAFv2 WebACL is structurally sound — REGIONAL scope is correct for an ALB, the three AWS managed rule groups plus a per-IP rate limit cover the expected baseline threat surface, and deploying every rule in COUNT mode is the right staged-rollout pattern. However, as deployed the WAF is operationally blind: no logging destination, unpinned managed rule versions, `sampled_requests_enabled=False` everywhere (set intentionally for privacy — the `DataProtectionConfig` middle path is the recommended resolution), CloudWatch metric names will collide across environments, and there is no documented graduation plan. None of these are hard blockers, but the combination means the WAF provides observability data that cannot be acted on, and COUNT mode risks becoming permanent.

| 0 | 7 | 2 | 2 | 2 |

**Files reviewed:** 1 / 1

### Cross-Cutting Concerns
- **No exit path from COUNT mode today.** `sampled_requests_enabled=False` (CR-001) + no `CfnLoggingConfiguration` (CR-002) + no documented graduation criteria (CR-005) collectively mean the WAF cannot be tuned, cannot be graduated to BLOCK mode, and cannot support incident response. Resolving CR-001 via `DataProtectionConfig` + sampling and CR-002 via a logging configuration would unblock all downstream graduation work.
- **No alerting surface.** The rate-limit threshold (CR-008) and the absence of CloudWatch alarms or a dashboard are related: even after logging and sampling are enabled, there is no alerting to notify the team when rules match at unusual rates. CloudWatch alarms on each rule's count metric should be a prerequisite before any BLOCK flip.
- All findings are in infrastructure configuration — no application-layer (Django) code was modified.

<details>
<summary><strong>CR-005</strong> &middot; Warning &middot; <code>docs</code> — No COUNT-to-BLOCK graduation plan — COUNT mode risks becoming permanent</summary>

The inline comment correctly states 'all rules start in COUNT mode for safe rollout', but there is no documented observation period, no success criteria for graduation, no tracking ticket reference, and no CDK context variable to control `override_action`. Without a graduation plan, COUNT mode tends to persist indefinitely: nobody owns the decision to flip, the alert threshold has not been set, and the WAF becomes a permanent monitoring artifact rather than a protection mechanism.

> **Suggestion:** Add a comment block that documents: (1) the minimum observation window before graduation is considered (e.g., 30 days), (2) the criteria (e.g., false-positive rate below X% for managed rules, alert on `RateLimitPerIP` metric), (3) a link to a tracking ticket. Optionally introduce a CDK context variable (e.g., `cdk.json: waf_mode: 'count' | 'block'`) to make the flip a one-line deploy change rather than a code edit.

> **Why it matters:** A WAF permanently stuck in COUNT mode provides no blocking protection. The inline comment makes intent clear but creates no accountability for follow-through.

Applies to: `iac/stacks/common.py`

<sub>Found by: scaffold, sentinel</sub>
</details>

<details>
<summary><strong>CR-009</strong> &middot; Question &middot; <code>correctness</code> — Rule priority ordering: managed rules (1-3) before rate limit (4) — is rate limiting first more appropriate?</summary>

WAF evaluates rules in ascending priority order, so the three managed rule groups fire before the rate-limit rule. This means a high-volume attacker's requests are evaluated against all three managed rule groups before the rate limit terminates evaluation. Placing the rate-limit rule at priority 1 would short-circuit managed-rule evaluation for traffic already destined to be counted or blocked, which can reduce WCU consumption at scale.

> **Suggestion:** Confirm whether rate limiting first (priority 1) is the intended evaluation order. For a low-traffic scientific API this is unlikely to matter on cost, but the ordering should be a deliberate choice documented in a comment.

Applies to: `iac/stacks/common.py`

<sub>Found by: scaffold</sub>
</details>

<details>
<summary><strong>CR-010</strong> &middot; Info &middot; <code>test_gap</code> — No CDK snapshot or integration tests for the new WebACL construct</summary>

The `iac/` directory has no `test/` subdirectory. The WAFv2 WebACL and its association with the ALB are a significant infrastructure addition — a misconfiguration (e.g., accidental removal of the `WebACLAssociation`, a rule priority collision, or a metric name regression) would be invisible until a deploy or an audit.

> **Suggestion:** Add `iac/tests/test_common_stack.py` using `aws_cdk.assertions.Template`. At minimum assert: (1) one `CfnWebACL` resource exists with `scope=REGIONAL`, (2) one `CfnWebACLAssociation` exists referencing the ALB ARN, (3) exactly four rules are present with the expected priorities, (4) `default_action` is `allow` (not `block`).

Applies to: `iac/stacks/common.py`

<sub>Found by: scaffold</sub>
</details>

<details>
<summary><strong>CR-012</strong> &middot; Praise &middot; <code>security</code> — Correct WAFv2 construct structure and responsible COUNT-mode staged rollout</summary>

`scope=REGIONAL` is correct for an ALB-facing WAF. `self.web_acl.attr_arn` is the right attribute for the `WebACLAssociation`. `default_action=allow({})` correctly delegates blocking to individual rules rather than defaulting to deny. The distinction between `override_action` (used for managed rule groups) and `action` (used for the custom rate-based rule) is correctly applied — this is a common CDK/WAFv2 mistake that was avoided here. The inline comment documenting the COUNT-mode intent makes the rollout strategy clear to reviewers.

Applies to: `iac/stacks/common.py`

<sub>Found by: scaffold, sentinel, forge</sub>
</details>

<details>
<summary><strong>CR-013</strong> &middot; Praise &middot; <code>maintainability</code> — CfnWebACL (L1) is the correct construct choice</summary>

aws-cdk v2 ships no L2 constructs for WAFv2, making `CfnWebACL` the only available option. The `aws_wafv2` import is correctly placed in alphabetical order with the other `aws_cdk` imports. This is a detail that is easy to get wrong when following general CDK guidance that encourages L2 constructs.

Applies to: `iac/stacks/common.py`

<sub>Found by: forge</sub>
</details>

### Agent Coverage
| scaffold | infrastructure | 1 | 9 | 0 |
| sentinel | security | 1 | 9 | 0 |
| forge | backend | 1 | 5 | 0 |
| polisher | code_quality | 1 | 0 | 0 |

michaelconnor00 (commented):
Please click `Resolve Conversation` on all AI comments once you have reviewed them.

The WAF implementation looks good. 

What is the cost estimate for adding the WAF. If Cloudwatch metrics are being created, there will likely be a cost for that.

ms280690:
> Please click `Resolve Conversation` on all AI comments once you have reviewed them.
> 
> The WAF implementation looks good.
> 
> What is the cost estimate for adding the WAF. If Cloudwatch metrics are being created, there will likely be a cost for that.

Cost impact: +$6-12/mo (base) + $1/million requests

michaelconnor00 (approved):

PR #664, opened 2026-04-01
1888 multistage docker build

ms280690:
https://trello.com/c/KgGfshd9/1888-convert-dockerfile-to-multi-stage-build

## Multi-stage Docker build

Converts the single-stage Dockerfile into a 3-stage build (`builder` → `runtime` → `dev`) to reduce image size and attack surface.

### What changed

- **Dockerfile**: Split into 3 stages:
  - `builder` — installs build deps (`build-essential`, `libpq-dev`, `python3-dev`) and compiles pip packages; discarded after build
  - `runtime` — production image with only runtime OS deps (`postgresql-client-16`, `gdal-bin`, `python3-gdal`) and pre-built pip packages copied from builder
  - `dev` — extends `runtime` with test/dev deps for local development
- **Dockerfile**: Install CPU-only PyTorch (`--index-url https://download.pytorch.org/whl/cpu`) since ECS runs on `t3a` instances with no GPU. Saves ~1.5 GB over the default CUDA-enabled wheel.
- **Dockerfile**: Strip `__pycache__/`, `tests/`, `*.pyc`, `*.pyo` from installed packages. Purge `wget` after use.
- **requirements.txt**: Moved `pytest`, `pytest-cov`, `pytest-django` to requirements-dev.txt.
- **docker-compose.yml**: Added `target: dev` so local builds include test tools.
- **iac/stacks/api.py**: Added `target="runtime"` to `DockerImageAsset` so production deploys use the lean runtime stage.
- **.dockerignore**: Expanded to exclude `docs/`, `examples/`, `readme/`, `*.md`, `LICENSE`, `.env`, `.secrets.env`, and the Insomnia collection.

### Image size

❯ docker images
REPOSITORY        TAG       IMAGE ID       CREATED         SIZE
wcs/mermaid-api   latest    a703e2846627   6 minutes ago   2.01GB

### What's preserved

- Base image: `python:3.12-slim-bookworm`
- Non-root `webapp` user
- Runtime deps: GDAL, PostgreSQL client
- `collectstatic` build step
- docker-entry.sh CMD
- All paths (`/var/projects/webapp`, `/home/webapp/.local/bin`)

### CDK diff

Image hash update only — no infrastructure, IAM, networking, or config changes. All 5 ECS task definitions (API service, scheduled backup, summary cache, general worker, image processing worker) reference the new image.

### Migration risk

Low — backward compatible. All runtime paths, environment variables, and the entrypoint are unchanged.

## Summary by CodeRabbit

* **Chores**
  * Optimized container builds using multi-stage images, reduced build context, and smaller runtime images to speed builds and shrink artifacts.
  * Upgraded the Python runtime to 3.13 and updated CI to use the new runtime.
  * Adjusted compose/build targets to use the new build stages.

* **Tests / Dev tooling**
  * Moved test-related packages into development dependencies and removed them from production requirements.


Dockerfile
```diff
@@ -1,4 +1,49 @@
...
+COPY requirements.txt .
+# Pre-install CPU-only PyTorch before requirements.txt so pyspacer
+# (which depends on torch) picks up the lighter wheel (~280 MB vs ~2 GB).
+# ECS tasks run on t3a instances with no GPU.
+# Versions pinned to match pyspacer==0.12.0 constraints (torch>=2.6,<2.7).
+RUN su ${APP_USER} -c "\
+    pip install --upgrade pip \
+ && pip install --no-cache-dir --no-compile \
+        torch==2.6.0 torchvision==0.21.0 --index-url https://download.pytorch.org/whl/cpu \
+ && pip install --no-cache-dir --no-compile -r requirements.txt"
```
gridcell: **⚠️ Warning** · `dependency` · confidence: high

**PyTorch installed from alternate index without hash pinning and not tracked in requirements.txt**

`torch==2.6.0` and `torchvision==0.21.0` are installed directly in the Dockerfile from `https://download.pytorch.org/whl/cpu` without `--require-hashes`. They are **not recorded in `requirements.txt`**, so SCA tooling pointed at requirements.txt will not see them. A compromised endpoint, MITM with a trusted CA, or a supply-chain attack could yield a malicious wheel executed during every build. Availability is also tied to a third-party endpoint — an outage blocks all deploys.

> **Suggestion:** Pin wheel hashes via `--hash=sha256:...` for both torch and torchvision, or move into `requirements.txt` with hashes and use `pip-compile` / `uv lock`. Consider vendoring wheels to S3 to remove the external availability dependency.

> **Why it matters:** pyspacer transitively imports torch on reachable code paths, so a compromised wheel executes during application requests.

<sub>Found by: sentinel, scaffold, forge · CR-002 · CWE-494, CWE-829, OWASP A06:2021, A08:2021</sub>

Dockerfile
```diff
@@ -10,22 +55,17 @@ ENV PATH="/home/webapp/.local/bin:${PATH}"
...
 
-# Install OS dependencies
+# Install runtime-only OS deps (no build-essential, libpq-dev, python3-dev)
 RUN apt-get update && apt-get install -y --no-install-recommends \
-    wget gnupg lsb-release ca-certificates \
- && echo "deb http://apt.postgresql.org/pub/repos/apt $(lsb_release -cs)-pgdg main" > /etc/apt/sources.list.d/pgdg.list \
- && wget --quiet -O - https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add - \
+    wget ca-certificates \
+ && wget --quiet -O /usr/share/keyrings/pgdg.asc https://www.postgresql.org/media/keys/ACCC4CF8.asc \
+ && echo "deb [signed-by=/usr/share/keyrings/pgdg.asc] https://apt.postgresql.org/pub/repos/apt bookworm-pgdg main" > /etc/apt/sources.list.d/pgdg.list \
```
gridcell, line 68: **⚠️ Warning** · `security` · confidence: medium

**PostgreSQL signing key fetched over HTTPS without independent fingerprint verification**

The runtime stage fetches the PostgreSQL APT signing key and writes it to `/usr/share/keyrings/pgdg.asc`, then uses it via `signed-by=`. Better than the deprecated `apt-key add`, but the build still trusts whatever the endpoint serves. A key swap via upstream compromise, MITM, or DNS hijack would allow malicious PGDG packages to install as root during image build.

> **Suggestion:** Pin the expected fingerprint and verify after download (e.g., via `gpg --with-fingerprint` and grep for `B97B0AFCAA1A47F044F244A07FCC7D46ACCC4CF8`), or commit the key file to the repo and `COPY` it in — eliminating the network fetch entirely.

> **Why it matters:** The base image build chain is a high-value supply-chain target. A compromised key could yield a malicious runtime image deployed to ECS with full DB and S3 credential access.

<sub>Found by: sentinel · CR-003 · CWE-494, CWE-347, OWASP A08:2021</sub>

Dockerfile
```diff
@@ -1,4 +1,49 @@
...
+    build-essential \
+    libpq-dev \
+    python3-dev \
+ && rm -rf /var/lib/apt/lists/*
+
+ARG APP_USER=webapp
+RUN groupadd ${APP_USER} && useradd -m --no-log-init -g ${APP_USER} ${APP_USER}
+
+WORKDIR /var/projects/${APP_USER}
+COPY requirements.txt .
```
gridcell, line 20: **⚠️ Warning** · `correctness` · confidence: high

**UID drift between builder and runtime stages could cause silent ownership mismatches**

`groupadd` / `useradd` runs independently in both the builder and runtime stages. Both currently get UID 1000, but this is not guaranteed across future base-image updates. The `--chown=webapp:webapp` flag on `COPY --from=builder` remaps at copy time, so ownership is correct today — but the fragility is undocumented.

> **Suggestion:** Add `ARG APP_UID=1000` and use `useradd --uid ${APP_UID}` in both stages to make the UID deterministic and explicit.

> **Why it matters:** Silent UID drift between stages could cause permission errors in the final image that only manifest at runtime.

<sub>Found by: scaffold · CR-005</sub>

Dockerfile
```diff
@@ -53,3 +91,15 @@ USER ${APP_USER}:${APP_USER}
...
 CMD ["/var/projects/webapp/docker-entry.sh"]
+
+# ============================================================
+# Stage 3: Dev — adds test/dev dependencies on top of runtime
+# ============================================================
+FROM runtime AS dev
+
+ARG APP_USER=webapp
+USER root
+COPY requirements-dev.txt /tmp/requirements-dev.txt
```
gridcell, line 108: **⚠️ Warning** · `correctness` · confidence: medium

**Dev stage pip invocation via `su webapp` may not resolve pip from webapp's PATH**

In the dev stage, root calls `su webapp -c "pip install ..."` without a login shell. The `ENV PATH=/home/webapp/.local/bin:$PATH` directive may not be inherited inside the `su` subshell. If pip is not resolved correctly, the build either silently falls back to system pip or fails.

> **Suggestion:** Use `su -l ${APP_USER} -c "pip install ..."` (login shell) or the explicit path `/home/${APP_USER}/.local/bin/pip`.

> **Why it matters:** If pip resolves incorrectly, `docker compose build` breaks local developer setup without a clear error message.

<sub>Found by: scaffold · CR-006</sub>

Dockerfile
```diff
@@ -1,4 +1,49 @@
...
+ && pip install --no-cache-dir --no-compile -r requirements.txt"
+
+# Strip unnecessary files from installed packages to shrink the layer
+RUN find /home/${APP_USER}/.local -type d -name '__pycache__' -exec rm -rf {} + \
+ && find /home/${APP_USER}/.local -type d -name 'tests' \
+        ! -path '*/django/contrib/admin/tests' \
+        ! -path '*/pandas/tests*' \
+        ! -path '*/pandas/_testing*' \
+        -exec rm -rf {} + \
+ && find /home/${APP_USER}/.local -name '*.pyc' -delete \
```
gridcell, line 40: **⚠️ Warning** · `correctness` · confidence: medium

**Test-directory pruning allowlist may be too narrow — psycopg and pyarrow not included**

The `find`-based cleanup exempts `django/contrib/admin/tests`, `pandas/tests*`, and `pandas/_testing*`, but other packages such as `psycopg` and `pyarrow` sometimes import from their own `.tests` modules at import time. The hand-curated allowlist will silently rot as packages update. The current smoke test covers torch/torchvision/pandas/spacer but not the full runtime dep chain.

> **Suggestion:** Add `psycopg` and `pyarrow` to the allowlist. More robustly, add a CI step that builds the runtime image and runs `python -c "import django; import psycopg; import pyarrow; import spacer"` or the full test suite against the built image.

> **Why it matters:** Silently deleted test directories can cause confusing ImportErrors in production that are difficult to trace back to the image build step.

<sub>Found by: scaffold, sentinel · CR-007</sub>

iac/stacks/api.py
```diff
@@ -181,6 +181,7 @@ def get_secret_object(stack: Stack, secret_name: str):
             "ApiImage",
             directory="../",
             file="Dockerfile",
+            target="runtime",
```
gridcell, line 206: **⚠️ Warning** · `correctness` · confidence: high

**CDK shared image asset — target change affects all 5 ECS tasks (undocumented coupling)**

All 5 ECS task definitions (API, ScheduledBackupTask, SummaryCacheTask, QueueWorker General, QueueWorker ImageProcess) share this single `DockerImageAsset`. Adding `target="runtime"` applies to all of them simultaneously. This coupling is not documented — a future maintainer could change the target believing it affects only one workload when it actually affects every deployed service.

> **Suggestion:** Add a comment at the `DockerImageAsset` definition explicitly noting that this asset is shared across all ECS task definitions and that target changes affect all workloads.

> **Why it matters:** Undocumented coupling between a shared CDK asset and multiple ECS services is a maintenance hazard that could cause unintended production changes.

<sub>Found by: scaffold · CR-008</sub>

Dockerfile
```diff
@@ -1,4 +1,49 @@
...
+# Pre-install CPU-only PyTorch before requirements.txt so pyspacer
+# (which depends on torch) picks up the lighter wheel (~280 MB vs ~2 GB).
+# ECS tasks run on t3a instances with no GPU.
+# Versions pinned to match pyspacer==0.12.0 constraints (torch>=2.6,<2.7).
+RUN su ${APP_USER} -c "\
+    pip install --upgrade pip \
+ && pip install --no-cache-dir --no-compile \
+        torch==2.6.0 torchvision==0.21.0 --index-url https://download.pytorch.org/whl/cpu \
+ && pip install --no-cache-dir --no-compile -r requirements.txt"
+
```
gridcell, line 30: **❓ Question** · `correctness` · confidence: medium

**pip --upgrade pip in builder — is leaving system pip at its base version intentional?**

The first `pip install --upgrade pip` runs as webapp via `su`, upgrading `~/.local/bin/pip` only. System pip remains at the base image version. This is likely intentional (all subsequent pip work is user-local), but worth confirming since system pip could be targeted if anything later calls it directly.

<sub>Found by: scaffold · CR-010</sub>

Dockerfile
```diff
@@ -1,4 +1,49 @@
...
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    git \
+    build-essential \
+    libpq-dev \
+    python3-dev \
+ && rm -rf /var/lib/apt/lists/*
+
+ARG APP_USER=webapp
```
gridcell, line 15: **❓ Question** · `maintainability` · confidence: medium

**`git` installed in builder but absent from runtime — confirm no VCS pip dependencies**

Builder installs `git` and runtime omits it. No VCS URLs (`git+https://`) are present in `requirements.txt`, so this is consistent. Flagged for visibility so it is not forgotten if a VCS-sourced package is ever added (the runtime stage would then be missing git).

<sub>Found by: scaffold · CR-011</sub>

Dockerfile
```diff
@@ -1,4 +1,49 @@
...
+# Smoke-test: verify top-level dependencies import successfully after cleanup
+RUN su ${APP_USER} -c "python -c 'import django; import torch; import torchvision; import pandas; import spacer'"
+
+# ============================================================
+# Stage 2: Runtime — lean production image
+# ============================================================
+FROM python:3.13-slim-bookworm AS runtime
 LABEL maintainer="<sysadmin@datamermaid.org>"
 
 ENV DEBIAN_FRONTEND=noninteractive
```
gridcell, line 52: **ℹ️ Info** · `maintainability` · confidence: high

**No HEALTHCHECK instruction in runtime or dev stages**

Docker `HEALTHCHECK` is absent from the Dockerfile. The API service is covered by ALB `/health/` at the load balancer level, but the `ScheduledBackupTask` and `SummaryCacheTask` ECS tasks have no container-level health signal.

> **Suggestion:** Add a `HEALTHCHECK CMD` against `/health/` (or a task-appropriate signal) in the runtime stage so Docker and ECS can detect a hung or partially-started container.

<sub>Found by: scaffold · CR-012</sub>

Dockerfile
```diff
@@ -1,4 +1,49 @@
...
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    git \
+    build-essential \
+    libpq-dev \
+    python3-dev \
+ && rm -rf /var/lib/apt/lists/*
+
+ARG APP_USER=webapp
+RUN groupadd ${APP_USER} && useradd -m --no-log-init -g ${APP_USER} ${APP_USER}
```
gridcell: **ℹ️ Info** · `security` · confidence: medium

**Duplicate webapp user/group creation across builder and runtime stages is benign but worth noting**

`groupadd` + `useradd` runs in both stages. `COPY --from=builder --chown=webapp:webapp` re-chowns at copy time, so file ownership is correct. Benign today but documented here for visibility in the context of the UID drift warning (CR-005).

<sub>Found by: sentinel · CR-013</sub>

gridcell (dismissed):
## AI Code Review

**Assessment:** The multi-stage Docker refactor is architecturally sound and a meaningful security improvement: smaller runtime attack surface, dev tooling cleanly separated, deprecated apt-key replaced, and .env files excluded from build context. However, the PR is **blocked by a confirmed Django 4.2 / Python 3.13 incompatibility** — Django 4.2 LTS officially supports only Python 3.10–3.12. The base-image upgrade must not merge until either Django is upgraded to 5.1+ or the base image is reverted to 3.12.

| 1 | 7 | 3 | 4 | 6 |

**Files reviewed:** 6 / 6

### Cross-Cutting Concerns
- **Supply chain risk** spans multiple files: the unverified PostgreSQL APT key, unhashed PyTorch wheels from `download.pytorch.org`, and the absence of torch/torchvision from `requirements.txt` mean neither pip-audit nor any SCA tool will catch a compromise of either upstream. A single supply-chain policy is needed: vendor critical wheels to controlled S3 or enforce `--require-hashes` across all external sources.
- **Python 3.13 migration readiness** must be validated end-to-end. The Django 4.2 blocker (CR-001) is the immediate gate, but the full stack — djangorestframework-gis, psycopg, pyarrow, pyspacer, python-jose — should be exercised against the chosen 3.13 + Django 5.x combination on a dedicated test branch before merging.
- **No CI gate on the `runtime` stage**: `docker-compose.yml target: dev` and `iac/stacks/api.py target='runtime'` are correctly split, but there is no CI job that builds the `runtime` stage and runs a smoke test. This leaves a gap between local developer builds and the image that ships to ECS.

---

### Additional findings (not anchored to diff lines)

<details>
<summary><strong>CR-001</strong> &middot; 🔴 Severe &middot; <code>dependency</code> — Django 4.2 does not support Python 3.13 — production will break at runtime</summary>

**BLOCKING.** `requirements.txt` pins `Django==4.2.26`. Django 4.2 LTS officially supports Python 3.10, 3.11, and 3.12 only. Django 5.1 is the first release to officially support Python 3.13. The Dockerfile base-image change to `python:3.13-slim-bookworm` therefore creates a known-incompatible combination that can cause silent data corruption, crashes in ORM internals, and security patches that never land — the Django security team only backports to supported Python versions.

> **Suggestion:** Either (a) upgrade Django to 5.1.x or 5.2.x LTS and audit dependent packages for 5.x compatibility, or (b) revert the base image to `python:3.12-slim-bookworm` until the Django upgrade is complete. Run the full test suite on the chosen combination before merging.

> **Why it matters:** An unsupported CPython/Django combination can cause silent data corruption, crashes in ORM internals, or security patches that never reach the deployed version.

Evidence:
- `requirements.txt` line 3: `Django==4.2.26`
- [Django 4.2 FAQ](https://docs.djangoproject.com/en/4.2/faq/install/#what-python-version-can-i-use-with-django): Python compatibility 3.10, 3.11, 3.12
- [Django 5.1 release notes](https://docs.djangoproject.com/en/5.1/releases/5.1/): Python 3.13 support added

Applies to: `requirements.txt`

<sub>Found by: forge, sentinel &middot; CWE-1104, OWASP A06:2021</sub>
</details>

<details>
<summary><strong>CR-004</strong> &middot; ⚠️ Warning &middot; <code>dependency</code> — Audit Django ecosystem packages for Python 3.13 and Django 5.x compatibility before upgrading</summary>

Once the Django 4.2 / Python 3.13 blocker (CR-001) is resolved, several ecosystem packages need explicit validation: `djangorestframework==3.16.0`, `djangorestframework-gis==1.2.0` (historically lags major Django releases), `django-extensions==4.1`, `django-maintenance-mode`, `django-import-export`, `psycopg==3.2.9`.

> **Suggestion:** Run the full test suite on Python 3.13 + Django 5.1 in a dedicated branch. Pay particular attention to djangorestframework-gis (GIS serialization), psycopg (async cursor behavior changes), and django-maintenance-mode.

> **Why it matters:** GIS serialization and ORM layer breakages are hard to diagnose in production and may only surface under specific data conditions.

Applies to: `requirements.txt` (lines 4-14)

<sub>Found by: forge</sub>
</details>

<details>
<summary><strong>CR-009</strong> &middot; ❓ Question &middot; <code>dependency</code> — CI workflows pin Python 3.12 — should they align with the app runtime (3.13)?</summary>

`.github/workflows/deploy-cdk.yml` and `pr.yml` still specify Python `^3.12.0`. CDK synthesis itself does not need to match the application runtime, but if any pre-synth step imports application modules, it could fail or produce incorrect results after the base image moves to 3.13.

Applies to: `.github/workflows/deploy-cdk.yml`, `.github/workflows/pr.yml`

<sub>Found by: scaffold</sub>
</details>

<details>
<summary><strong>CR-014</strong> &middot; ℹ️ Info &middot; <code>maintainability</code> — Commented-out deployment circuit breaker lines remain in CDK stack</summary>

Two commented-out `circuit_breaker=ecs.DeploymentCircuitBreaker(enable=True, rollback=True)` lines remain in `iac/stacks/api.py` around lines 245-246. Not introduced by this PR but flagged for cleanup. If intentionally disabled, a comment should explain why; otherwise enable it.

Applies to: `iac/stacks/api.py`

<sub>Found by: scaffold</sub>
</details>

<details>
<summary><strong>CR-015</strong> &middot; ℹ️ Info &middot; <code>dependency</code> — pytest move to requirements-dev.txt confirmed safe</summary>

A grep of `src/` confirms all 43 files matching `import pytest` are located under `src/api/tests/` or `src/sqltables/tests/`. No application code, management commands, signals, middleware, or settings modules import pytest. The move is safe and correct.

Applies to: `requirements.txt`, `requirements-dev.txt`

<sub>Found by: forge</sub>
</details>

---

### Praise

- **CR-016** (`security`) — Multi-stage build cleanly removes build toolchain and dev dependencies from the production image. Only compiled site-packages are copied into runtime. *Found by: sentinel, scaffold*
- **CR-017** (`security`) — Deprecated `apt-key` replaced with scoped `signed-by` keyring using the explicit `bookworm-pgdg` suite and https:// repo URL (Debian best practice). *Found by: sentinel, scaffold*
- **CR-018** (`security`) — `.dockerignore` explicitly excludes `.env` and `.secrets.env` from the build context (and the new `**/__pycache__` glob is more comprehensive). *Found by: sentinel*
- **CR-019** (`security`) — IaC correctly pins production image to the `runtime` stage via `target="runtime"`; dev tooling never ships to any of the 5 ECS tasks. *Found by: sentinel, forge*
- **CR-020** (`correctness`) — Builder smoke test validates torch/torchvision/pandas/spacer imports **as the webapp user** after cleanup — matches production runtime user and catches pruning breakage early. *Found by: scaffold*
- **CR-021** (`maintainability`) — `ADD` replaced with `COPY` throughout — more explicit and avoids `ADD`'s implicit URL-fetch / tar-extract behaviors. *Found by: scaffold*

---

### Agent Coverage

| sentinel | security | 6 | 10 | 0 |
| scaffold | infrastructure | 6 | 15 | 0 |
| forge | backend | 6 | 6 | 1 |

Dockerfile
```diff
@@ -1,4 +1,52 @@
-FROM python:3.12-slim-bookworm
+# ============================================================
+# Stage 1: Builder — install build deps and compile pip pkgs
+# ============================================================
+FROM python:3.13-slim-bookworm AS builder
```
gridcell, line 4: **Severe** · `dependency` · confidence: high

**Django 4.2 does not support Python 3.13 — the base-image bump creates a known-incompatible combination**

`requirements.txt` pins `Django==4.2.26`. Django 4.2 LTS officially supports Python 3.10, 3.11, and 3.12 only — Python 3.13 support was added in Django 5.1. Moving the base image to `python:3.13-slim-bookworm` therefore creates a combination that is explicitly unsupported by upstream Django. The Django security team does not backport fixes to unsupported Python versions, and subtle ORM/C-extension behaviour can diverge silently.

> **Suggestion:** Either (a) upgrade Django to 5.1.x or 5.2.x LTS and audit dependent packages (djangorestframework-gis, django-extensions, django-import-export, django-maintenance-mode, psycopg) for 5.x compatibility, or (b) revert the base image to `python:3.12-slim-bookworm` until the Django upgrade is complete. Run the full test suite on the chosen combination before merging.

> **Why it matters:** An unsupported CPython/Django combination can cause silent data corruption, crashes in ORM internals, or security patches that never reach the deployed version. This is a hard compatibility wall documented by Django upstream.

References: [Django 4.2 Python compatibility](https://docs.djangoproject.com/en/4.2/faq/install/#what-python-version-can-i-use-with-django), [Django 5.1 release notes](https://docs.djangoproject.com/en/5.1/releases/5.1/)

<sub>Found by: verdict (post-aggregation) · CR-001 · CWE-1104, OWASP A06:2021</sub>

Dockerfile
```diff
@@ -1,4 +1,52 @@
...
+WORKDIR /var/projects/${APP_USER}
+COPY requirements.txt .
+# Pre-install CPU-only PyTorch before requirements.txt so pyspacer
+# (which depends on torch) picks up the lighter wheel (~280 MB vs ~2 GB).
+# ECS tasks run on t3a instances with no GPU.
+# Versions pinned to match pyspacer==0.12.0 constraints (torch>=2.6,<2.7).
+RUN su ${APP_USER} -c "\
+    pip install --upgrade pip \
+ && pip install --no-cache-dir --no-compile \
+        torch==2.6.0 torchvision==0.21.0 --index-url https://download.pytorch.org/whl/cpu \
```
gridcell, line 28: **Severe** · `correctness` · confidence: high

**Builder pip install runs without a login shell and without PATH extension, so packages install under root rather than the app user**

The builder stage runs `su ${APP_USER} -c "pip install ..."` (non-login shell). The `ENV PATH="/home/webapp/.local/bin:${PATH}"` extension exists only in the runtime stage, not the builder stage. When the builder executes pip as the non-login webapp user, `~/.local/bin` is not on PATH. On Debian slim images pip is not installed system-wide, so the `su -c` shell inherits root's ENV, meaning `pip install --upgrade pip` and all subsequent installs run under `/root/.local` (or system site-packages) rather than `/home/webapp/.local`. The `COPY --from=builder /home/${APP_USER}/.local` in the runtime stage will then copy nothing useful, and the container will fail at startup. The dev stage avoids this because it correctly uses `su -l` (login shell).

> **Suggestion:** Either add `ENV PATH="/home/webapp/.local/bin:${PATH}"` at the top of the builder stage (before the `su` call), or switch the builder install command to `su -l ${APP_USER} -c "pip install ..."` — the same pattern already used correctly in the dev stage.

> **Why it matters:** If packages land under root's home directory instead of the app user's home, the `COPY --from=builder /home/webapp/.local` instruction copies nothing useful. The runtime image will be missing all Python dependencies and will crash at startup across all five ECS task definitions. The builder-stage smoke-test import check would also run with the wrong paths and could pass vacuously, hiding the failure until deploy time.

<sub>Found by: scaffold, sentinel · CR-002 · CWE-427</sub>

Dockerfile
```diff
@@ -1,4 +1,52 @@
...
+ && pip install --no-cache-dir --no-compile -r requirements.txt"
+
+# Strip unnecessary files from installed packages to shrink the layer
+RUN find /home/${APP_USER}/.local -type d -name '__pycache__' -exec rm -rf {} + \
+ && find /home/${APP_USER}/.local -type d -name 'tests' \
+        ! -path '*/django/contrib/admin/tests' \
+        ! -path '*/pandas/tests*' \
+        ! -path '*/pandas/_testing*' \
```
gridcell: **Warning** · `correctness` · confidence: high

**Smoke-test import check in builder uses the same broken non-login shell and may pass vacuously**

The `RUN su ${APP_USER} -c "python -c 'import django; ...'"` verification line uses the same non-login `su` form as the install step. If CR-002 is real and packages installed under root, this check runs with the system Python and resolves the wrong site-packages, potentially passing with system-installed packages or failing with an obscure ImportError rather than a clear diagnostic about packages not being at the expected path.

> **Suggestion:** Change the smoke-test line to `su -l ${APP_USER} -c ...` consistent with the fix for CR-002, and additionally assert that packages come from the expected path: `import sys; assert any('/home/webapp' in p for p in sys.path), sys.path` before the other imports.

> **Why it matters:** A smoke-test that can pass incorrectly provides false confidence and hides the deployment failure until ECS task startup.

<sub>Found by: scaffold · CR-003</sub>

Dockerfile
```diff
@@ -1,4 +1,52 @@
...
+        ! -path '*/psycopg/tests*' \
+        ! -path '*/pyarrow/tests*' \
+        -exec rm -rf {} + \
+ && find /home/${APP_USER}/.local -name '*.pyc' -delete \
+ && find /home/${APP_USER}/.local -name '*.pyo' -delete
+
+# Smoke-test: verify top-level dependencies import successfully after cleanup
```
gridcell, line 43: **Warning** · `maintainability` · confidence: high

**LANG, LANGUAGE, LC_ALL, and PYTHONPATH env vars dropped from runtime stage without explanation**

The original single-stage image set `LANG=C.UTF-8`, `LANGUAGE=C.UTF-8`, `LC_ALL=C.UTF-8`, and `PYTHONPATH="/var/projects/webapp"`. The new runtime stage omits all four. `PYTHONPATH` is load-bearing: it ensures `manage.py` can resolve `app.settings` and the project root without an explicit `sys.path` append. Removing it may work when the working directory is always `/var/projects/webapp`, but any `docker exec` or ECS command override that changes directory will break imports.

> **Suggestion:** Restore the dropped environment variables in the runtime stage:
> ```
> ENV LANG=C.UTF-8 \
>     LANGUAGE=C.UTF-8 \
>     LC_ALL=C.UTF-8 \
>     PYTHONPATH="/var/projects/webapp"
> ```
> If any were intentionally removed, add an inline comment explaining the rationale.

> **Why it matters:** Missing `PYTHONPATH` can cause `ModuleNotFoundError` for `app.settings` in ECS scheduled tasks and queue workers that run management commands with a different working directory, potentially causing silent task failures.

<sub>Found by: scaffold · CR-004</sub>

Dockerfile
gridcell, line 4: **Warning** · `maintainability` · confidence: high

**Python 3.12-to-3.13 version bump is undocumented in the PR description and base image is not digest-pinned**

Both `FROM python:3.13-slim-bookworm AS builder` and `FROM python:3.13-slim-bookworm AS runtime` use Python 3.13. The PR description states 'Base image: python:3.12-slim-bookworm preserved', which is incorrect. This discrepancy means the version upgrade is effectively undocumented as an intentional change. When diagnosing a future regression, the team may anchor on 3.12 as the last known-good baseline. Additionally, neither base image is pinned to a digest or a patch version, so `docker build --pull` will silently pull a newer patch at any future build, making builds non-reproducible.

> **Suggestion:** 1. Update the PR description to explicitly call out the 3.12 → 3.13 upgrade as intentional. 2. Pin the base image to a full digest for reproducible builds (e.g. `python:3.13.3-slim-bookworm@sha256:<digest>`).

> **Why it matters:** Without documenting the version change, the team loses the ability to bisect whether a future production issue predates or postdates this change.

<sub>Found by: scaffold · CR-005</sub>

Dockerfile
```diff
@@ -1,4 +1,52 @@
...
+    pip install --upgrade pip \
+ && pip install --no-cache-dir --no-compile \
+        torch==2.6.0 torchvision==0.21.0 --index-url https://download.pytorch.org/whl/cpu \
+ && pip install --no-cache-dir --no-compile -r requirements.txt"
+
+# Strip unnecessary files from installed packages to shrink the layer
+RUN find /home/${APP_USER}/.local -type d -name '__pycache__' -exec rm -rf {} + \
+ && find /home/${APP_USER}/.local -type d -name 'tests' \
+        ! -path '*/django/contrib/admin/tests' \
+        ! -path '*/pandas/tests*' \
```
gridcell: **Warning** · `maintainability` · confidence: medium

**Test-directory pruning allowlist is fragile and will silently grow stale**

The `find ... -type d -name 'tests'` RUN step hard-codes exclusions for django, pandas, psycopg, and pyarrow. Any new package whose `tests/` directory is required at load time will be silently deleted, causing opaque import errors that will be difficult to attribute. The exclusion list does not cover every variant (the pattern `*/psycopg/tests*` matches psycopg3 but not a hypothetical `psycopg_binary` wheel path).

> **Suggestion:** Consider whether this step is worth the fragility: `pip install --no-compile` (already present) already reduces size significantly, and the marginal saving from additionally removing test directories is small. If pruning must remain, document each allowlisted package with an inline comment explaining why its `tests/` directory must be preserved.

> **Why it matters:** Silent deletion of package internals required at runtime will produce `ModuleNotFoundError` at deploy time across all five ECS tasks.

<sub>Found by: scaffold · CR-006</sub>

Dockerfile
gridcell, line 28: **Warning** · `dependency` · confidence: high

**PyTorch wheels installed from external index without hash pinning**

`torch==2.6.0` and `torchvision==0.21.0` are installed from `https://download.pytorch.org/whl/cpu` using `--index-url` without `--require-hashes`. Versions are pinned but binary content is not verified. A compromise of the pytorch.org CDN, a TLS downgrade in the build environment, or a mirror swap could cause tampered wheels to be installed with no detection. The rest of requirements.txt is similarly unhashed, but adding a second externally-controlled index URL widens the attack surface.

> **Suggestion:** Adopt hash-pinned requirements for PyTorch (and ideally the full requirements.txt). Generate pinned hashes via `pip-compile --generate-hashes` (pip-tools) and install with `pip install --require-hashes -r requirements.lock`. At minimum, record the SHA256 of the two wheels and pin them via `--hash=sha256:...` on the torch/torchvision lines.

> **Why it matters:** Without content pinning, builds are not reproducible and the image is vulnerable to silent supply-chain substitution. PyTorch wheels execute arbitrary native code on install (setup hooks) and at runtime, so a compromised wheel is effectively RCE in every ECS task.

<sub>Found by: sentinel · CR-007 · CWE-494, CWE-829, OWASP A08:2021</sub>

Dockerfile
```diff
@@ -10,41 +58,35 @@ ENV PATH="/home/webapp/.local/bin:${PATH}"
...
 
-# Install OS dependencies
+# Install runtime-only OS deps (no build-essential, libpq-dev, python3-dev)
 RUN apt-get update && apt-get install -y --no-install-recommends \
-    wget gnupg lsb-release ca-certificates \
- && echo "deb http://apt.postgresql.org/pub/repos/apt $(lsb_release -cs)-pgdg main" > /etc/apt/sources.list.d/pgdg.list \
- && wget --quiet -O - https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add - \
+    wget gnupg ca-certificates \
+ && wget --quiet -O /usr/share/keyrings/pgdg.asc https://www.postgresql.org/media/keys/ACCC4CF8.asc \
+ && gpg --dry-run --import --import-options show-only /usr/share/keyrings/pgdg.asc | grep -q 'B97B0AFCAA1A47F044F244A07FCC7D46ACCC4CF8' \
```
gridcell: **Question** · `security` · confidence: medium

**Confirm GPG fingerprint grep format matches gpg --dry-run output**

The fingerprint check uses `gpg --dry-run --import --import-options show-only /usr/share/keyrings/pgdg.asc | grep -q 'B97B0AFCAA1A47F044F244A07FCC7D46ACCC4CF8'`. However, gpg prints fingerprints with spaces (e.g. `B97B 0AFC AA1A 47F0 44F2 44A0 7FCC 7D46 ACCC 4CF8`), not as a contiguous 40-character string. The grep pattern may never match the default show-only output, meaning the build either fails every time (safe but misleading) or the fingerprint check is effectively bypassed if an older gpg emits a different format.

> **Suggestion:** Use `--with-colons` for machine-parseable output and match on the structured field:
> ```
> gpg --dry-run --import --import-options show-only --with-colons /usr/share/keyrings/pgdg.asc \
>   | awk -F: '/^fpr:/ {print $10}' \
>   | grep -qx 'B97B0AFCAA1A47F044F244A07FCC7D46ACCC4CF8'
> ```
> Also run the exact command in a throwaway container to confirm the current grep matches today.

<sub>Found by: sentinel · CR-009 · CWE-347, CWE-494</sub>

Dockerfile
gridcell, line 43: **Info** · `maintainability` · confidence: high

**No HEALTHCHECK instruction in runtime or dev stages**

Neither the runtime nor dev stage defines a Docker `HEALTHCHECK`. While the ALB health check covers the API service, tasks running as ECS scheduled/batch commands (ScheduledBackupTask, SummaryCacheTask, queue workers) have no container-level liveness check.

> **Suggestion:** Add a `HEALTHCHECK` to the runtime stage appropriate for the API service, e.g. `HEALTHCHECK --interval=30s --timeout=10s --start-period=90s --retries=3 CMD curl -fs http://localhost:8081/health/ || exit 1`.

<sub>Found by: scaffold · CR-012</sub>

gridcell (changes requested):
## AI Code Review

**Assessment:** The multi-stage conversion is architecturally sound, but the PR is **blocked by a Django 4.2 vs Python 3.13 incompatibility**: Django 4.2 LTS officially supports only Python 3.10–3.12, and Django 5.1 is the first release to support 3.13. A second blocker (CR-002) is the builder stage running pip without a login shell or PATH extension, which likely causes packages to install under root instead of the app user. The remaining findings include modern APT keyrings, build toolchain excluded from runtime, dev deps separated from production: modern APT keyrings, build toolchain excluded from runtime, dev deps separated from production, and secrets excluded from the build context. One blocking issue must be resolved before merge: the builder stage runs pip install without a login shell and without the PATH extension that puts `~/.local/bin` first, meaning all Python packages likely install under root rather than the app user, causing the `COPY --from=builder` step to copy nothing and the runtime container to fail at startup. Four warnings cover a fragile test-directory pruning allowlist, missing environment variables dropped from the original Dockerfile, an undocumented Python 3.12-to-3.13 version bump, and the `rich` package missing from `requirements-dev.txt`. Two questions address supply-chain hardening and CDK task scope.

| 2 | 6 | 2 | 3 | 5 |

**Files reviewed:** 6 / 6

### Cross-Cutting Concerns
- The builder stage PATH issue (CR-001) has a cascading effect on four other findings: the smoke-test check (CR-002) shares the same broken context; the test-directory pruning (CR-005) operates on packages that may have installed to the wrong location; and CR-003 (missing `PYTHONPATH`) compounds the risk of import failures. Fixing CR-001 first will clarify which of the downstream symptoms are real vs. hypothetical.
- The Python 3.12-to-3.13 upgrade is undocumented in the PR description (CR-004) and surfaces in three separate findings across agents: Sentinel's base-image digest concern, Scaffold's maintainability concern, and Forge's scikit-learn wheel availability concern (CR-011). The upgrade should be explicitly called out as intentional and validated end-to-end in CI before merge.
- Polisher found no style or linting violations in `iac/stacks/api.py`, confirming the CDK changes are clean.

### Additional findings

<details>
<summary><strong>CR-007</strong> &middot; Warning &middot; <code>dependency</code> — rich package missing from requirements-dev.txt despite pytest --rich flag in pytest.ini</summary>

pytest.ini configures addopts with `--rich`, which requires the `rich` package at test time. The PR moves pytest, pytest-cov, and pytest-django into requirements-dev.txt, but `rich` is not added alongside them. If `rich` was previously pulled in transitively via another package in requirements.txt, running tests after a clean install of only requirements-dev.txt could fail with a plugin-not-found or import error.

> **Suggestion:** Verify whether `rich` is a direct test-time dependency or pulled in transitively. If direct, add `rich==<pinned>` to requirements-dev.txt alongside the three pytest packages.

> **Why it matters:** Developers setting up a fresh environment from requirements-dev.txt could get a broken test runner without an obvious error pointing to the missing package.

Applies to: `requirements-dev.txt`

<sub>Found by: sentry</sub>
</details>

<details>
<summary><strong>CR-009</strong> &middot; Question &middot; <code>correctness</code> — CDK target=runtime affects all 5 ECS tasks — confirm none need dev-stage packages</summary>

The updated comment in `iac/stacks/api.py` correctly documents that the single `DockerImageAsset` drives all five task definitions. With `target="runtime"`, none of the tasks receive dev stage packages (Sphinx, sphinx-rtd-theme, pre-commit, and the newly moved pytest family). That is almost certainly correct — documentation and test tooling have no place in ECS. Worth explicitly confirming that no scheduled task or queue worker relies on anything that would logically belong in dev requirements.

Applies to: `iac/stacks/api.py`, `requirements-dev.txt`

<sub>Found by: scaffold</sub>
</details>

<details>
<summary><strong>CR-010</strong> &middot; Info &middot; <code>security</code> — Python 3.13 bump changes the interpreter under all C-extensions — ensure CI validates compatibility</summary>

The base image jump from `python:3.12-slim-bookworm` to `python:3.13-slim-bookworm` changes the interpreter under every C-extension in the image (torch, psycopg, pyarrow, GDAL). The smoke-test import line is a good sanity check but does not exercise runtime behaviour. Confirm via the full pytest suite that no deprecations or C-extension ABI issues surface on 3.13.

> **Suggestion:** Before merge, ensure CI runs `make test` against the new dev stage image to confirm 3.13 compatibility end-to-end. Consider pinning the base image by digest for build reproducibility.

Applies to: `Dockerfile`

<sub>Found by: sentinel</sub>
</details>

<details>
<summary><strong>CR-011</strong> &middot; Info &middot; <code>dependency</code> — Confirm scikit-learn (transitive pyspacer dep) has a Python 3.13 wheel available</summary>

pyspacer 0.12.0 brings scikit-learn (~1.5.x) as a transitive dependency. scikit-learn added cp313 wheels in 1.5.x, so this should resolve, but the pip resolver's exact behaviour under Python 3.13 for pyspacer's pinned constraints is worth confirming via a local build of the builder stage before merge.

Applies to: `Dockerfile`, `requirements.txt`

<sub>Found by: forge</sub>
</details>

### Praise

<details>
<summary><strong>CR-013</strong> &middot; Praise &middot; <code>security</code> — APT keyring modernised: signed-by keyring with GPG fingerprint verification replaces deprecated apt-key add</summary>

Moving from `wget ... | apt-key add -` (deprecated, globally trusted for all repos) to a per-repo `signed-by=/usr/share/keyrings/pgdg.asc` with HTTPS fetch and fingerprint verification is the current Debian best practice. The addition of the fingerprint check before trusting the key is excellent defense-in-depth.

Applies to: `Dockerfile` (runtime stage)

<sub>Found by: sentinel, scaffold</sub>
</details>

<details>
<summary><strong>CR-014</strong> &middot; Praise &middot; <code>security</code> — .env and .secrets.env correctly excluded from the Docker build context</summary>

Adding `.env` and `.secrets.env` to `.dockerignore` meaningfully reduces the risk of accidentally baking real secrets into image layer history during local builds. Confirmed that neither file is ever `COPY`ed or `ADD`ed in the Dockerfile — they are only injected at runtime via docker-compose `env_file`. Combined with stripping `__pycache__`, `docs/`, and the Insomnia collection, this is a clean hygiene pass on the build context.

Applies to: `.dockerignore`

<sub>Found by: sentinel, sentry</sub>
</details>

<details>
<summary><strong>CR-015</strong> &middot; Praise &middot; <code>security</code> — Clean separation of dev/test deps from the runtime image</summary>

Moving pytest, pytest-cov, and pytest-django out of requirements.txt and into requirements-dev.txt, combined with the 3-stage build where only the `dev` target installs them, keeps the production runtime image free of test-framework attack surface. The CDK stack correctly pins `target="runtime"` so ECS never pulls the dev layer.

Applies to: `Dockerfile`, `requirements.txt`, `requirements-dev.txt`, `iac/stacks/api.py`

<sub>Found by: sentinel</sub>
</details>

<details>
<summary><strong>CR-016</strong> &middot; Praise &middot; <code>maintainability</code> — Build toolchain correctly excluded from the runtime stage</summary>

Keeping compiler toolchain and header files (`build-essential`, `libpq-dev`, `python3-dev`, `git`) out of the runtime image is a meaningful security and size win. The builder-to-runtime `COPY` of `/home/webapp/.local` transfers only the compiled wheels without carrying the toolchain forward.

Applies to: `Dockerfile`

<sub>Found by: scaffold</sub>
</details>

<details>
<summary><strong>CR-017</strong> &middot; Praise &middot; <code>security</code> — APP_UID ARG ensures numeric UID consistency between builder and runtime stages</summary>

Explicit `--uid ${APP_UID}` on both `useradd` calls ensures that `COPY --from=builder --chown=${APP_USER}:${APP_USER}` transfers files whose ownership numeric IDs match the runtime user. This is a subtle but important correctness detail in multi-stage builds where UID allocation could otherwise differ between stages.

Applies to: `Dockerfile` (builder & runtime stages)

<sub>Found by: scaffold</sub>
</details>

### Agent Coverage
| sentinel | security | 4 | 7 | 0 |
| scaffold | infrastructure | 4 | 10 | 1 |
| forge | backend | 2 | 1 | 0 |
| polisher | python-quality | 1 | 0 | 0 |
| sentry | general | 4 | 5 | 0 |

Dockerfile
```diff
@@ -10,41 +58,37 @@ ENV PATH="/home/webapp/.local/bin:${PATH}"
...
 
-# Install OS dependencies
+# Install runtime-only OS deps (no build-essential, libpq-dev, python3-dev)
 RUN apt-get update && apt-get install -y --no-install-recommends \
-    wget gnupg lsb-release ca-certificates \
- && echo "deb http://apt.postgresql.org/pub/repos/apt $(lsb_release -cs)-pgdg main" > /etc/apt/sources.list.d/pgdg.list \
- && wget --quiet -O - https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add - \
+    wget gnupg ca-certificates \
+ && wget --quiet -O /usr/share/keyrings/pgdg.asc https://www.postgresql.org/media/keys/ACCC4CF8.asc \
+ && gpg --dry-run --import --import-options show-only --with-colons /usr/share/keyrings/pgdg.asc \
```
michaelconnor00, line 65: I don't understand what this line is doing. Assume `--dry-run` is doing nothing.

Dockerfile
```diff
@@ -10,41 +58,37 @@ ENV PATH="/home/webapp/.local/bin:${PATH}"
...
+      | awk -F: '/^fpr:/ {print $10}' \
+      | grep -qx 'B97B0AFCAA1A47F044F244A07FCC7D46ACCC4CF8' \
+ && echo "deb [signed-by=/usr/share/keyrings/pgdg.asc] https://apt.postgresql.org/pub/repos/apt bookworm-pgdg main" > /etc/apt/sources.list.d/pgdg.list \
```
michaelconnor00, line 68: `bookworm` is a reference to the base image version. It should be variable so it can be changed easily. It is easy to miss in this line.

michaelconnor00 (changes requested):
This is great. Long over due. I added a couple small comments.

michaelconnor00:
Also, please click `Resolve Conversation` on all AI comments if you have evaluated them. 
Also, you need to update your branch.

ms280690:
@coderabbitai review this PR

ms280690:
build and test are successful locally

PR #669, opened 2026-04-14
[M1458] remove outer transaction from FishIngester

saanobhaai:
## Summary by CodeRabbit

* **Bug Fixes**
  * Added pre-import duplicate detection to surface family/genus conflicts before writing.
  * Dry-run now exits after successful prevalidation to avoid partial work.
  * Import now aborts immediately on first error and logs an aborted run.
  * Improved robustness for missing region data and missing parent relationships during import.

### DIFF
--- a/src/api/ingest/attributes_ingester.py
+++ b/src/api/ingest/attributes_ingester.py
@@ -252,6 +252,7 @@ class FishIngester(BaseAttributeIngester):
     ERROR = "ERROR"
     EXISTING_FAMILY = "EXISTING_FAMILY"
     NEW_FAMILY = "NEW_FAMILY"
+    DUPLICATE_FAMILY = "DUPLICATE_FAMILY"
 
     EXISTING_GENUS = "EXISTING_GENUS"
     NEW_GENUS = "NEW_GENUS"
@@ -322,8 +323,10 @@ class FishIngester(BaseAttributeIngester):
         )
 
     def _prevalidate_rows(self, rows):
-        """Check all lookup and region values without touching the DB. Returns list of error strings."""
+        """Validate all rows before any writes. Returns list of error strings."""
         errors = []
+
+        # Per-row field/lookup/region checks
         for i, row in enumerate(rows, start=2):
             try:
                 mapped = self._map_fields(
@@ -344,6 +347,39 @@ class FishIngester(BaseAttributeIngester):
                     name = region.strip().lower()
                     if name and name not in self.regions:
                         errors.append(f"Row {i} - Unknown region: '{region.strip()}'")
+
+        # Check for duplicate family names in the DB — these would cause silent skips
+        csv_family_names = {
+            (row.get("Family") or "").strip() for row in rows if (row.get("Family") or "").strip()
+        }
+        for name in sorted(csv_family_names):
+            if FishFamily.objects.filter(name__iexact=name).count() > 1:
+                errors.append(
+                    f"Family '{name}': multiple entries with this name exist in the database"
+                )
+
+        # Check for duplicate genus names (per family) in the DB
+        csv_genera_by_family = {}
+        for row in rows:
+            family = (row.get("Family") or "").strip()
+            genus = (row.get("Genus") or "").strip()
+            if family and genus:
+                csv_genera_by_family.setdefault(family, set()).add(genus)
+
+        for family_name, genus_names in sorted(csv_genera_by_family.items()):
+            try:
+                family = FishFamily.objects.get(name__iexact=family_name)
+            except FishFamily.DoesNotExist:
+                continue  # new family — no existing genera to check
+            except FishFamily.MultipleObjectsReturned:
+                continue  # already flagged as a duplicate family above
+            for genus_name in sorted(genus_names):
+                if FishGenus.objects.filter(name__iexact=genus_name, family=family).count() > 1:
+                    errors.append(
+                        f"Genus '{genus_name}' under family '{family_name}': "
+                        f"multiple entries with this name exist in the database"
+                    )
+
         return errors
 
     def ingest(self, dry_run=False, allow_multiword_species=False):
@@ -412,6 +448,8 @@ class FishIngester(BaseAttributeIngester):
         return fish_family
 
     def _ingest_fish_genus(self, row, fish_family):
+        if fish_family is None:
+            return None
         genus_row = self._map_fields(row, self.fish_genus_field_map, self.fish_genus_lookups)
         genus_name = genus_row["name"]
         try:
@@ -426,6 +464,8 @@ class FishIngester(BaseAttributeIngester):
         return genus
 
     def _ingest_fish_species(self, row, fish_genus):
+        if fish_genus is None:
+            return
         species_row = self._map_fields(
             row,
             self.fish_species_field_map,
@@ -437,7 +477,7 @@ class FishIngester(BaseAttributeIngester):
         try:
             species = FishSpecies.objects.get(name__iexact=species_name, genus=fish_genus)
             has_edits = False
-            region_names = species_row.pop("regions")
+            region_names = species_row.pop("regions", None)
             updates = []
             for k, v in species_row.items():
                 original_val = getattr(species, k)
@@ -462,7 +502,7 @@ class FishIngester(BaseAttributeIngester):
                 self.write_log(self.EXISTING_SPECIES, f"{genus_name}-{species_name}")
 
         except FishSpecies.DoesNotExist:
-            region_names = species_row.pop("regions")
+            region_names = species_row.pop("regions", None)
             species = FishSpecies.objects.create(genus=fish_genus, **species_row)
             self._update_regions(species, region_names, is_combined=False)
 
--- a/src/api/ingest/attributes_ingester.py
+++ b/src/api/ingest/attributes_ingester.py
@@ -367,29 +367,31 @@ class FishIngester(BaseAttributeIngester):
         if prevalidation_errors:
             return False, self.log
 
+        if dry_run:
+            self.write_log("DRY_RUN", "No changes committed.")
+            return True, self.log
+
         n = 2
         is_successful = True
-        with transaction.atomic():
-            for row in rows:
-                try:
+        for row in rows:
+            try:
+                with transaction.atomic():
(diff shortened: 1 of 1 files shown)


saanobhaai:
[update fish species](https://trello.com/c/pfxXBCqn/1458-update-fish-species)

src/api/ingest/attributes_ingester.py
```diff
@@ -410,8 +448,11 @@ def _ingest_fish_family(self, row):
         return fish_family
 
     def _ingest_fish_genus(self, row, fish_family):
+        if fish_family is None:
+            return None
```
gridcell: **Severe** · `correctness` · confidence: high

**Duplicate family/genus at runtime returns None and silently skips the row**

When `_ingest_fish_family` catches `MultipleObjectsReturned` (line 445-446), `fish_family` remains `None`. This null guard returns `None`, and the null guard in `_ingest_fish_species` (line 468) returns without writing. No exception is raised, so the `except Exception` branch in the `ingest` loop is never reached. The per-row `transaction.atomic()` commits cleanly (a no-op), `is_successful` stays `True`, and the loop continues.

The prevalidation catches duplicates that existed *before* the run started, but cannot prevent duplicates created by earlier rows in the same run or by concurrent processes.

> **Suggestion:** Raise an explicit exception inside the `MultipleObjectsReturned` handlers:
> ```python
> except FishFamily.MultipleObjectsReturned:
>     raise ValueError(f"Multiple families named '{family_name}' exist; cannot determine which to use.")
> ```
> This lets the existing `except Exception` block catch it, log the error, and break.

> **Why it matters:** Species rows for families with database duplicates are silently dropped. The caller and operator see a successful run but data is missing.

<sub>Found by: forge, sentinel · CR-001</sub>

src/api/ingest/attributes_ingester.py
```diff
@@ -367,29 +403,31 @@ def ingest(self, dry_run=False, allow_multiword_species=False):
...
 
         if not is_successful:
             self.log = [e for e in self.log if e.startswith("ERROR")]
-            self.write_log("ROLLED_BACK", "All changes rolled back due to errors above.")
-        elif dry_run:
-            self.write_log("DRY_RUN", "No changes committed.")
+            self.write_log(
+                "ABORTED",
+                "Processing stopped due to error above. Successfully processed rows were committed.",
+            )
```
gridcell, line 465: **Severe** · `correctness` · confidence: high

**Log filtering drops audit trail of committed rows on error**

Under the new per-row atomic model, all rows processed *before* the failing row have been permanently committed. Filtering the log to only `ERROR` entries discards `EXISTING_*`, `NEW_*`, `UPDATE_*` entries — the only record of what was actually written to the database.

> **Suggestion:** Keep the full log and append the ABORTED entry:
> ```python
> if not is_successful:
>     self.write_log(
>         "ABORTED",
>         "Processing stopped due to error above. Successfully processed rows were committed.",
>     )
> # Do NOT filter self.log
> ```

> **Why it matters:** Operators cannot determine which rows were committed after an aborted partial run, making recovery and auditing impossible.

<sub>Found by: forge · CR-002</sub>

src/api/ingest/attributes_ingester.py
```diff
@@ -367,29 +403,31 @@ def ingest(self, dry_run=False, allow_multiword_species=False):
...
-                    break
-                finally:
-                    n = n + 1
-
-            if dry_run or not is_successful:
-                transaction.set_rollback(True)
+            except Exception as err:
+                self.write_log(self.ERROR, f"Row {n} - {str(err)}")
+                is_successful = False
+                break
```
gridcell, line 457: **Warning** · `error_handling` · confidence: high

**break-on-first-error is inconsistent with partial-commit semantics**

With per-row atomic blocks, a failure in row N rolls back only that row. Rows N+1 onward could succeed independently. The `break` discards them all. The ABORTED message implies partial commits are expected, but only rows *before* the first error are committed.

If partial commits are the intent, consider `continue` to accumulate all errors. If stop-on-first-error is deliberate, make that explicit in the log message (e.g., "Processing stopped at first error").

> **Why it matters:** Operators importing large CSVs with isolated bad rows must re-run the import repeatedly, re-processing already-committed rows each time.

<sub>Found by: forge, sentinel · CR-004</sub>

src/api/ingest/attributes_ingester.py
```diff
@@ -344,6 +347,39 @@ def _prevalidate_rows(self, rows):
                     name = region.strip().lower()
                     if name and name not in self.regions:
                         errors.append(f"Row {i} - Unknown region: '{region.strip()}'")
+
+        # Check for duplicate family names in the DB — these would cause silent skips
+        csv_family_names = {
+            (row.get("Family") or "").strip() for row in rows if (row.get("Family") or "").strip()
+        }
```
gridcell: **Warning** · `correctness` · confidence: high

**csv_family_names set does not normalize case, but DB queries use iexact**

The set comprehension uses `.strip()` without `.lower()`, so `'Acanthuridae'` and `'acanthuridae'` appear as two distinct entries. Each hits the DB separately with `filter(name__iexact=...)`. Set-level deduplication doesn't work for case variants.

> **Suggestion:** Normalize to lowercase:
> ```python
> csv_family_names = {
>     (row.get("Family") or "").strip().lower()
>     for row in rows
>     if (row.get("Family") or "").strip()
> }
> ```

> **Why it matters:** Case variants of the same family name bypass deduplication and produce redundant DB queries.

<sub>Found by: forge · CR-005</sub>

src/api/ingest/attributes_ingester.py
```diff
@@ -344,6 +347,39 @@ def _prevalidate_rows(self, rows):
...
+        for name in sorted(csv_family_names):
+            if FishFamily.objects.filter(name__iexact=name).count() > 1:
+                errors.append(
+                    f"Family '{name}': multiple entries with this name exist in the database"
+                )
```
gridcell: **Warning** · `performance` · confidence: high

**Prevalidation issues N+1 DB queries per unique family and genus**

The duplicate checks execute one `COUNT` query per family name, then one `GET` per family for the genus check, then one `COUNT` per genus. For a CSV with F families and G genera this is O(2F + G) sequential DB round-trips.

> **Suggestion:** Batch into bulk queries:
> ```python
> from django.db.models import Count
> dupes = FishFamily.objects.filter(
>     name__lower__in=[n.lower() for n in csv_family_names]
> ).values('name__lower').annotate(cnt=Count('id')).filter(cnt__gt=1)
> ```

> **Why it matters:** Sequential per-name queries are slow on large CSVs and increase lock contention time during ingestion.

<sub>Found by: forge · CR-006</sub>

gridcell (changes requested):
## AI Code Review

**Assessment:** The PR improves prevalidation (duplicate family/genus detection), makes dry-run a true no-op, and adds defensive null guards. However, the core transaction change introduces a data integrity concern: duplicate families/genera at runtime return None and silently skip the row without raising an exception, so the caller sees a successful run even though species data was not written. The log-filtering step also removes the audit trail of successfully committed rows, which is the opposite of what partial-commit semantics require.

| 3 | 3 | 1 | 1 | 3 |

**Files reviewed:** 1 / 1

### Cross-Cutting Concerns
- The shift from all-or-nothing to per-row transactions changes the error model fundamentally. Several downstream behaviors (log filtering, break-on-error, help text) still assume the old rollback semantics.
- Null returns from `_ingest_fish_family` / `_ingest_fish_genus` propagate silently through the call chain without raising exceptions, creating a gap between what the log reports and what was actually written.

---

<details>
<summary><strong>CR-003</strong> &middot; Severe &middot; <code>correctness</code> — loadfish --dry-run help text contradicts actual behavior</summary>

The management command's `--dry-run` help string reads: *"Runs load fish in a database transaction, then does a rollback."* After this PR, `dry_run` returns early before any DB writes — there is no transaction and no rollback.

> **Suggestion:** Update `src/tools/management/commands/loadfish.py` line 17:
> ```python
> help="Validate the CSV without writing any rows to the database.",
> ```

> **Why it matters:** Stale help text misleads operators about what dry-run validates. The old dry-run ran writes inside a transaction (catching constraint errors), while the new one only runs prevalidation checks.

Applies to: `src/tools/management/commands/loadfish.py`

<sub>Found by: forge</sub>
</details>

<details>
<summary><strong>CR-007</strong> &middot; Question &middot; <code>correctness</code> — DUPLICATE_FAMILY/GENUS log entries are dropped by error-path filter</summary>

The error-path log filter `[e for e in self.log if e.startswith('ERROR')]` only keeps entries prefixed with 'ERROR'. DUPLICATE_FAMILY and DUPLICATE_GENUS log entries are silently discarded. If a duplicate is encountered and then a later row errors, the operator loses the duplicate information.

> **Why it matters:** Related to CR-002 — the log filter is too aggressive for the new partial-commit model.

Applies to: `src/api/ingest/attributes_ingester.py`

<sub>Found by: forge</sub>
</details>

<details>
<summary><strong>CR-008</strong> &middot; Info &middot; <code>test_gap</code> — No tests cover the new prevalidation or partial-commit behavior</summary>

The existing test_ingest.py covers survey-data ingestion but not FishIngester. Key scenarios to test:
1. Prevalidation detects duplicate families in DB
2. `dry_run` returns True with no DB writes
3. A bad row mid-import leaves earlier rows committed
4. Duplicate family at runtime behavior

> **Why it matters:** The interacting subsystems (prevalidation, per-row atomics, null guards) are complex enough to warrant regression tests.

<sub>Found by: forge, polisher</sub>
</details>

### Praise

- **Prevalidation duplicate detection** — The new duplicate detection in `_prevalidate_rows` checks for ambiguous database state before any writes, preventing the ingester from silently picking the wrong parent record. *(sentinel)*
- **Dry-run early return** — Moving the `dry_run` check before the processing loop guarantees zero DB mutations, strictly better than the old `set_rollback(True)` approach. *(sentinel, forge)*
- **Defensive `.pop('regions', None)`** — Providing a default prevents `KeyError` when the 'regions' key is absent from the mapped row. *(forge)*

### Agent Coverage
| sentinel | security | 1 | 4 | 0 |
| forge | backend | 1 | 10 | 3 |
| polisher | python-quality | 1 | 6 | 0 |

gridcell (approved):

PR #671, opened 2026-04-23
[M1931] fixed UnicodeDecodeError in summary_report.py

saanobhaai:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CSV content handling to correctly process gzip-compressed streaming responses, ensuring proper decompression before parsing.

* **Tests**
  * Added comprehensive unit tests for CSV data retrieval, covering cache hits, plain responses, and gzip-compressed response scenarios.


saanobhaai:
[UnicodeDecodeError in create_sample_unit_method_summary_report](https://trello.com/c/PCHtvqie/1931-unicodedecodeerror-in-createsampleunitmethodsummaryreport)

src/api/reports/summary_report.py
```diff
@@ -261,7 +262,10 @@ def get_viewset_csv_content(view_cls, project_pk, request):
         print(resp.content)
```
gridcell: **warning** · `readability` · confidence: high

**Pre-existing debug `print(resp.content)` in error path should use the logger**

Line 262 calls `print(resp.content)` in the error branch. Pre-existing, but now more visible given the surrounding changes. Raw print bypasses the app's logging infrastructure and may leak large response bodies to stdout.

> **Suggestion:** Replace with `logger.error('Failed to get CSV content for project %s: %s', project_pk, resp.content)` using the module-level logger, consistent with how errors are logged elsewhere.

> **Why it matters:** Debug prints are invisible in production log aggregators and can leak large response bodies.

<sub>Found by: forge · CR-001</sub>

src/api/reports/summary_report.py
```diff
@@ -261,7 +262,10 @@ def get_viewset_csv_content(view_cls, project_pk, request):
...
         raise ValueError(f"Failed to get content for project {project_pk}")
 
-    content = list(csv.reader([str(row, "UTF-8").strip() for row in resp.streaming_content]))
+    raw_bytes = b"".join(resp.streaming_content)
+    if resp.get("Content-Encoding") == "gzip":
+        raw_bytes = gzip.decompress(raw_bytes)
+    content = list(csv.reader(raw_bytes.decode("UTF-8").splitlines()))
```
gridcell: **warning** · `style` · confidence: high

**Encoding string should be lowercase `utf-8` (matches Ruff UP009)**

Line uses `raw_bytes.decode("UTF-8")`. Python's canonical form is lowercase `utf-8`. The old code also used `UTF-8`, so this is a carry-over, but worth fixing while the line is being touched.

> **Suggestion:** `raw_bytes.decode("utf-8")`

> **Why it matters:** Inconsistent casing diverges from the Python-standard lowercase convention.

<sub>Found by: polisher · CR-002</sub>

src/api/tests/test_summary_report.py
```diff
@@ -11,3 +17,71 @@ def test_get_project_metadata_num_sample_units(project1, benthic_lit1):
 
     idx = header.index("Number of Sample Units")
     assert data[idx] == 1
+
+
+_CSV_ROWS = [
```
gridcell, line 25: **warning** · `test_gap` · confidence: high

**`_CSV_ROWS` lacks `site_id` — all tests hit the early-return branch only**

`get_viewset_csv_content` returns early when `'site_id' not in content[0]`. Because `_CSV_ROWS[0]` is `['protocol', 'name', 'value']`, every test exits before the covariate enrichment path (`_get_site_aca_covariate_columns`) and column filtering. The gzip test docstring also claims to test the 'cache-hit path in AggregatedViewMixin', but the cache is patched to return None — it actually tests the live streaming path.

> **Suggestion:** Either add a comment on `_CSV_ROWS` noting the intentional scope, or add a test with `site_id` to cover the enrichment branch. Also fix the misleading docstring on the gzip test.

> **Why it matters:** Tests silently skip the covariate-enrichment code path, leaving a substantial part of the function unexercised.

<sub>Found by: gauntlet · CR-004</sub>

src/api/tests/test_summary_report.py
```diff
@@ -11,3 +17,71 @@ def test_get_project_metadata_num_sample_units(project1, benthic_lit1):
...
+            self.kwargs = {}
+            self.request = None
+
+        def csv(self, request, *args, **kwargs):
+            return response
+
+    return MockView
+
+
+def test_get_viewset_csv_content_plain_response():
```
gridcell, line 56: **warning** · `test_gap` · confidence: high

**No test covers the `status_code != 200` error branch**

The function raises `ValueError` on non-200 status. `_make_view_cls` always returns a default 200 response, so the error path is completely uncovered.

> **Suggestion:** Add a test that constructs a StreamingHttpResponse with status=500 and asserts that `ValueError` is raised.

> **Why it matters:** The error-handling branch and its accompanying `print` (see CR-001) are exercised in neither the old nor new tests.

<sub>Found by: gauntlet · CR-005</sub>

src/api/tests/test_summary_report.py
```diff
@@ -11,3 +17,71 @@ def test_get_project_metadata_num_sample_units(project1, benthic_lit1):
...
+def test_get_viewset_csv_content_cached_file():
+    """When get_cached_textfile returns data, it is used directly without calling the viewset."""
+    csv_lines = [",".join(row) for row in _CSV_ROWS]
+
+    class MockView:
+        def __init__(self, **kwargs):
+            pass
+
+        def csv(self, request, *args, **kwargs):
+            raise AssertionError("viewset csv should not be called when cache is populated")
```
gridcell: **warning** · `maintainability` · confidence: high

**Duplicate MockView definitions with inconsistent shapes**

Two MockView definitions exist: the one inside `_make_view_cls` sets attrs from kwargs; the inline one in the cached test simply raises AssertionError from `csv()`. Both agents independently flagged this.

> **Suggestion:** Extend `_make_view_cls` to accept an optional `csv_side_effect` parameter (e.g. an AssertionError) so the cached test can reuse the shared helper.

> **Why it matters:** Duplicate helpers diverge over time, making test maintenance harder and obscuring intent.

<sub>Found by: polisher, gauntlet · CR-006</sub>

src/api/tests/test_summary_report.py
```diff
@@ -11,3 +17,71 @@ def test_get_project_metadata_num_sample_units(project1, benthic_lit1):
...
+]
+
+
+def _rows_to_bytes(rows):
+    buf = io.StringIO()
+    csv_module.writer(buf).writerows(rows)
+    return buf.getvalue().encode("utf-8")
+
+
+def _make_view_cls(response):
```
gridcell: **warning** · `style` · confidence: medium

**Test helpers lack type annotations**

`_rows_to_bytes` and `_make_view_cls` have no annotations. Python 3.12 allows modern syntax.

> **Suggestion:** `def _rows_to_bytes(rows: list[list[str]]) -> bytes:` and `def _make_view_cls(response: StreamingHttpResponse) -> type:`

> **Why it matters:** Unannotated helpers reduce readability and forfeit static analysis on the test utilities.

<sub>Found by: polisher · CR-007</sub>

src/api/reports/summary_report.py
```diff
@@ -261,7 +262,10 @@ def get_viewset_csv_content(view_cls, project_pk, request):
         print(resp.content)
         raise ValueError(f"Failed to get content for project {project_pk}")
 
-    content = list(csv.reader([str(row, "UTF-8").strip() for row in resp.streaming_content]))
+    raw_bytes = b"".join(resp.streaming_content)
+    if resp.get("Content-Encoding") == "gzip":
```
gridcell, line 273: **question** · `correctness` · confidence: medium

**Is `resp.get('Content-Encoding')` guaranteed to be set on every gzip response?**

The gzip branch in `cached.streaming_response()` sets `Content-Encoding: gzip` only when `content_encoding` is non-None. The caller in `AggregatedViewMixin.csv()` always passes `content_encoding='gzip'`, so the header is present on the cache-hit path; the live path returns a plain StreamingHttpResponse with no gzip. The edge case is if a future caller forgets to pass `content_encoding='gzip'` but S3 still returns gzip'd bytes — the raw gzip payload would reach the decode step silently.

> **Suggestion:** Worth considering a magic-byte fallback: `if raw_bytes[:2] == b'\x1f\x8b': raw_bytes = gzip.decompress(raw_bytes)`

<sub>Found by: forge · CR-008</sub>

src/api/tests/test_summary_report.py
```diff
@@ -1,6 +1,12 @@
+import csv as csv_module
```
gridcell: **question** · `readability` · confidence: medium

**Is `import csv as csv_module` necessary?**

Aliasing a stdlib module as `csv_module` is unconventional. There is no obvious `csv` name collision in the test module. Consider reverting to plain `import csv`.

<sub>Found by: polisher · CR-009</sub>

src/api/tests/test_summary_report.py
gridcell, line 56: **info** · `maintainability` · confidence: high

**Plain and gzip tests are duplicated — consider parametrizing**

`test_get_viewset_csv_content_plain_response` and `test_get_viewset_csv_content_gzip_response` share identical structure. Parametrizing on `(compress, set_encoding)` tuples would also make it trivial to add the non-ASCII regression case from CR-003.

<sub>Found by: gauntlet · CR-010</sub>

gridcell (commented):
## AI Code Review

**Assessment:** Correct, focused bug fix: joining all streaming chunks before decoding eliminates the chunk-boundary UnicodeDecodeError. No security or correctness blockers. Opportunities: replace a pre-existing debug print with the module logger, add a non-ASCII regression test that actually exercises the original failure mode, and cover the error (non-200) branch.

| 0 | 6 | 3 | 1 | 2 |

**Files reviewed:** 2 / 2

### Cross-Cutting Concerns
- The three test-gap warnings (CR-003, CR-004, CR-005) together mean that a broad swathe of `get_viewset_csv_content` — the error path, the covariate-enrichment path, and the exact multi-byte Unicode split that motivated the PR — remain unexercised. Addressing them as a group (ideally via a parametrized test, CR-010) would significantly improve confidence in the fix.
- The `print(resp.content)` call on the error path (CR-001) and the missing error-branch test (CR-005) are related: fixing both together would close the logging gap and provide coverage for the corrected log statement simultaneously.

### Additional Findings

<details>
<summary><strong>CR-003</strong> &middot; warning &middot; <code>test_gap</code> — No non-ASCII regression test — the original UnicodeDecodeError is not guarded against</summary>

The PR fixes a UnicodeDecodeError caused by multi-byte UTF-8 sequences split across chunk boundaries. None of the new tests include non-ASCII characters, so a regression would still pass the test suite. A Unicode field value (e.g. 'Récif Côtier') on the gzip path would act as a true regression guard.

> **Suggestion:** Add a parametrize case or extra test that includes non-ASCII text in `_CSV_ROWS` and runs through the gzip path.

> **Why it matters:** Without this, the exact failure mode named in the PR title has no automated coverage.

Applies to: `src/api/tests/test_summary_report.py`

<sub>Found by: gauntlet</sub>
</details>

<details>
<summary><strong>CR-011</strong> &middot; info &middot; <code>correctness</code> — `resp.get('Content-Encoding')` is the correct Django idiom</summary>

`HttpResponseBase.get(header, alternate=None)` is the documented case-insensitive header accessor. Using `resp['Content-Encoding']` would raise `KeyError` when the header is absent. `.get()` is the right choice here.

Applies to: `src/api/reports/summary_report.py`

<sub>Found by: forge</sub>
</details>

<details>
<summary><strong>CR-012</strong> &middot; praise &middot; <code>correctness</code> — Correct fix for chunk-boundary UnicodeDecodeError</summary>

Joining all chunks into a single bytes object before decoding is exactly the right approach. The old code decoded each chunk independently, which fails when a multi-byte UTF-8 sequence is split across chunk boundaries. `str.splitlines()` is also more robust than per-chunk row-splitting.

Applies to: `src/api/reports/summary_report.py`

<sub>Found by: forge</sub>
</details>

<details>
<summary><strong>CR-013</strong> &middot; praise &middot; <code>correctness</code> — AssertionError-in-mock pattern for cache short-circuit is clean</summary>

Using `AssertionError` inside the mock's `csv()` method fails loudly at the point of an unwanted call rather than via a post-hoc call-count assertion. This is a cleaner and more informative test design.

Applies to: `src/api/tests/test_summary_report.py`

<sub>Found by: gauntlet</sub>
</details>

### Agent Coverage
| sentinel | security | 2 | 0 | 0 |
| forge | backend | 1 | 4 | 0 |
| polisher | code_quality | 2 | 4 | 0 |
| gauntlet | testing | 1 | 5 | 0 |

src/api/tests/test_summary_report.py
saanobhaai: Not worth doing. The two tests are only ~8 lines each and already read clearly as independent cases. A parametrized form with a lambda b: b identity transform and conditional
   header-setting would be harder to scan than two explicit functions. The gzip test also has a meaningful docstring that would be lost or awkwardly attached to a parameter  
  tuple. Per the project's own guidance, three similar lines beats a premature abstraction — and the non-ASCII regression case is speculative

gridcell (approved):

PR #675, opened 2026-04-29
[1938] handle intermittent ACA API error

saanobhaai:
## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced resilience of covariate data retrieval with automatic retry mechanism for failed API requests, reducing failure rate from transient errors.
  * Improved error reporting and logging infrastructure to better capture and communicate data processing issues for troubleshooting.

### DIFF
--- a/src/api/covariates/coral_atlas.py
+++ b/src/api/covariates/coral_atlas.py
@@ -1,4 +1,5 @@
 import datetime
+import time
 from concurrent.futures import ThreadPoolExecutor
 from typing import List, Tuple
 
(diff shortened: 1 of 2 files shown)


saanobhaai:
[CovariateRequestError](https://trello.com/c/ksyb3odl/1938-covariaterequesterror)

src/api/covariates/coral_atlas.py
```diff
@@ -32,10 +33,19 @@ def _fetch(self, x: float, y: float, radius: float, request_datetime: datetime)
...
             "User-Agent": "Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:91.0) Gecko/20100101 Firefox/91.0"
         }
-        resp = requests.get(url, headers=headers)
-        status_code = resp.status_code
-        if status_code != 200:
-            raise CovariateRequestError(resp.text)
+        last_error = None
+        for attempt in range(3):
+            try:
+                resp = requests.get(url, headers=headers, timeout=(3.05, 15))
```
gridcell: Does 3.05s (to connect) and 15s (read timeout) make sense for this?

saanobhaai: Tightened a bit, but remember this is called by a worker, not in synchronous context. Claude's response:
The 3.05s connect value is deliberately chosen to sit just past the TCP SYN retransmit threshold (~3s) — a well-known requests best practice. The read timeout has been       
  updated to 10s: this call runs in a background SQS worker (not a web request), so it's not latency-critical, but 10s is still generous for a geospatial point query while     
  keeping worst-case per-site time (3 retries + backoff) to ~42s if ACA is degraded.

gridcell (approved):

PR #679, opened 2026-05-05
[M1941] image upload improvements

saanobhaai:
## Summary by CodeRabbit

* **Bug Fixes**
  * Enforced maximum image upload size (30MB); oversized uploads are rejected.

* **Improvements**
  * Faster image listing and loading via backend query optimizations.
  * More reliable thumbnail generation and checksum handling, including support for in-memory image buffers.

* **Tests**
  * Added a test to verify image upload size limits.

* **Chores**
  * Tuned server worker settings for improved stability and throughput.


saanobhaai:
@coderabbitai review

saanobhaai:
[intermittent API 502s](https://trello.com/c/AiLOItpK/1941-intermittent-api-502s)

gridcell (commented):
## AI Code Review

**Assessment:** The PR delivers solid improvements — a 30 MB upload guard, prefetch-aware serializer queries, in-memory buffer support for thumbnails/checksums, and gunicorn worker recycling — and introduces no security vulnerabilities or blocking correctness issues. One real bug warrants attention before merge: the _normalized_image_buf BytesIO reference is not released on the common steady-state path (unchanged checksum), leaking the decoded image in memory for the lifetime of the instance. The remaining findings are test-coverage gaps for all new behavioral paths, a question about File vs ContentFile size semantics, and minor style/documentation observations.

| 0 | 5 | 4 | 4 | 3 |

**Files reviewed:** 5 / 5

### Cross-Cutting Concerns
- The MAX_IMAGE_FILE_SIZE check is an application-layer guard only — it runs after Django has fully received and parsed the multipart body. A true streaming reject requires an upstream limit (nginx client_max_body_size, ALB, or CloudFront). Confirm this layer exists in the deployment configuration.
- The _normalized_image_buf instance-attribute pattern relies on the in-memory buffer being set, consumed, and deleted within a single signal invocation. The current code has a gap on the unchanged-checksum path (CR-001) and no test coverage for the lifecycle (CR-005). These two findings should be addressed together.
- None of the six new behavioral paths (file-size guard, buffer-based checksum, buffer-based thumbnail, save_normalized_imagefile, _counts_cache memoisation, max() for classification status) have corresponding tests. The test suite would not catch a regression in any of them.
- The _counts_cache instance-attribute lazy-init pattern is safe only while DRF instantiates a fresh serializer per request. This invariant is implicit and undocumented; initializing the dict in __init__ would make it explicit and eliminate the fragility.

### Highlights
- **Python max() over prefetched statuses correctly avoids extra DB queries** &mdash; `src/api/resources/classification/image.py` (`get_classification_status`)
- **Correct gunicorn max-requests recycling mitigates worker memory growth** &mdash; `src/docker-entry.sh` (`gunicorn`)
- **Upload-size limit correctly placed as a fast-fail before the heavy image pipeline** &mdash; `src/api/resources/classification/image.py` (`ImageViewSet.create`)

### Additional Findings

<details>
<summary><strong>CR-009</strong> &middot; Question &middot; <code>docs</code> &mdash; New image_buf parameter and _normalized_image_buf attribute lack documentation</summary>

Three utility functions — create_image_checksum, create_thumbnail, and save_normalized_imagefile — gained an optional image_buf parameter, and post_save_classification_image now stashes a private _normalized_image_buf attribute on the Image instance. Neither the parameter nor the attribute has a docstring or comment explaining: why it exists (avoiding S3 re-reads), the buffer semantics (caller must pass a readable, seekable BytesIO positioned at offset 0), the lifecycle (who owns the buffer, when it is consumed vs. reused), and that it is intentionally deleted after use. This context is non-obvious and important for any future maintainer tracing the signal handler.

Applies to: `src/api/utils/classification.py`, `src/api/signals/classification.py`

<sub>Found by: polisher</sub>

</details>

<details>
<summary><strong>CR-004</strong> &middot; Warning &middot; <code>test_gap</code> &mdash; image_buf branch in create_image_checksum and create_thumbnail is untested</summary>

Both functions gained an optional image_buf parameter. When image_buf is supplied they read from the in-memory buffer instead of reopening the file from storage. There are no unit tests for either the new buffer-path branch or for the correctness invariant that the buffer-based and file-based code paths produce identical output. The existing regression test exercises the full Image.objects.create signal chain but does not isolate or assert on these helper functions directly.

> **Suggestion:** Add unit tests for create_image_checksum and create_thumbnail that supply an in-memory BytesIO and verify the output matches what the file-path code path produces for the same image data.

> **Why it matters:** The buffer path is the hot path for all new image uploads. If buffer and file paths diverge (e.g. due to seek position or format conversion difference), thumbnails or checksums will silently be wrong for every new upload while existing records remain unaffected.

Location: `src/api/utils/classification.py:88-133`

<sub>Found by: gauntlet</sub>

</details>

<details>
<summary><strong>CR-013</strong> &middot; Info &middot; <code>maintainability</code> &mdash; Commented-out exec line is dead code (pre-existing)</summary>

Line 7 contains `# exec "$@"` (the standard Docker entrypoint passthrough idiom) commented out. This is pre-existing and unrelated to this PR, but it causes confusion for anyone reading the entrypoint to understand how the container starts. Worth removing in a follow-up cleanup.

Location: `src/docker-entry.sh:7-7`

<sub>Found by: scaffold</sub>

</details>

### Agent Coverage
| sentinel | security | 5 | 3 | 0 |
| forge | backend | 5 | 5 | 0 |
| polisher | python-quality | 4 | 4 | 0 |
| scaffold | infrastructure | 1 | 3 | 0 |
| gauntlet | testing | 3 | 5 | 0 |

src/api/signals/classification.py
```diff
@@ -63,20 +63,28 @@ def post_save_classification_image(sender, instance, created, **kwargs):
...
+            instance.image, image_buf=buf
+        )
         # Saving thumbnail (save=True), causes double save but it's necessary
         # to have thumbnail created and saved in the post_save so thumbnails
         # don't get orphaned if done in a pre_save signal.
+        # buf is intentionally still set here so the recursive post_save can
+        # use it for the checksum comparison, avoiding an S3 read.
         instance.thumbnail.save(thumb_file.name, thumb_file, save=True)
+        if buf is not None:
+            del instance._normalized_image_buf
```
gridcell: **Warning** · `correctness` · confidence: high

**_normalized_image_buf is never deleted when needs_new_thumbnail is False**

When instance.thumbnail is truthy and the checksum is unchanged, needs_new_thumbnail stays False and execution skips the entire `if needs_new_thumbnail:` block. The `del instance._normalized_image_buf` inside that block is therefore unreachable on the common steady-state path (existing thumbnail, unchanged image), leaving the BytesIO object pinned on the instance until Python's GC eventually collects it.

For large images this is a meaningful memory hold per-save. More critically, if the same model instance is reused by a follow-up signal call (e.g. the double-save that records the thumbnail back onto the Image row), _normalized_image_buf may still be present at a potentially stale seek position and could be passed to the next checksum/thumbnail helper call.

> **Suggestion:** Move the cleanup unconditionally after the `if needs_new_thumbnail:` block, guarded only by a `hasattr(instance, '_normalized_image_buf')` (or `buf is not None`) check, so the buffer is always released regardless of the thumbnail-rebuild decision.

> **Why it matters:** The buffer holds the full decoded, orientation-corrected image in memory. Failing to release it on the common steady-state path means every save of an already-processed Image leaks the buffer for the instance's lifetime — a significant memory overhead in an image-processing service.

<sub>Found by: forge · CR-001</sub>

src/api/resources/classification/image.py
```diff
@@ -65,37 +66,43 @@ class Meta:
...
+        # first one found is authoritative.
+        for point in obj.points.all():
+            for annotation in point.annotations.all():
+                if annotation.classifier_id is not None:
+                    return annotation.classifier.patch_size
+        return None
 
     def _summarize_counts(self, obj):
+        if not hasattr(self, "_counts_cache"):
+            self._counts_cache = {}
```
gridcell: **Warning** · `maintainability` · confidence: high

**hasattr/setattr lazy-init pattern for _counts_cache is fragile and non-idiomatic**

The `if not hasattr(self, '_counts_cache'): self._counts_cache = {}` pattern for lazy initialisation is less idiomatic Python and harder to reason about. Forge notes that the pattern is safe in practice because DRF instantiates a fresh ImageSerializer per request, so the cache will never bleed across requests or different Image objects in separate requests — but this invariant is invisible at the call site. Polisher flags the same pattern as harder to grep and surprising to maintainers.

The risk is low in the current DRF usage, but if the serializer were ever reused across objects (e.g. in a batch context), the cache would serve stale data silently.

> **Suggestion:** Initialize _counts_cache in the serializer's __init__ (e.g. `super().__init__(*args, **kwargs); self._counts_cache = {}`), or use `self._counts_cache = getattr(self, '_counts_cache', {})` at the top of _summarize_counts. Add a brief inline comment explaining the expected per-request lifetime so the safety invariant is explicit.

> **Why it matters:** A class-level or stale cache could silently serve wrong annotation counts to API consumers. The explicit __init__ pattern makes the lifecycle visible and prevents accidental promotion of the attribute to class-level scope.

<sub>Found by: forge, polisher · CR-002</sub>

src/api/resources/classification/image.py
```diff
@@ -183,6 +197,13 @@ def create(self, request, *args, **kwargs):
         collect_record_id = request.data.get("collect_record_id")
         trigger_classification = truthy(request.data.get("classify", True))
 
+        if image_file and image_file.size > settings.MAX_IMAGE_FILE_SIZE:
+            mb = settings.MAX_IMAGE_FILE_SIZE // (1024 * 1024)
+            return Response(
+                {"error": f"Image exceeds the {mb} MB size limit."},
+                status=status.HTTP_400_BAD_REQUEST,
+            )
```
gridcell, line 207: **Warning** · `test_gap` · confidence: high

**30 MB file-size guard has no test coverage**

ImageViewSet.create now checks image_file.size against settings.MAX_IMAGE_FILE_SIZE and returns HTTP 400 with an error message when exceeded. There is no test for the over-limit path, the boundary condition, or the normal (under-limit) pass-through. This is the only new user-facing HTTP error path introduced by the PR.

> **Suggestion:** Add a parametrized test using @override_settings(MAX_IMAGE_FILE_SIZE=10) and a tiny SimpleUploadedFile to cover at-limit, over-limit, and under-limit cases. This avoids constructing actual multi-MB fixtures.

> **Why it matters:** Without a test, a future refactor that moves or removes the size check would silently regress the validation, allowing arbitrarily large uploads to reach S3 and the PIL processing pipeline.

<sub>Found by: gauntlet · CR-003</sub>

src/api/signals/classification.py
gridcell: **Warning** · `test_gap` · confidence: high

**post_save_classification_image _normalized_image_buf lifecycle is untested**

post_save_classification_image reads instance._normalized_image_buf, passes it to the checksum/thumbnail helpers, and deletes the attribute — but only inside the `if needs_new_thumbnail` branch. No test confirms the attribute is deleted after a normal save, nor what happens on the unchanged-checksum fast path where needs_new_thumbnail is False (the buffer leak path identified in CR-001).

> **Suggestion:** Add a signal-level test that (a) stashes a BytesIO on the instance before save, (b) asserts the attribute is absent after save on both the needs_new_thumbnail=True and needs_new_thumbnail=False paths, and (c) confirms no exception is raised if the attribute is absent entirely.

> **Why it matters:** Without this test, the buffer-leak bug (CR-001) and any future regression in cleanup logic would go undetected by the test suite.

<sub>Found by: gauntlet · CR-005</sub>

src/api/utils/classification.py
```diff
@@ -244,7 +256,12 @@ def save_normalized_imagefile(instance: Image):
         img_content = BytesIO()
         image_file.save(img_content, format=image_format)
 
-        instance.image = ContentFile(img_content.getvalue(), name=instance.image.name)
+        # Wrap the BytesIO directly (no copy) and stash it for post_save so
+        # thumbnail creation and checksum can read from the buffer already in
+        # memory instead of fetching the file back from S3.
+        img_content.seek(0)
+        instance.image = File(img_content, name=instance.image.name)
+        instance._normalized_image_buf = img_content
```
gridcell, line 264: **Question** · `correctness` · confidence: medium

**File(BytesIO, name=...) may raise on .size access — ContentFile avoids this**

save_normalized_imagefile wraps the in-memory BytesIO with django.core.files.base.File rather than ContentFile. File wraps a file-like object without copying bytes, which is the intended optimisation. However, File.size delegates to the underlying file object's .size attribute or os.fstat(), both of which will raise UnsupportedOperation on a BytesIO. If any code path downstream calls instance.image.size before the FieldFile is saved to storage (e.g. in the MAX_IMAGE_FILE_SIZE guard in the view, or in model validation), it could raise an unhandled exception. ContentFile sets self.size = len(content) eagerly and avoids this entirely.

Is there a known safe order-of-operations here that prevents .size from being accessed on the pre-save File wrapper?

<sub>Found by: forge · CR-006</sub>

src/docker-entry.sh
```diff
@@ -11,6 +11,8 @@ gunicorn app.wsgi \
   --workers 2 \
   --threads 4 \
   --worker-class gthread \
+  --max-requests 1000 \
+  --max-requests-jitter 100 \
```
gridcell: **Question** · `correctness` · confidence: medium

**Does the max-requests threshold account for 4 threads per worker?**

With --worker-class gthread and --threads 4, each worker services requests concurrently across 4 threads. Gunicorn's --max-requests counter increments once per completed request regardless of which thread handled it, so a single worker recycles after approximately 1000 requests total — not 1000 per thread, which amounts to roughly 250 requests per thread on average. Whether 1000 is the right total depends on observed per-request memory delta. If image upload requests are large, 1000 total may be appropriately conservative; if most traffic is lightweight API calls, workers may be recycling more aggressively than necessary and causing unnecessary churn.

<sub>Found by: scaffold · CR-007</sub>

src/api/resources/classification/image.py
```diff
@@ -65,37 +66,43 @@ class Meta:
...
-            return annotation.classifier.patch_size
-        else:
-            return None
+        # All machine annotations on an image share one classifier, so the
+        # first one found is authoritative.
+        for point in obj.points.all():
+            for annotation in point.annotations.all():
+                if annotation.classifier_id is not None:
+                    return annotation.classifier.patch_size
+        return None
```
gridcell, line 86: **Question** · `maintainability` · confidence: medium

**get_patch_size invariant (one classifier per image) is asserted in a comment but not enforced**

The comment states 'all machine annotations on an image share one classifier' and uses the first one found as authoritative. Nothing in the data model or this code path enforces that constraint. _write_classification_results uses a single classifier_record for a given classification job (the normal path), but if a re-classification runs against the same image with a different Classifier version, or if data is imported externally, multiple classifier_id values could appear across annotations for a single image. In that case get_patch_size silently returns the patch size of whichever classifier happens to be attached to the first point/annotation encountered — non-deterministic.

<sub>Found by: forge · CR-008</sub>

src/api/resources/classification/image.py
```diff
@@ -65,37 +66,43 @@ class Meta:
...
                 count_unclassified += 1
 
-        return (
-            count_confirmed,
-            count_unconfirmed,
-            count_unclassified,
-        )
+        result = (count_confirmed, count_unconfirmed, count_unclassified)
+        self._counts_cache[obj.pk] = result
+        return result
```
gridcell, line 107: **Info** · `test_gap` · confidence: medium

**_counts_cache serializer-instance isolation has no unit test**

_summarize_counts lazily initialises a _counts_cache dict on the serializer instance and memoises results by obj.pk. A single test that calls get_num_confirmed / get_num_unconfirmed / get_num_unclassified twice on the same object and asserts that underlying queries are only executed once would both verify the cache works and guard against a future change that accidentally promotes the attribute to class scope (which would break per-request isolation).

<sub>Found by: gauntlet · CR-010</sub>

src/api/resources/classification/image.py
```diff
@@ -65,37 +66,43 @@ class Meta:
...
 
     def get_classification_status(self, obj):
-        latest_status = obj.statuses.order_by("-created_on").first()
-        if latest_status:
-            return ClassificationStatusSerializer(latest_status).data
+        statuses = obj.statuses.all()
+        latest = max(statuses, key=lambda s: s.created_on, default=None)
+        if latest:
+            return ClassificationStatusSerializer(latest).data
         return None
```
gridcell, line 77: **Info** · `test_gap` · confidence: medium

**get_classification_status empty-statuses case has no test**

get_classification_status switched from queryset ordering + .first() to Python max() with default=None. Both are correct for non-empty inputs, but the empty case now relies on max()'s default= keyword, which was added explicitly to avoid ValueError. The existing tests do not include a serializer-level test for an Image with zero associated statuses to confirm None is returned (rather than max raising ValueError without the default kwarg, which would be a silent regression if the default= were accidentally removed).

<sub>Found by: gauntlet · CR-011</sub>

src/api/resources/classification/image.py
```diff
@@ -183,6 +197,13 @@ def create(self, request, *args, **kwargs):
         collect_record_id = request.data.get("collect_record_id")
         trigger_classification = truthy(request.data.get("classify", True))
 
+        if image_file and image_file.size > settings.MAX_IMAGE_FILE_SIZE:
+            mb = settings.MAX_IMAGE_FILE_SIZE // (1024 * 1024)
+            return Response(
+                {"error": f"Image exceeds the {mb} MB size limit."},
+                status=status.HTTP_400_BAD_REQUEST,
+            )
+
```
gridcell, line 208: **Info** · `security` · confidence: medium

**MAX_IMAGE_FILE_SIZE is enforced after the upload is fully received by Django**

The new size guard is a useful application-layer cap, but it runs only after request.data.get('image') has already triggered multipart parsing — meaning Django has already streamed the entire body (to memory up to FILE_UPLOAD_MAX_MEMORY_SIZE, then to a temp file) before the 400 is returned. An attacker can therefore force the worker to receive arbitrarily large bodies (bounded only by the gunicorn 120 s timeout and any upstream proxy limit) before the server rejects them.

This is an architectural constraint, not a code defect — Content-Length can be omitted or spoofed, so the upstream proxy limit is the real defense. It is worth confirming there is an nginx client_max_body_size, ALB body-size cap, or CloudFront limit in front of the workers that enforces a hard ceiling before bodies reach Django.

<sub>Found by: sentinel · CR-012</sub>

src/api/resources/classification/image.py
gridcell, line 107: **Info** · `style` · confidence: high

**Missing return type annotation on _summarize_counts**

_summarize_counts computes and caches annotation counts, returning a 3-tuple of integers. The return type annotation is missing. Adding `-> tuple[int, int, int]` would make the contract explicit and consistent with the rest of the file's type-annotation style.

<sub>Found by: polisher · CR-014</sub>

src/api/utils/classification.py
saanobhaai: File(BytesIO).size works correctly via the seek/tell fallback — verified above. No issue.

src/docker-entry.sh
saanobhaai: @gridcell what do you think of Claude's answer? Maybe I should double as a compromise?

This is a valid observation about semantics, not a defect. The reviewer is correct: --max-requests 1000 counts total completed requests across all four threads in the worker process, so the effective per-thread count before recycling is ~250 on average.

Whether that's right depends on traffic volume:                                                                                                                               
  - At high sustained load (e.g. 10 req/sec per worker), 1000 requests is reached in ~100 seconds — workers cycling every couple of minutes, which is unnecessary churn         
  - At the kind of load MERMAID sees (moderate general API traffic + burst image uploads), it's probably fine as a starting point

The memory-growth argument that motivated --max-requests is per-process, not per-thread, so the counter being shared across threads is correct behavior. The question is purely whether 1000 is calibrated for your actual request rate.

No code change needed. What I'd suggest instead: once this is deployed, watch the Gunicorn logs for 'Worker with pid X was sent SIGTERM!' and count how frequently workers are recycling. If it's every few minutes under normal load, bump to --max-requests 4000 --max-requests-jitter 400. If it's every hour or so, 1000 is fine.

src/api/resources/classification/image.py
saanobhaai: `client_max_body_size` is 100M, and I think this is a tangential devops question not worth dealing with in this ticket.

src/api/resources/classification/image.py
saanobhaai: No other methods in this file carry annotations; adding one here would be inconsistent. Could address comprehensively in another ticket.

src/api/signals/classification.py
saanobhaai: Not worth doing. It would test that a private implementation attribute gets cleaned up, which is testing internals, not behavior. The end-to-end path (image created, thumbnail generated, correct bucket) is already covered by test_image_bucket_routing.py. If the cleanup logic broke, the behavior test would still pass — which tells you the cleanup test would be fragile and wouldn't catch real regressions anyway.

src/api/resources/classification/image.py
saanobhaai: Not worth doing. The concern was about accidental class-scope promotion, which is now moot since the CR-002 fix moves init to __init__ — class-scope promotion would require an explicit class-level statement. Testing that queries are only executed once is testing the cache mechanism rather than the behavior. A correctness test (right counts come back) is more valuable, and that's already implicitly covered by any test calling those fields.

gridcell (approved):

PR #684, opened 2026-05-15
[M1961] invert supporting viewsets

saanobhaai:
## Summary by CodeRabbit

* **New Features**
  * Added API endpoints for invertebrate belt-transect records and hierarchical invert attributes (ancestry, parent, species fields).
  * Enabled offline sync for invert-attribute data.
  * Search now includes belt-invert observer first/last name.
  * Choice list for group-of-interest now includes per-item updated_on; report export shows updated class/group-of-interest sourcing.

* **Documentation**
  * Added guidance for integrating an attribute taxonomy API.

* **Tests**
  * Added fixtures and tests for invert attributes, belt-transect CRUD, filtering, access control, and search.

[Review Change Stack](https://app.coderabbit.ai/change-stack/data-mermaid/mermaid-api/pull/684?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/new_protocol_readme.md
+++ b/new_protocol_readme.md
@@ -43,9 +43,12 @@
 
 If the protocol introduces its own attribute/taxonomy hierarchy (Django MTI from `BaseAttributeModel`), add:
 
-- A viewset (`BaseAttributeApiViewSet`) with a serializer that exposes all taxonomy levels and the full ancestry chain (e.g. `parent`, `class_goi`); use `select_related` across all MTI subtypes to avoid N+1 queries; exclude any internal-only nodes (e.g. backing models not part of the user-visible hierarchy)
+- A viewset (`BaseAttributeApiViewSet`) with a serializer that exposes all taxonomy levels and the full ancestry chain (e.g. `parent`, `class_goi`); use `select_related` across all MTI subtypes to avoid N+1 queries; exclude any internal-only nodes (e.g. backing models not part of the user-visible hierarchy) by filtering `.filter(<internal_model>__isnull=True)` — see `InvertAttributeViewSet` for the implementation pattern and `test_invert_attributes.py::test_list_invert_attributes_returns_user_visible_hierarchy` for the corresponding test pattern
 - Register the viewset in `urls.py` (global router, alongside `benthicattributes`)
-- Register in `src/api/resources/sync/views.py` `non_project_sources` so offline clients sync the taxonomy
+- Register in `src/api/resources/sync/views.py` `non_project_sources` so offline clients sync the taxonomy:
+  - `"read_only": False` — enables the propose-new-attribute flow for authenticated users
+  - `"visibility_filtered": True` — filters nodes based on status/visibility so only approved entries are synced to clients
+  - Add the source type constant to `CACHEABLE_SOURCE_TYPES` so pull responses are cached
 
 ## Choices Endpoint
 
--- a/new_protocol_readme.md
+++ b/new_protocol_readme.md
@@ -39,6 +39,14 @@
 
 - Update `annotate_num_sample_units()` in `src/api/resources/project.py` with a new UNION ALL branch counting sample units for the new protocol (joining through method → transect → sample event → site → project)
 
+## Attribute Taxonomy API (if protocol has its own MTI taxonomy)
+
+If the protocol introduces its own attribute/taxonomy hierarchy (Django MTI from `BaseAttributeModel`), add:
+
+- A viewset (`BaseAttributeApiViewSet`) with a serializer that exposes all taxonomy levels and the full ancestry chain (e.g. `parent`, `class_goi`); use `select_related` across all MTI subtypes to avoid N+1 queries; exclude any internal-only nodes (e.g. backing models not part of the user-visible hierarchy)
+- Register the viewset in `urls.py` (global router, alongside `benthicattributes`)
+- Register in `src/api/resources/sync/views.py` `non_project_sources` so offline clients sync the taxonomy
+
 ## Choices Endpoint
 
 - Import new lookup models in `src/api/resources/choices.py`
--- a/new_protocol_readme.md
+++ b/new_protocol_readme.md
@@ -43,7 +43,7 @@
 
 If the protocol introduces its own attribute/taxonomy hierarchy (Django MTI from `BaseAttributeModel`), add:
 
-- A viewset (`BaseAttributeApiViewSet`) with a serializer that exposes all taxonomy levels and the full ancestry chain (e.g. `parent`, `class_goi`); use `select_related` across all MTI subtypes to avoid N+1 queries; exclude any internal-only nodes (e.g. backing models not part of the user-visible hierarchy) by filtering `.filter(<internal_model>__isnull=True)` — see `InvertAttributeViewSet` for the implementation pattern and `test_invert_attributes.py::test_list_invert_attributes_returns_user_visible_hierarchy` for the corresponding test pattern
+- A viewset (`BaseAttributeApiViewSet`) with a serializer that exposes all taxonomy levels and the full ancestry chain (e.g. `parent`, `class_goi`); use `select_related` across all MTI subtypes to avoid N+1 queries — see `InvertAttributeViewSet` for the implementation pattern and `test_invert_attributes.py::test_list_invert_attributes_returns_user_visible_hierarchy` for the corresponding test pattern
 - Register the viewset in `urls.py` (global router, alongside `benthicattributes`)
 - Register in `src/api/resources/sync/views.py` `non_project_sources` so offline clients sync the taxonomy:
   - `"read_only": False` — enables the propose-new-attribute flow for authenticated users
--- a/src/api/migrations/0114_goi_data.py
+++ b/src/api/migrations/0114_goi_data.py
@@ -1,0 +1,105 @@
+from django.db import migrations
+
+SUPERUSER_APPROVED = 90
+
+
+def populate_genus_group_of_interest(apps, schema_editor):
+    """Walk genus → family → order → class_goi → group_of_interest and populate the new FK."""
+    InvertGenus = apps.get_model("api", "InvertGenus")
+    genera = list(InvertGenus.objects.select_related("family__order__class_goi__group_of_interest"))
+    for genus in genera:
+        genus.group_of_interest_id = genus.family.order.class_goi.group_of_interest_id
+    InvertGenus.objects.bulk_update(genera, ["group_of_interest_id"])
+
+
+def populate_order_invert_class(apps, schema_editor):
+    """Walk order → class_goi → invert_class and populate the new FK."""
+    InvertOrder = apps.get_model("api", "InvertOrder")
+    orders = list(InvertOrder.objects.select_related("class_goi__invert_class"))
+    for order in orders:
+        order.invert_class_id = order.class_goi.invert_class_id
+    InvertOrder.objects.bulk_update(orders, ["invert_class_id"])
+
+
+def promote_goi_to_invert_attribute(apps, schema_editor):
+    """Create an InvertAttribute row for each InvertGroupOfInterest using the same UUID.
+
+    By reusing the same UUID, any FK in InvertGenus.group_of_interest_id that was
+    populated in the previous step continues to resolve correctly after the PK swap
+    in migration 0115 — no FK value updates are needed.
+    """
+    InvertAttribute = apps.get_model("api", "InvertAttribute")
+    InvertGroupOfInterest = apps.get_model("api", "InvertGroupOfInterest")
+
+    gois = list(InvertGroupOfInterest.objects.all())
+    for goi in gois:
+        InvertAttribute.objects.create(id=goi.id, status=SUPERUSER_APPROVED)
+        goi.invertattribute_ptr_id = goi.id
+    InvertGroupOfInterest.objects.bulk_update(gois, ["invertattribute_ptr_id"])
+
+
+def deduplicate_invert_order(apps, schema_editor):
+    """Merge duplicate InvertOrder rows (same name, different class_goi) into one.
+
+    The canonical row is the one whose class_goi.group_of_interest.name sorts
+    alphabetically first, for determinism.  All InvertFamily rows pointing to
+    non-canonical orders are re-pointed to the canonical row before deletion.
+    """
+    InvertAttribute = apps.get_model("api", "InvertAttribute")
+    InvertOrder = apps.get_model("api", "InvertOrder")
+    InvertFamily = apps.get_model("api", "InvertFamily")
+
+    orders = list(
+        InvertOrder.objects.select_related("class_goi__group_of_interest").order_by("name")
+    )
+    by_name = {}
+    for order in orders:
+        by_name.setdefault(order.name, []).append(order)
+
+    for name, dups in by_name.items():
+        if len(dups) <= 1:
+            continue
+        canonical = min(dups, key=lambda o: o.class_goi.group_of_interest.name)
+        non_canonical_ids = [o.pk for o in dups if o.pk != canonical.pk]
+        InvertFamily.objects.filter(order_id__in=non_canonical_ids).update(order=canonical)
+        InvertAttribute.objects.filter(pk__in=non_canonical_ids).delete()
+
+
+def deduplicate_invert_family(apps, schema_editor):
+    """Merge duplicate InvertFamily rows (same name + order) created by order merge.
+
+    After order de-duplication both Muricidae rows share the same (name, order) pair.
+    The canonical row is chosen by lowest pk (alphabetical UUID sort).  All InvertGenus
+    rows pointing to non-canonical families are re-pointed before deletion.
+    """
+    InvertAttribute = apps.get_model("api", "InvertAttribute")
+    InvertFamily = apps.get_model("api", "InvertFamily")
+    InvertGenus = apps.get_model("api", "InvertGenus")
+
+    families = list(InvertFamily.objects.order_by("name", "order_id"))
+    by_key = {}
+    for family in families:
+        key = (family.name, str(family.order_id))
+        by_key.setdefault(key, []).append(family)
+
+    for key, dups in by_key.items():
+        if len(dups) <= 1:
+            continue
+        canonical = min(dups, key=lambda f: str(f.pk))
+        non_canonical_ids = [f.pk for f in dups if f.pk != canonical.pk]
+        InvertGenus.objects.filter(family_id__in=non_canonical_ids).update(family=canonical)
+        InvertAttribute.objects.filter(pk__in=non_canonical_ids).delete()
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("api", "0113_goi_schema_additions"),
+    ]
+
+    operations = [
+        migrations.RunPython(populate_genus_group_of_interest, migrations.RunPython.noop),
+        migrations.RunPython(populate_order_invert_class, migrations.RunPython.noop),
+        migrations.RunPython(promote_goi_to_invert_attribute, migrations.RunPython.noop),
+        migrations.RunPython(deduplicate_invert_order, migrations.RunPython.noop),
+        migrations.RunPython(deduplicate_invert_family, migrations.RunPython.noop),
+    ]
--- a/src/api/migrations/0115_goi_schema_finalize.py
+++ b/src/api/migrations/0115_goi_schema_finalize.py
@@ -1,0 +1,138 @@
+import django.db.models.deletion
+from django.db import migrations, models
+
+
+def delete_invert_class_group_of_interest(apps, schema_editor):
+    """Delete all InvertClassGroupOfInterest rows via their InvertAttribute parent.
+
+    Must run after RemoveField(invertorder, class_goi) so that the PROTECT FK from
+    InvertOrder no longer blocks deletion.
+    """
+    InvertAttribute = apps.get_model("api", "InvertAttribute")
+    InvertClassGroupOfInterest = apps.get_model("api", "InvertClassGroupOfInterest")
+    attr_ids = list(
+        InvertClassGroupOfInterest.objects.values_list("invertattribute_ptr_id", flat=True)
+    )
+    InvertAttribute.objects.filter(pk__in=attr_ids).delete()
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("api", "0114_goi_data"),
+    ]
+
+    operations = [
+        # 1. Make InvertGenus.group_of_interest non-nullable (all rows populated in 0114).
+        migrations.AlterField(
+            model_name="invertgenus",
+            name="group_of_interest",
+            field=models.ForeignKey(
+                on_delete=django.db.models.deletion.PROTECT,
+                to="api.invertgroupofinterest",
+            ),
+        ),
+        # 2. Make InvertOrder.invert_class non-nullable (all rows populated in 0114).
+        migrations.AlterField(
+            model_name="invertorder",
+            name="invert_class",
+            field=models.ForeignKey(
+                on_delete=django.db.models.deletion.PROTECT,
+                related_name="orders",
+                to="api.invertclass",
+            ),
+        ),
+        # 3. Remove the old class_goi FK from InvertOrder.
+        #    This unblocks deletion of InvertClassGroupOfInterest rows in step 4.
+        migrations.RemoveField(
+            model_name="invertorder",
+            name="class_goi",
+        ),
+        # 4. Delete InvertClassGroupOfInterest rows (and their InvertAttribute parents).
+        #    Safe now that no InvertOrder rows reference them.
+        migrations.RunPython(
+            delete_invert_class_group_of_interest,
+            migrations.RunPython.noop,
+        ),
+        # 5. Re-add unique constraint on InvertOrder(name, invert_class).
+        migrations.AddConstraint(
+            model_name="invertorder",
+            constraint=models.UniqueConstraint(
+                fields=["name", "invert_class"],
+                name="unique_invertorder_name_invert_class",
+            ),
+        ),
+        # 6. Re-add unique constraint on InvertFamily(name, order) — safe now that
+        #    Muricidae is no longer duplicated.
+        migrations.AddConstraint(
+            model_name="invertfamily",
+            constraint=models.UniqueConstraint(
+                fields=["name", "order"],
+                name="unique_invertfamily_name_order",
+            ),
+        ),
+        # 7. Promote invertattribute_ptr to PK on InvertGroupOfInterest.
+        #
+        #    Django cannot drop a primary key column directly, so this uses
+        #    SeparateDatabaseAndState:
+        #      - Database side: raw SQL to swap the PK and restore the FK from
+        #        invert_genus (which is cascade-dropped when id is dropped).
+        #      - State side: delete + recreate the model with the correct MTI bases
+        #        so future makemigrations detects no drift.
+        migrations.SeparateDatabaseAndState(
+            database_operations=[
+                migrations.RunSQL(
+                    sql="""
+                    -- Drop id column; CASCADE drops the old PK constraint and all FK
+                    -- constraints from other tables that reference invert_group_of_interest(id).
+                    ALTER TABLE invert_group_of_interest DROP COLUMN id CASCADE;
+
+                    -- Make invertattribute_ptr_id the new primary key.
+                    ALTER TABLE invert_group_of_interest
+                        ADD PRIMARY KEY (invertattribute_ptr_id);
+
+                    -- Restore the FK from invert_genus.group_of_interest_id.
+                    -- The UUID values are unchanged (same UUIDs used for both id and
+                    -- invertattribute_ptr_id in migration 0114), so no data updates needed.
+                    ALTER TABLE invert_genus
+                        ADD CONSTRAINT invert_genus_group_of_interest_id_fk
+                        FOREIGN KEY (group_of_interest_id)
+                        REFERENCES invert_group_of_interest (invertattribute_ptr_id);
+                    """,
+                    reverse_sql=migrations.RunSQL.noop,
+                ),
+            ],
+            state_operations=[
+                # Recreate InvertGroupOfInterest as a proper InvertAttribute MTI child
+                # so that the migration state matches the model code.
+                migrations.DeleteModel(name="InvertGroupOfInterest"),
+                migrations.CreateModel(
+                    name="InvertGroupOfInterest",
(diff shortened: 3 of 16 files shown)


saanobhaai:
[Inverts Phase 3: Supporting Viewsets](https://trello.com/c/f37BDGK7/1961-inverts-phase-3-supporting-viewsets)

src/api/resources/sync/views.py
```diff
@@ -105,6 +108,12 @@
         "required_filters": NO_FILTERS,
         "read_only": False,
     },
+    INVERT_ATTRIBUTES_SOURCE_TYPE: {
+        "view": invert_attribute.InvertAttributeViewSet,
+        "required_filters": NO_FILTERS,
+        "read_only": False,
+        "visibility_filtered": True,
+    },
```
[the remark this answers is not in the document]
saanobhaai: This is like benthic attributes -- users will be able to propose new attributes via the collect app ("propose new species" flow). Fish species is the same pattern. Fish families/genera/groupings are not; those are controlled by admins only.

src/api/tests/test_invert_attributes.py
```diff
@@ -0,0 +1,146 @@
...
+    data = response.json()
+    assert data["count"] == 1
+    assert str(data["results"][0]["id"]) == str(invert_belt_transect1.pk)
+
+
+def test_invert_belt_transect_create(
+    db_setup, api_client1, project1, sample_event1, invert_belt_transect_width_1m, invert_size_bin_1
+):
+    """POST creates a new InvertBeltTransect."""
+    payload = {
```
gridcell, line 111: **Warning** · `security` · confidence: high

**No cross-project IDOR test for InvertBeltTransect**

`api_client1` is a member of `project1`. There is no test asserting that a user authenticated to a different project cannot read, modify, or delete `project1`'s invertbelttransects via the project-scoped URL. `BaseProjectApiViewSet` filters by project membership, but an explicit regression test locks that behavior in place for the new endpoint.

> **Suggestion:** Add `def test_invertbelttransect_not_accessible_cross_project(db_setup, api_client2, project1, invert_belt_transect1): response = api_client2.get(f"/v1/projects/{project1.pk}/invertbelttransects/", format="json"); assert response.status_code in (403, 404)`

> **Why it matters:** Without this test an IDOR regression in `BaseProjectApiViewSet` filtering would go undetected for the new endpoint.

<sub>Found by: gauntlet · CR-002</sub>

src/api/resources/sync/views.py
```diff
@@ -105,6 +108,12 @@
         "required_filters": NO_FILTERS,
         "read_only": False,
     },
+    INVERT_ATTRIBUTES_SOURCE_TYPE: {
+        "view": invert_attribute.InvertAttributeViewSet,
+        "required_filters": NO_FILTERS,
```
gridcell, line 118: **Warning** · `correctness` · confidence: medium

**visibility_filtered: True on INVERT_ATTRIBUTES_SOURCE_TYPE has no peer — what does the flag do and is it consumed?**

This `INVERT_ATTRIBUTES_SOURCE_TYPE` entry in `non_project_sources` includes `"visibility_filtered": True`. No other entry in `non_project_sources` (including `BENTHIC_ATTRIBUTES_SOURCE_TYPE` and `FISH_SPECIES_SOURCE_TYPE`) carries this key. Either (a) this is a new mechanism introduced alongside this PR but not visible in the diff, meaning there is no consumer of the flag yet, or (b) it is a copy-paste artifact. Please clarify what the flag does and confirm there is a consumer.

> **Suggestion:** Clarify in the PR description what `visibility_filtered` does and confirm a consumer reads this key during the sync response. If no consumer exists yet, remove the key until the consuming code lands.

> **Why it matters:** An undocumented config key with no peer and no visible consumer is either dead code (harmless but confusing) or a load-bearing flag with no tests — both warrant clarification. See related concern CR-003 in the review body for the cache+visibility risk.

<sub>Found by: forge · CR-004</sub>

src/api/resources/invert_belt_transect.py
```diff
@@ -0,0 +1,50 @@
...
+
+class InvertBeltTransectSerializer(SampleUnitSerializer):
+    class Meta:
+        model = InvertBeltTransect
+        exclude = []
+        extra_kwargs = {
+            "number": {"error_messages": {"null": "Transect number is required"}},
+            "len_surveyed": {"error_messages": {"null": "Transect length surveyed is required"}},
+            "width": {"error_messages": {"null": "Width is required"}},
+        }
```
gridcell, line 29: **Warning** · `correctness` · confidence: medium

**extra_kwargs.update(parent) puts parent last, so parent keys silently overwrite child keys**

Inside the nested `Meta` class, `extra_kwargs` is built as a dict literal first (with child-specific error messages for `number`, `len_surveyed`, and `width`), then `.update(SampleUnitSerializer.extra_kwargs)` is called. Because `.update()` overwrites existing keys with the incoming values, any key present in both dicts will have the parent's value in the final result — silently discarding the child's customization.

Currently `SampleUnitSerializer.extra_kwargs = {}` so no messages are lost today. The pattern is fragile, though: if a future commit adds an entry for `number` or `len_surveyed` to `SampleUnitSerializer.extra_kwargs`, the child's error messages will be silently overwritten. The identical pattern exists in `FishBeltTransectSerializer`.

> **Suggestion:** Reverse the merge so the child's entries win:
```python
extra_kwargs = {**SampleUnitSerializer.extra_kwargs, **{
    "number": {"error_messages": {"null": "Transect number is required"}},
    "len_surveyed": {"error_messages": {"null": "Transect length surveyed is required"}},
    "width": {"error_messages": {"null": "Width is required"}},
}}
```
Alternatively, document explicitly that parent-overrides-child is intentional.

> **Why it matters:** When `SampleUnitSerializer.extra_kwargs` is eventually extended, the child's custom error messages will be lost silently, producing confusing API validation errors for end users.

<sub>Found by: forge · CR-005</sub>

src/api/tests/test_invert_attributes.py
```diff
@@ -0,0 +1,146 @@
...
+    response = api_client1.get(
+        f"/v1/projects/{project1.pk}/invertbelttransects/{invert_belt_transect1.pk}/",
+        format="json",
+    )
+    assert response.status_code == 200
+    data = response.json()
+    assert str(data["id"]) == str(invert_belt_transect1.pk)
+    assert str(data["width"]) == str(invert_belt_transect_width_1m.pk)
+    assert data["number"] == 1
+
```
gridcell, line 142: **Warning** · `test_gap` · confidence: high

**No PATCH or PUT test for InvertBeltTransect**

The test suite covers POST, GET list, GET detail, and DELETE. No PATCH/PUT update test exists. `InvertBeltTransectSerializer` declares custom `error_messages` for `number`, `len_surveyed`, and `width` exercised during writes — PATCH/PUT tests would catch misconfiguration of those messages or unexpected read-only field promotion.

> **Suggestion:** Add a PATCH test that updates a mutable field (e.g. `depth`) and asserts the new value is returned with status 200.

> **Why it matters:** Mutation paths are where serializer validation and field exclusion bugs surface. Skipping update tests leaves the write path entirely unverified.

<sub>Found by: gauntlet · CR-006</sub>

src/api/resources/invert_belt_transect.py
```diff
@@ -0,0 +1,50 @@
...
+    len_surveyed = django_filters.RangeFilter(field_name="len_surveyed")
+
+    class Meta:
+        model = InvertBeltTransect
+        fields = [
+            "beltinvert_method",
+            "sample_event",
+            "len_surveyed",
+            "width",
+            "size_bin",
```
gridcell, line 42: **Warning** · `test_gap` · confidence: medium

**InvertBeltTransectFilterSet fields are entirely untested end-to-end**

`InvertBeltTransectFilterSet` adds a `len_surveyed` `RangeFilter` plus `width`, `size_bin`, and `depth` filters. None are exercised by the existing tests. The list test only asserts `count == 1` with a single fixture, which cannot distinguish a working filter from one that silently ignores the query parameter.

> **Suggestion:** Add parametrized filter tests covering `len_surveyed_min`/`max`, `depth`, `width`, and `size_bin` — each with one match and one no-match case.

> **Why it matters:** A filterset registered but never end-to-end tested can have silently broken field names that only surface when real clients attempt to use them.

<sub>Found by: gauntlet · CR-007</sub>

src/api/tests/test_invert_attributes.py
```diff
@@ -0,0 +1,146 @@
...
+    assert "class" not in ranks
+    assert ranks == {"class_goi", "order", "family", "genus", "species"}
+
+
+def test_invert_attribute_species_fields(
+    db_setup, api_client1, all_test_invert_attributes, invert_species_1
+):
+    """Species nodes include max_length, max_length_type, notes; others return None."""
+    response = api_client1.get("/v1/invertattributes/", format="json")
+    data = response.json()
```
gridcell, line 25: **Question** · `correctness` · confidence: medium

**float() cast on a Decimal field may hide precision drift in assertions**

`float(species_rec["max_length"])` and `float(invert_species_1.max_length)` both convert to IEEE-754 float before comparing. If `max_length` is a `DecimalField`, the serializer returns a string and the model holds `Decimal`. The double-float cast masks any future precision mismatch. Is this intentional, or should the assertion use `Decimal` comparison or direct string equality?

<sub>Found by: gauntlet · CR-009</sub>

new_protocol_readme.md
```diff
@@ -39,6 +39,14 @@
...
 - Update `annotate_num_sample_units()` in `src/api/resources/project.py` with a new UNION ALL branch counting sample units for the new protocol (joining through method → transect → sample event → site → project)
 
+## Attribute Taxonomy API (if protocol has its own MTI taxonomy)
+
+If the protocol introduces its own attribute/taxonomy hierarchy (Django MTI from `BaseAttributeModel`), add:
+
+- A viewset (`BaseAttributeApiViewSet`) with a serializer that exposes all taxonomy levels and the full ancestry chain (e.g. `parent`, `class_goi`); use `select_related` across all MTI subtypes to avoid N+1 queries; exclude any internal-only nodes (e.g. backing models not part of the user-visible hierarchy)
+- Register the viewset in `urls.py` (global router, alongside `benthicattributes`)
+- Register in `src/api/resources/sync/views.py` `non_project_sources` so offline clients sync the taxonomy
+
```
gridcell, line 52: **Info** · `docs` · confidence: high

**Sync registration config flags not documented in protocol checklist**

The new section instructs adding the source to `non_project_sources` in `src/api/resources/sync/views.py` but omits three configuration details used by the implementation: `read_only: False` (enables the propose-new-attribute flow), `visibility_filtered: True`, and adding the source to `CACHEABLE_SOURCE_TYPES`. A future implementer following this checklist would likely skip these flags.

> **Suggestion:** Append a sub-bullet documenting all three flags and their purpose so the checklist is self-contained.

<sub>Found by: scribe · CR-010</sub>

new_protocol_readme.md
```diff
@@ -39,6 +39,14 @@
 
 - Update `annotate_num_sample_units()` in `src/api/resources/project.py` with a new UNION ALL branch counting sample units for the new protocol (joining through method → transect → sample event → site → project)
 
+## Attribute Taxonomy API (if protocol has its own MTI taxonomy)
+
+If the protocol introduces its own attribute/taxonomy hierarchy (Django MTI from `BaseAttributeModel`), add:
+
```
gridcell, line 45: **Info** · `docs` · confidence: medium

**Filtering technique generalized without referencing the implementation or test pattern**

The bullet about excluding internal-only nodes generalizes the concept but does not point to the implementation pattern (`.filter(<internal-model>__isnull=True)`) or recommend a test pattern. Consider citing `test_invert_attributes.py::test_list_invert_attributes_returns_user_visible_hierarchy` as a template for verifying taxonomy visibility in future protocols.

<sub>Found by: scribe · CR-011</sub>

src/api/resources/invert_belt_transect.py
gridcell, line 42: **Info** · `maintainability` · confidence: high

**InvertBeltTransectFilterSet includes size_bin; FishBeltTransectFilterSet does not**

`InvertBeltTransectFilterSet.fields` includes `"size_bin"`. The peer `FishBeltTransectFilterSet` does not list `size_bin` even though `FishBeltTransect` also has the field. If this is intentional (invert workflow uses `size_bin` as a primary filter dimension), no action — but if it is a gap in the fish peer now being corrected only for invert, the asymmetry may confuse future maintainers.

<sub>Found by: forge · CR-012</sub>

gridcell (commented):
## AI Code Review

**Assessment:** The invert supporting viewsets are well-structured and faithfully follow peer patterns. No blocking issues exist. The primary areas needing attention are a potential cache/visibility key interaction risk for `INVERT_ATTRIBUTES_SOURCE_TYPE` (question-level, needs author confirmation), a cluster of missing auth/IDOR/mutation/filter test coverage (warning-level), and partial documentation of sync config flags. The `select_related` chain is exemplary and the fixture hierarchy is clean.

| 0 | 5 | 3 | 3 | 2 |

**Files reviewed:** 9 / 9

### Cross-Cutting Concerns

- Cache and visibility are coupled: `INVERT_ATTRIBUTES_SOURCE_TYPE` is the only entry that is both in `CACHEABLE_SOURCE_TYPES` and has `visibility_filtered: True`. The missing auth/visibility test coverage (CR-001) and the cache key risk (CR-003) are two sides of the same gap.
- This PR is the "supporting viewsets" half of the invert protocol integration. Method-level (`BeltInvertViewSet`) and observation-level endpoints are absent (CR-008). Confirm whether a follow-up ticket tracks the remaining work.
- The sync configuration documentation in `new_protocol_readme.md` is partially updated — it captures the `non_project_sources` registration step but omits the three flags the implementation uses: `read_only`, `visibility_filtered`, and `CACHEABLE_SOURCE_TYPES` membership (CR-010).

### General Findings

<details>
<summary><strong>CR-001</strong> &middot; Warning &middot; <code>security</code> — No authentication or authorization tests for either new endpoint</summary>

Neither `/v1/invertattributes/` nor `/v1/projects/{id}/invertbelttransects/` has a test verifying that an unauthenticated caller receives 401/403, that a non-project-member receives 403/404 on the project-scoped endpoint, or that the `visibility_filtered` flag on `InvertAttribute` actually filters rows per user. The permission classes are inherited from the base viewsets, but explicit tests document the security contract and catch future viewset misconfiguration. This is especially important given the cache/visibility interaction question raised in CR-003: a test asserting per-user scoping is the strongest defense against a caching regression silently leaking rows.

> **Suggestion:** Add at minimum: (1) anonymous-caller 401 test on `/v1/invertattributes/`, (2) non-member 403/404 test on `/v1/projects/{id}/invertbelttransects/`, (3) a visibility test demonstrating per-user filtering of `InvertAttribute` rows.

> **Why it matters:** Without auth tests the permission boundary is untested. A future change to viewset mixins or `DEFAULT_AUTHENTICATION_CLASSES` could silently open the endpoint to anonymous callers.

Applies to: `src/api/tests/test_invert_attributes.py`

<sub>Found by: sentinel, gauntlet</sub>
</details>

<details>
<summary><strong>CR-003</strong> &middot; Warning &middot; <code>security</code> — INVERT_ATTRIBUTES_SOURCE_TYPE is the only cacheable source with visibility_filtered: True — cache key may not incorporate the visibility dimension</summary>

In `src/api/resources/sync/views.py` the new `INVERT_ATTRIBUTES_SOURCE_TYPE` is added to both `CACHEABLE_SOURCE_TYPES` and given `visibility_filtered: True` in `non_project_sources`. Every other cacheable entry is global — none sets `visibility_filtered`. If the sync cache layer keys responses only by `source_type` or `source_type + revision_num` without folding the per-user or per-project visibility dimension into the cache key, a response computed for one user could be served to another user with a different visibility scope, leaking `InvertAttribute` rows the second user is not entitled to see.

> **Suggestion:** Either (a) confirm the sync cache key for `visibility_filtered` sources includes the per-user/per-project dimension, or (b) remove `INVERT_ATTRIBUTES_SOURCE_TYPE` from `CACHEABLE_SOURCE_TYPES` until that is confirmed.

> **Why it matters:** Cache poisoning / cross-user data leakage is a classic side effect of caching responses whose contents depend on the requester's authorization scope. The structural divergence from every other cacheable entry is exactly the kind of pattern break that creates this bug class. (CWE-524, CWE-639)

Applies to: `src/api/resources/sync/views.py` (this spans the `CACHEABLE_SOURCE_TYPES` tuple and the `non_project_sources` entry)

<sub>Found by: sentinel</sub>
</details>

<details>
<summary><strong>CR-008</strong> &middot; Question &middot; <code>api_contract</code> — No method-level or obs-level endpoints in this PR — intentionally deferred?</summary>

This PR registers `invertbelttransects` (sample-unit layer) and `invertattributes` (taxonomy layer) but there is no method-level viewset (`BeltInvertViewSet`) and no observation-level endpoints. The `new_protocol_readme.md` checklist requires Obs/SU/SE summary endpoints for a complete protocol integration. Is this intentional — with method/obs endpoints tracked in a follow-up — or were they accidentally omitted?

Applies to: `src/api/urls.py`, `new_protocol_readme.md`

<sub>Found by: forge</sub>
</details>

<details>
<summary><strong>CR-013</strong> &middot; Info &middot; <code>security</code> — No secrets, SQL injection, XSS, or infrastructure changes detected</summary>

Confirmed clean across the diff: no hardcoded credentials, API keys, or `.env` values; no raw SQL or string-interpolated queries (all ORM with `select_related`/`filter`); no template rendering; no Dockerfile/IaC/CORS/header changes; no new third-party dependencies. The `parent` `SerializerMethodField` exposes only related-model PKs (UUIDs), matching the disclosure surface of peer attribute serializers.

<sub>Found by: sentinel</sub>
</details>

<details>
<summary><strong>CR-014</strong> &middot; Info &middot; <code>test_gap</code> — Search-field test is structural only — no end-to-end search behavior verified</summary>

`test_sample_unit_method_search_fields_include_beltinvert` only asserts the new paths are present in `SEARCH_FIELDS`. It does not verify that a search query actually filters sample unit method results by beltinvert observer name. A functional search test would catch regressions in the ORM path (missing `related_name`, typo, or missing `select_related`).

Applies to: `src/api/tests/test_invert_attributes.py`

<sub>Found by: gauntlet</sub>
</details>

<details>
<summary><strong>CR-015</strong> &middot; Praise &middot; <code>maintainability</code> — Fixture hierarchy correctly mirrors the taxonomy MTI chain</summary>

The fixture dependency chain (`invert_class_1` and `invert_group_of_interest_1` → `invert_class_goi_1` → `invert_order_1` → `invert_family_1` → `invert_genus_1` → `invert_species_1`) precisely models the domain MTI hierarchy. Test failures surface at the exact taxonomy level rather than producing opaque DB constraint errors. The `all_test_invert_attributes` aggregate fixture is a clean pattern.

Applies to: `src/api/tests/fixtures/macroinvertebrate.py`

<sub>Found by: gauntlet</sub>
</details>

<details>
<summary><strong>CR-016</strong> &middot; Praise &middot; <code>performance</code> — select_related chain fully covers all serializer traversals — no N+1 queries</summary>

The `InvertAttributeViewSet.queryset` builds a comprehensive `select_related` across all six MTI subtype paths. Every object traversal in `get_class_goi()` and `get_parent()` either accesses an FK `_id` column (auto-fetched without a JOIN) or traverses a chain exactly covered by the `select_related` declaration. Full taxonomy list serialization will issue exactly one SQL query regardless of result set size.

Applies to: `src/api/resources/invert_attribute.py`

<sub>Found by: forge</sub>
</details>

### Agent Coverage

| sentinel | security | 9 | 3 | 0 |
| forge | correctness | 9 | 5 | 0 |
| polisher | style | 8 | 0 | 0 |
| gauntlet | test_coverage | 3 | 7 | 0 |
| scribe | documentation | 1 | 2 | 0 |

gridcell (commented):
I've gone through and ran my ai-reviewer.  I've resolved some of the comments from the ai review but left a couple I thought were worthin considering

saanobhaai:
Resolved outstanding ai-reviewer comments after attempting to address

new_protocol_readme.md
```diff
@@ -39,6 +39,17 @@
...
+- A viewset (`BaseAttributeApiViewSet`) with a serializer that exposes all taxonomy levels and the full ancestry chain (e.g. `parent`, `class_goi`); use `select_related` across all MTI subtypes to avoid N+1 queries; exclude any internal-only nodes (e.g. backing models not part of the user-visible hierarchy) by filtering `.filter(<internal_model>__isnull=True)` — see `InvertAttributeViewSet` for the implementation pattern and `test_invert_attributes.py::test_list_invert_attributes_returns_user_visible_hierarchy` for the corresponding test pattern
```
gridcell: **warning** · `docs` · confidence: high

**Readme cites InvertAttributeViewSet as an example of `.filter(internal_model__isnull=True)`, but this PR removes that filter**

The expanded guidance here tells future protocol authors to exclude internal-only nodes with `.filter(<internal_model>__isnull=True)` and points to `InvertAttributeViewSet` as the canonical implementation. After commit 4bda225d, `InvertAttributeViewSet.queryset` has no `.filter(...)` at all — the test `test_list_invert_attributes_returns_user_visible_hierarchy` now expects 6 ranks including `class` (previously excluded). The referenced viewset no longer demonstrates the pattern the prose describes.

> **Suggestion:** Either (a) drop the `.filter(<internal_model>__isnull=True)` paragraph if the new design is that all MTI subtypes are user-visible; or (b) keep the guidance generic and reference a different protocol where the filter is still used; or (c) keep both but note that `InvertAttribute` no longer needs to exclude any subtypes after the GoI promotion.

> **Why it matters:** Future protocol authors reading this checklist will follow the link to `InvertAttributeViewSet`, see no filtering, and be confused about what the pattern actually is.

<sub>Found by: scribe · CR-001</sub>

src/api/migrations/0114_goi_data.py
```diff
@@ -0,0 +1,105 @@
...
+        key = (family.name, str(family.order_id))
+        by_key.setdefault(key, []).append(family)
+
+    for key, dups in by_key.items():
+        if len(dups) <= 1:
+            continue
+        canonical = min(dups, key=lambda f: str(f.pk))
+        non_canonical_ids = [f.pk for f in dups if f.pk != canonical.pk]
+        InvertGenus.objects.filter(family_id__in=non_canonical_ids).update(family=canonical)
+        InvertAttribute.objects.filter(pk__in=non_canonical_ids).delete()
```
gridcell, line 91: **question** · `correctness` · confidence: medium

**deduplicate_invert_order/family will fail if any ObsBeltInvert references the orders/families being deleted**

`deduplicate_invert_order` re-points `InvertFamily.order` to the canonical row, then calls `InvertAttribute.objects.filter(pk__in=non_canonical_ids).delete()`. Because each `InvertOrder` is also an `InvertAttribute` MTI child, and `ObsBeltInvert.invert_attribute = ForeignKey(InvertAttribute, on_delete=PROTECT)`, any observation row that references a non-canonical order (or family in the analogous `deduplicate_invert_family`) will raise `ProtectedError` at migration time. The migration implicitly assumes no such observations exist in production. The same concern applies to `deduplicate_invert_family` — genus FKs are re-pointed, but observations against the duplicate families/genera themselves are not.

Evidence:
- `src/api/migrations/0114_goi_data.py:65` — `InvertAttribute.objects.filter(pk__in=non_canonical_ids).delete()`
- `src/api/models/protocols/macroinvertebrate.py:341` — `invert_attribute = models.ForeignKey(InvertAttribute, on_delete=models.PROTECT)`

> **Suggestion:** Either (a) add an explicit pre-flight check (`if ObsBeltInvert.objects.filter(invert_attribute_id__in=non_canonical_ids).exists(): raise RuntimeError(...)`) to fail fast with a clear message, or (b) re-point any matching `ObsBeltInvert.invert_attribute` to the canonical row before deleting, or (c) document/confirm in the migration docstring that this is safe because the Muricidae duplication only existed in reference data and no observations have ever referenced the non-canonical rows.

> **Suggestion trade-offs:** Option (b) silently moves observation rows to a new taxon — confirm with domain owners that's the intended outcome before applying. Option (a) is the safest default.

> **Why it matters:** If any prod database has an obs row pointing to a duplicate order/family, deploy of this migration will fail mid-flight and require manual recovery.

<sub>Found by: bedrock · CR-002</sub>

src/api/migrations/0114_goi_data.py
```diff
@@ -0,0 +1,105 @@
...
+    in migration 0115 — no FK value updates are needed.
+    """
+    InvertAttribute = apps.get_model("api", "InvertAttribute")
+    InvertGroupOfInterest = apps.get_model("api", "InvertGroupOfInterest")
+
+    gois = list(InvertGroupOfInterest.objects.all())
+    for goi in gois:
+        InvertAttribute.objects.create(id=goi.id, status=SUPERUSER_APPROVED)
+        goi.invertattribute_ptr_id = goi.id
+    InvertGroupOfInterest.objects.bulk_update(gois, ["invertattribute_ptr_id"])
```
gridcell, line 38: **info** · `performance` · confidence: medium

**bulk_update calls omit batch_size**

`populate_genus_group_of_interest`, `populate_order_invert_class`, and `promote_goi_to_invert_attribute` each call `bulk_update` over the entire result set in a single statement. For the invert taxonomy this is small (hundreds of rows at most), so this is unlikely to be a practical issue — flagging only because the codebase's other large bulk_update calls do pass batch_size, and consistency helps future maintainers.

<sub>Found by: bedrock · CR-003</sub>

src/api/migrations/0115_goi_schema_finalize.py
```diff
@@ -0,0 +1,138 @@
...
+                    ALTER TABLE invert_genus
+                        ADD CONSTRAINT invert_genus_group_of_interest_id_fk
+                        FOREIGN KEY (group_of_interest_id)
+                        REFERENCES invert_group_of_interest (invertattribute_ptr_id);
+                    """,
+                    reverse_sql=migrations.RunSQL.noop,
+                ),
+            ],
+            state_operations=[
+                # Recreate InvertGroupOfInterest as a proper InvertAttribute MTI child
```
gridcell, line 105: **info** · `security` · confidence: high

**Recreated FK omits ON DELETE clause (NO ACTION default)**

The raw-SQL FK on `invert_genus.group_of_interest_id` is recreated without an explicit `ON DELETE` clause, so PostgreSQL defaults to `NO ACTION`. Django's `on_delete=PROTECT` is enforced at the application layer regardless, so this is functionally equivalent and matches what Django would generate. Flagging only for awareness — if anyone ever runs a direct DB-level delete on `invert_group_of_interest` that bypasses the ORM, NO ACTION will still block it (semantically similar to PROTECT), but the error messaging will differ.

<sub>Found by: sentinel · CR-004</sub>

gridcell (commented):
## AI Code Review

**Scope:** last 3 commits (`a87457f9..4bda225d`) — the GoI schema refactor.

**Assessment:** The 3-commit GoI schema refactor is well-structured: a phased 0113→0114→0115 migration with clear inline comments, a clean serializer rewrite, and good new test coverage including the IDOR fix. One real concern: the migration's order-dedup step has an implicit assumption that no `ObsBeltInvert` rows reference the duplicate orders/families being deleted (PROTECT FK would otherwise block). Also one doc/code mismatch: the protocol readme references a `.filter(internal_model__isnull=True)` pattern that was removed from `InvertAttributeViewSet` in this same PR. Neither is merge-blocking but both warrant a quick author response.

| 0 | 1 | 1 | 2 | 3 |

**Files reviewed:** 12 / 12
**Previously addressed:** 9 finding(s) suppressed (resolved or acknowledged in prior review)

### Agent Coverage
| sentinel | security | 12 | 2 | 0 |
| forge | backend | 6 | 1 | 0 |
| bedrock | database | 4 | 2 | 0 |
| polisher | code_quality | 11 | 0 | 0 |
| gauntlet | testing | 2 | 1 | 0 |
| scribe | documentation | 1 | 1 | 0 |

src/api/migrations/0114_goi_data.py
gridcell, line 38: **praise** · `maintainability` · confidence: high

**Phased 0113/0114/0115 migration design with clear inline rationale**

The split into (1) schema additions with nullable FKs, (2) data population + dedupe + MTI promotion via shared UUIDs, (3) schema finalization with non-null constraints and the PK swap is a textbook safe-migration pattern. The inline comments explaining *why* each step is ordered the way it is — particularly the explanation of UUID reuse so FK values stay valid across the PK swap — make this very easy to review and to revert mentally if needed.

<sub>Found by: bedrock · CR-005</sub>

src/api/tests/test_invert_attributes.py
```diff
@@ -0,0 +1,228 @@
...
+    assert not InvertBeltTransect.objects.filter(pk=pk).exists()
+
+
+def test_invertbelttransect_not_accessible_cross_project(
+    db_setup, api_client3, project1, invert_belt_transect1
+):
+    """User with no membership in project1 cannot read project1's invertbelttransects."""
+    response = api_client3.get(f"/v1/projects/{project1.pk}/invertbelttransects/", format="json")
+    assert response.status_code == 403
+
```
gridcell, line 161: **praise** · `security` · confidence: high

**Cross-project IDOR test added for InvertBeltTransect**

`test_invertbelttransect_not_accessible_cross_project` exercises the cross-project authorization path explicitly with a non-member user. Good addition — closes the gap raised in the earlier review.

<sub>Found by: sentinel · CR-006</sub>

src/api/resources/invert_attribute.py
```diff
@@ -0,0 +1,109 @@
...
+
+        species = getattr(obj, "invertspecies", None)
+        if species is not None:
+            return species.genus.group_of_interest_id
+
+        genus = getattr(obj, "invertgenus", None)
+        if genus is not None:
+            return genus.group_of_interest_id
+
+        return None
```
gridcell, line 72: **praise** · `correctness` · confidence: high

**Serializer cleanly models the new hierarchy: parent/group_of_interest each return None where appropriate**

`get_parent` returns None for both class and goi (the two roots of the new hierarchy), and `get_group_of_interest` returns None for class/order/family (which sit on the other branch). The tests assert exactly this, so the semantics are crisp.

<sub>Found by: forge · CR-007</sub>

src/api/migrations/0114_goi_data.py
saanobhaai: Disregarding -- nothing has been entered in production.

src/api/migrations/0114_goi_data.py
saanobhaai: Disregarding as overkill in this context.

gridcell (approved):

PR #688, opened 2026-05-20
[M1962] inverts submission/validation

saanobhaai:
## Summary by CodeRabbit

* **New Features**
  * Full support for macroinvertebrate belt-transect surveys: CSV ingest, submission writer, and comprehensive validation for transects and observations.

* **Changes**
  * Choice data: the prior "invertharvesttypes" choice set has been removed from choices/results.

* **Tests**
  * Added unit tests covering macroinvertebrate ingest and validation flows.

[Review Change Stack](https://app.coderabbit.ai/change-stack/data-mermaid/mermaid-api/pull/688?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


saanobhaai:
@coderabbitai review

src/api/ingest/macroinvertebrate_serializer.py
```diff
@@ -0,0 +1,166 @@
...
+__all__ = ["MacroInvertebrateCSVSerializer"]
+
+
+class MacroInvertebrateCSVSerializer(CollectRecordCSVSerializer):
+    protocol = MACROINVERTEBRATE_PROTOCOL
+    sample_unit = "beltinvert_transect"
+    observations_fields = ["data__obs_belt_inverts"]
+    additional_group_fields = CollectRecordCSVSerializer.additional_group_fields.copy()
+    additional_group_fields.append("data__beltinvert_transect__label")
+    composite_fields = {"data__sample_event__sample_date": ["year", "month", "day"]}
```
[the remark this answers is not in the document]
saanobhaai: Ruff only enables E4/E7/E9/F; RUF012 is not active. Identical pattern in FishBeltCSVSerializer unchanged.

src/api/resources/sampleunitmethods/beltinvertmethod.py
```diff
@@ -0,0 +1,36 @@
...
+from ...models import BeltInvert, ObsBeltInvert
+from ..base import BaseAPISerializer
+
+__all__ = ["BeltInvertSerializer", "ObsBeltInvertSerializer"]
+
+
+class BeltInvertSerializer(BaseAPISerializer):
+    class Meta:
+        model = BeltInvert
+        exclude = []
```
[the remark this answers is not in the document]
saanobhaai: Same reason; exclude = [] is standard DRF convention and matches beltfishmethod.py.

src/api/submission/validations/validators/invert_belt_transect.py
```diff
@@ -0,0 +1,103 @@
...
+
+        return {
+            "sample_event__site": site,
+            "sample_event__management": management,
+            "sample_event__sample_date": sample_date,
+            "number": number,
+            "label": label,
+            "depth": depth,
+            "width_id": width,
+        }, profiles
```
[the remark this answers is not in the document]
saanobhaai: UniqueFishbeltTransectValidator._get_query_args has the identical pattern — relative_depth_path is stored but excluded from the query. This is a deliberate pre-existing design choice, not introduced here.

src/api/tests/test_validations.py
```diff
@@ -232,3 +233,240 @@ def test_habcomp_protocol_validation_ok(valid_habitat_complexity_collect_record,
...
+
+
+def test_belt_invert_density_per_goi_warn(
+    valid_belt_invert_collect_record,
+    profile1_request,
+    invert_species_1,
+    invert_belt_transect_width_1m,
+):
+    # invert_species_1 links to invert_genus_1 → invert_group_of_interest_1 (sea urchins, 5500/ha).
+    # 1m wide × 50m long = 50 m²; need > 5500/ha → >27.5 individuals → 28 individuals.
```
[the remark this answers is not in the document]
saanobhaai: RUF003 is not in the enabled ruleset; these comments passed pre-commit unchanged.

saanobhaai:
[Inverts Phase 4: Submission & Validation](https://trello.com/c/c3xz0Lku/1962-inverts-phase-4-submission-validation)

saanobhaai:
@coderabbitai review

src/api/submission/validations/validators/sample_unit_consistency.py
```diff
@@ -181,6 +184,45 @@ def __call__(self, collect_record, **kwargs):
...
+        for su in queryset:
+            other_width_id = valid_id(su.width_id)
+            if other_width_id:
+                if str(other_width_id) != width_id:
+                    return (
+                        WARN,
+                        self.DIFFERENT_TRANSECT_WIDTH,
+                        {
+                            "width": width_id,
+                            "other_width": other_width_id,
```
gridcell: **Warning** · `correctness` · confidence: high

**other_width_id is a raw UUID in the context dict — JSON serialization will fail at runtime**

`width_id` is explicitly cast to `str` two lines before the loop, but `other_width_id = valid_id(su.width_id)` is placed into the context dict as a raw UUID object. The context dict therefore holds `'width': str` and `'other_width': UUID`. When validation results are persisted to a JSONB column, Django's default JSON encoder will raise `TypeError: Object of type UUID is not JSON serializable`, causing an unhandled 500 or a silently swallowed exception depending on the error boundary. This affects every macroinvertebrate transect that shares a sample event with a transect of a different width.

> **Suggestion:** Stringify `other_width_id` before inserting it into the context:
```python
{
    "width": width_id,
    "other_width": str(other_width_id),
}
```

> **Why it matters:** A UUID in a JSON-serialized context dict will crash result persistence for any transect that triggers the WARN branch.

**Evidence:**
- `width_id = str(width_id)` — explicit cast present
- `other_width_id = valid_id(su.width_id)` placed directly into context dict without str()

src/api/ingest/choices.py
```diff
@@ -59,3 +67,28 @@ def score_choices():
...
+        choices.append((str(obj.pk), obj.name))
+    for obj in InvertOrder.objects.all():
+        choices.append((str(obj.pk), obj.name))
+    for obj in InvertFamily.objects.all():
+        choices.append((str(obj.pk), obj.name))
+    for obj in InvertGenus.objects.all():
+        choices.append((str(obj.pk), obj.name))
+    for obj in InvertSpecies.objects.select_related("genus"):
+        choices.append((str(obj.pk), f"{obj.genus.name} {obj.name}"))
+    return sorted(choices, key=lambda x: x[1])
```
gridcell, line 86: **Warning** · `performance` · confidence: high

**invert_attributes_choices fires 6 separate DB queries; 4 querysets have no order_by so results are non-deterministic**

The function issues six independent SELECT statements (InvertGroupOfInterest, InvertClass, InvertOrder, InvertFamily, InvertGenus, InvertSpecies) each time it is called. Four of those (Class, Order, Family, Genus) have no `order_by`, so each returns rows in heap-scan order, which can vary across autovacuum or CLUSTER operations. The final `sorted(..., key=lambda x: x[1])` sorts all six result sets combined, but within-rank ordering before that point is not guaranteed stable across calls. Even called once per request, six round-trips add latency.

> **Suggestion:** Add `order_by('name')` to each queryset (and `order_by('genus__name', 'name')` for species). Consider merging into a single polymorphic query against the base `InvertAttribute` table, or apply a short-TTL cache on the result for the duration of the request.

> **Why it matters:** Non-deterministic ordering makes the ingest template choices tab differ across deployments. Six sequential queries per call add DB load without proportional benefit.

**Evidence:**
- InvertClass, InvertOrder, InvertFamily, InvertGenus querysets have no order_by clause
- Six separate SELECT statements issued per function call

<sub>Found by: reviewer-backend-python, reviewer-database · CR-002</sub>

gridcell (commented):
## AI Code Review

**Assessment:** The macroinvertebrate belt-transect protocol is well-implemented: all dispatch registration sites are correctly wired, validators are cleanly organised, and the batched InvertSizeValidator query avoids the N+1 anti-pattern. The most important item to address before merge is a UUID serialization bug in DifferentInvertTransectWidthValidator that will crash result persistence at runtime. The destructive migration 0116 should be verified against production row counts before applying, and test coverage for the new writer and two record-level validators is missing.

| 0 | 8 | 4 | 4 | 5 |

**Files reviewed:** 24 / 24

### Cross-Cutting Concerns
- The new BeltInvertProtocolWriter submission path has no end-to-end test (CR-005), and the new DifferentInvertTransectWidthValidator with the UUID-serialization bug (CR-001) has no test (CR-006). A submission test would have surfaced CR-001 immediately. Adding the writer test should be the highest-priority follow-up.
- Migration 0116 and the choices endpoint removal are coupled — confirming the table is empty in production and confirming no live client reads invertharvesttypes (CR-003, CR-010) must be done together before deploy, not independently.
- sample_unit_consistency.py and sample_units.py are shared utilities used by all protocols. The UUID-not-stringified bug (CR-001) was introduced into the shared file; reviewing similar context-dict constructions in the other (non-invert) width validators in that file is advisable to catch any inherited defects.
- Mode B (PR URL) was used for this review — local .gitignore/.dockerignore filtering was skipped because the working tree is not the target repo. All 24 changed files were reviewed.

### Findings (cross-file / praise)

<details>
<summary><strong>CR-005</strong> &middot; Warning &middot; <code>test_gap</code> — BeltInvertProtocolWriter.write() has no test</summary>

The new submission writer introduces four cooperating methods: `get_or_create_invertbelt_transect`, `get_or_create_beltinvert`, `create_obsbeltinvert`, and `write()`. None of these are exercised by any test in the diff. Other protocols (fish belt, benthic LIT) have end-to-end submission tests that call the writer and assert resulting DB rows. The macroinvertebrate protocol is missing a parallel test.

> **Suggestion:** Add a submission test analogous to the existing fish belt one. Use `valid_belt_invert_collect_record`, call the submission endpoint or invoke `BeltInvertProtocolWriter(collect_record, context).write()` directly, then assert that `InvertBeltTransect`, `BeltInvert`, `ObsBeltInvert`, and `Observer` records were created with the expected field values.

> **Why it matters:** The writer integrates four new DB models and two new parser functions. Without an integration test, regressions in any of those layers — wrong field mapping, missing transect, missing observations — will go undetected.

Applies to: `src/api/submission/writer.py`, `src/api/tests/test_validations.py`

</details>

<details>
<summary><strong>CR-006</strong> &middot; Warning &middot; <code>test_gap</code> — UniqueInvertBeltTransectValidator and DifferentInvertTransectWidthValidator are untested</summary>

Two new record-level validators — `UniqueInvertBeltTransectValidator` (detects duplicate transect submissions against existing DB records) and `DifferentInvertTransectWidthValidator` (warns when belt widths differ across the same sample event) — have no test coverage. These validators require pre-existing DB rows to trigger their error or warning path, which the current test suite never sets up for macroinvertebrate.

> **Suggestion:** For `UniqueInvertBeltTransectValidator`: submit a valid collect record to create a BeltInvert row, then validate a second collect record with matching site/management/date/number/depth/width/observers and assert ERROR with code `duplicate_invert_belt_transect`. For `DifferentInvertTransectWidthValidator`: create an `InvertBeltTransect` with one width, then validate a collect record targeting the same sample event with a different width ID and assert WARN with code `different_transect_width_se`.

> **Why it matters:** The duplicate-transect check is a critical data-quality guard. Leaving it untested means a regression in the observer-profile filter could allow duplicate submissions silently. Note the DifferentInvertTransectWidth path also has the UUID serialization bug (CR-001) which a test would have surfaced.

Applies to: `src/api/submission/validations/validators/invert_belt_transect.py`, `src/api/submission/validations/validators/sample_unit_consistency.py`, `src/api/tests/test_validations.py`

</details>

<details>
<summary><strong>CR-010</strong> &middot; Question &middot; <code>api_contract</code> — Was invertharvesttypes ever used by any active API client?</summary>

The choices endpoint previously exposed `invertharvesttypes` to API consumers. Were any mobile or third-party clients relying on that key? If genuinely unused, the cleanup is fully safe. If clients are still reading the key, they will silently receive a response missing that field after deploy. Older client versions in the wild may break without a deprecation window.

Applies to: `src/api/resources/choices.py`

</details>

<details>
<summary><strong>CR-011</strong> &middot; Question &middot; <code>test_gap</code> — invert_group_of_interest_2 fixture is added but never referenced in any test</summary>

`macroinvertebrate.py` gains a new `invert_group_of_interest_2` fixture (Crown-of-thorns starfish / COTS), but no test in the diff uses it as a parameter. Is this intended for a follow-up test that was intentionally deferred, or did the test that needs it get accidentally dropped?

Applies to: `src/api/tests/fixtures/macroinvertebrate.py`

</details>

<details>
<summary><strong>CR-012</strong> &middot; Question &middot; <code>test_gap</code> — InvertSizeValidator boundary value (exactly at 1.5x max_length) is not tested</summary>

The validator warns when `size > max_length * 1.5`. The existing test uses `size=50`, well above the threshold. There is no test for `size == max_length * 1.5` (which should pass) or `size == max_length * 1.5 + epsilon` (which should warn). The exact boundary is a common location for off-by-one mistakes.

Applies to: `src/api/tests/test_validations.py`

</details>

<details>
<summary><strong>CR-015</strong> &middot; Info &middot; <code>maintainability</code> — Three collect record fixtures share nearly identical boilerplate</summary>

`valid_belt_invert_collect_record`, `valid_belt_invert_collect_record_goi`, and `valid_belt_invert_collect_record_family` are approximately 50 lines each and differ only in the `invert_attribute` fixture and observation list. The repeated `beltinvert_transect`/`sample_event`/`observers` construction could be extracted into a shared factory helper to reduce maintenance surface.

Applies to: `src/api/tests/fixtures/collect_records.py`

</details>

<details>
<summary><strong>CR-016</strong> &middot; Info &middot; <code>maintainability</code> — macroinvertebrate_file fixture defined in test module rather than conftest</summary>

All other protocol file fixtures (`fishbelt_file`, `benthicpit_file`, `bleaching_file`, etc.) live in a `conftest.py`, making them available project-wide. The new `macroinvertebrate_file` is defined inline in `test_ingest.py` and cannot be shared with any other test module that might need it in the future.

Applies to: `src/api/tests/test_ingest.py`

</details>

<details>
<summary><strong>CR-017</strong> &middot; Praise &middot; <code>performance</code> — InvertSizeValidator batches all InvertAttribute lookups into a single DB query — avoids N+1</summary>

`InvertSizeValidator.__call__` collects all unique `invert_attribute` IDs from the observation list in a set comprehension, then issues one `InvertAttribute.objects.select_related(...).filter(pk__in=...)` query for all of them. The resulting `max_length_lookup` dict is built once and consulted in O(1) per `check_invert_size` call.

Applies to: `src/api/submission/validations/validators/invert_obs.py`

<sub>Found by: reviewer-backend-python, reviewer-database</sub>

</details>

<details>
<summary><strong>CR-018</strong> &middot; Praise &middot; <code>maintainability</code> — All protocol dispatch tables correctly updated — no missed registration sites</summary>

The PR correctly wires `MACROINVERTEBRATE_PROTOCOL` into every dispatch point that existing protocols use: `ingest/utils.py` (`ingest()`), `submission/utils.py` (`get_writer()` and `_validate_collect_record()`), `submission/validations/utils.py` (`PROTOCOL_SAMPLE_UNIT_MAP` and `PROTOCOL_SAMPLE_EVENT_PATH`), `utils/sample_units.py` (`migrate_collect_record_sample_event()`), and `models/core.py` (`CollectRecord.sample_unit` property). These registration sites are easy to miss in a protocol addition and the author has covered all of them.

Applies to: `src/api/ingest/utils.py`, `src/api/submission/utils.py`, `src/api/submission/validations/utils.py`, `src/api/utils/sample_units.py`, `src/api/models/core.py`

</details>

<details>
<summary><strong>CR-019</strong> &middot; Praise &middot; <code>maintainability</code> — Well-structured validation module with clear separation of concerns</summary>

The `belt_invert.py` validation module is comprehensively organised with clear section comments (field-level, row-level, record-level) and appropriate validator composition. The structure mirrors established patterns from other protocols and will be easy to maintain and extend.

</details>

<details>
<summary><strong>CR-020</strong> &middot; Praise &middot; <code>maintainability</code> — Robust invert_obs validators with proper null-handling and polymorphic attribute resolution</summary>

`InvertSizeValidator` and `InvertSizeBinRequiredValidator` properly handle optional/nullable size fields and apply sensible defaults. The polymorphic attribute type resolution (species → genus → family → order → class) is cleanly implemented and handles cases where an attribute might not have a `max_length`, preventing false-positive size warnings for GOI and family observations.

</details>

<details>
<summary><strong>CR-021</strong> &middot; Praise &middot; <code>test_gap</code> — Attribute-rank coverage for InvertSizeValidator is precisely targeted</summary>

Adding separate fixtures and tests for species-level, GOI-level, and family-level observations is a precisely targeted set of edge cases. The `InvertSizeValidator` iterates a fixed priority list of subattr names to find `max_length` — testing that GOI and family observations do not produce false-positive size warnings directly exercises the two paths where the lookup returns `None` and the validator must silently pass.

Applies to: `src/api/tests/test_validations.py`, `src/api/tests/fixtures/collect_records.py`

</details>

### Agent Coverage

| reviewer-backend-python | backend | 21 | 8 | 0 |
| reviewer-python-quality | code_quality | 21 | 2 | 0 |
| reviewer-database | database | 3 | 7 | 0 |
| reviewer-tests | testing | 5 | 9 | 0 |
| reviewer-security | security | 24 | 1 | 0 |

src/api/migrations/0116_remove_invert_harvest_type.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("api", "0115_goi_schema_finalize"),
+    ]
+
+    operations = [
+        migrations.DeleteModel(
+            name="InvertHarvestType",
```
gridcell, line 11: **Warning** · `api_contract` · confidence: high

**DeleteModel migration is destructive — verify InvertHarvestType is empty in production before applying**

Django's `DeleteModel` drops the table and all of its rows. If `invert_harvest_type` has any rows in production at apply time, they are permanently lost. The migration is implicitly reversible via migration history replay, but a rollback requires both migration replay and a fresh model definition — and the model was removed in the same PR. Additionally, the choices endpoint previously exposed `invertharvesttypes` to API consumers; any mobile or third-party clients reading that key will silently receive a response missing the field after deploy.

> **Suggestion:** Before merging, confirm via SQL (`SELECT COUNT(*) FROM invert_harvest_type;`) on production that the table is empty. Also verify no active client reads the `invertharvesttypes` key from the choices endpoint, and that no older client versions in the wild depend on it.

> **Suggestion consequences:** If a rollback is ever needed and the reverse path is not exercised, re-applying migration 0115 will succeed but the table will not exist. Any code still referencing InvertHarvestType will fail.

> **Why it matters:** Destructive migrations applied without verification are a real source of incident risk. A confirmation step prevents silent data loss and API breakage for existing clients.

<sub>Found by: reviewer-backend-python, reviewer-database · CR-003</sub>

src/api/submission/validations/validators/invert_belt_transect.py
```diff
@@ -0,0 +1,103 @@
...
+
+        for profile in profiles:
+            queryset = queryset.filter(beltinvert_method__observers__profile_id=profile)
+
+        for result in queryset:
+            transect_methods = get_related_transect_methods(result)
+            duplicate_check = self._check_for_duplicate_transect_methods(transect_methods, protocol)
+            if duplicate_check != OK:
+                return duplicate_check
+        return OK
```
gridcell, line 103: **Warning** · `performance` · confidence: medium

**Bare select_related() in UniqueInvertBeltTransectValidator combined with potential N+1 from get_related_transect_methods in a loop**

`InvertBeltTransect.objects.select_related().filter(**qry)` uses `select_related()` with no arguments, which tells Django to follow every non-null FK relationship by depth limit, potentially producing wider SQL JOINs than this uniqueness check actually needs. Additionally, `get_related_transect_methods(result)` is called once per row in the queryset. If that utility accesses a reverse FK (e.g. `beltinvert_method_set`), each iteration issues at least one additional query, creating an N+1 pattern.

> **Suggestion:** Identify the exact FK paths accessed by `get_related_transect_methods` and specify them explicitly, e.g. `InvertBeltTransect.objects.select_related('beltinvert_method').filter(**qry)`. If the method accesses a reverse relation, use `prefetch_related` instead.

> **Suggestion consequences:** Narrowing `select_related` means any relationship path inside `get_related_transect_methods` not listed will generate an additional per-row query. Review the call chain before tightening.

> **Why it matters:** Unbounded `select_related()` is a known Django anti-pattern producing wide JOINs that can be slower than targeted selects. The potential N+1 makes validation time scale linearly with the queryset size.

<sub>Found by: reviewer-database · CR-004</sub>

src/api/tests/test_ingest.py
```diff
@@ -427,3 +428,72 @@ def test_benthicpqt_ingest(
...
+
+    assert sample_event.get("site") == str(site1.id)
+    assert sample_event.get("management") == str(management1.id)
+    assert str(sample_event.get("sample_date")) == "2021-04-10"
+
+    assert len(observations) == 6
+    assert observations[0].get("invert_attribute") == str(invert_species_1.id)
+    assert observations[0].get("count") == 3
+    assert observations[0].get("include") is True
+    assert observations[5].get("include") is False
```
gridcell, line 501: **Warning** · `test_gap` · confidence: high

**Ingest test validates only the first of two produced collect records**

The test asserts `len(new_records) == 2` but retrieves and inspects only `new_records[0]`. The second collect record (Site 2, rows 7-12 in the CSV, 6 observations) is not checked — its site assignment, observation count, and the correct grouping of CSV rows are entirely untested.

> **Suggestion:** Add assertions for `new_records[1]`: verify its `sample_event` site matches `str(site2.id)`, its `obs_belt_inverts` list has length 6, and `observations[5].get('include')` is `False` (the last row in Site 2 has `Include=False` in the CSV).

> **Why it matters:** The grouping logic that partitions 12 CSV rows into exactly 2 collect records is the most complex part of the ingest path. Verifying only one group provides no confidence that the second group was assembled correctly.

<sub>Found by: reviewer-tests · CR-007</sub>

src/api/tests/test_ingest.py
```diff
@@ -427,3 +428,72 @@ def test_benthicpqt_ingest(
     new_records[1].data["quadrat_transect"]["reef_slope"] = str(reef_slope1.id)
     new_records[1].data["quadrat_transect"]["relative_depth"] = str(relative_depth2.id)
     new_records[1].data["quadrat_transect"]["num_quadrats"] = 2
+
+
+@pytest.fixture
+def macroinvertebrate_file(db):
+    return open(os.path.join(csv_data_dir, "macroinvertebrate.csv"))
```
gridcell: **Warning** · `maintainability` · confidence: high

**macroinvertebrate_file fixture leaks an open file handle**

The fixture opens the CSV with a bare `open()` call and returns the file object without any teardown. If the test raises before the file is consumed, the handle is never closed.

> **Suggestion:** Use `yield` with explicit teardown:
```python
@pytest.fixture
def macroinvertebrate_file(db):
    f = open(os.path.join(csv_data_dir, "macroinvertebrate.csv"))
    yield f
    f.close()
```

> **Why it matters:** Leaked file handles can mask test failures and accumulate during long test runs.

<sub>Found by: reviewer-tests · CR-008</sub>

src/api/submission/validations/validators/invert_obs.py
```diff
@@ -0,0 +1,168 @@
...
+        context = {"observation_id": obs.get("id")}
+        size = self.get_value(obs, self.size_path)
+        if size is not None and size != "" and not size_bin:
+            return ERROR, self.SIZE_BIN_REQUIRED, context
+        return OK
+
+    def __call__(self, collect_record, **kwargs):
+        observations = self.get_value(collect_record, self.observations_path) or []
+        size_bin = self.get_value(collect_record, self.size_bin_path)
+        return [self.check_size_bin(obs, size_bin) for obs in observations]
```
gridcell, line 100: **Question** · `correctness` · confidence: medium

**InvertSizeBinRequiredValidator raises one error per observation row — intended UX vs. single field-level error?**

`check_size_bin` returns `ERROR` for every observation that has a non-null `size` when `size_bin` is absent. A transect with 20 size-bearing observations will surface 20 identical errors. Other transect-level constraints in this codebase are configured as single `FIELD_LEVEL` checks that produce one error on the transect field. The intent here may be deliberate — to highlight each offending row — but it differs from the pattern used elsewhere. Please confirm this is the intended error presentation.

<sub>Found by: reviewer-backend-python · CR-009</sub>

src/api/ingest/macroinvertebrate_serializer.py
```diff
@@ -0,0 +1,166 @@
...
+
+    data__sample_event__site = serializers.CharField(
+        label="Site",
+        help_text="A unique name of a site where data was collected. Every site must be defined before ingestion and set up in the project in the web app.",
```
gridcell, line 30: **Info** · `security` · confidence: low

**Pre-existing: CSV ingest lacks explicit row-count and payload-size guardrails (cross-protocol)**

`MacroInvertebrateCSVSerializer` inherits from `CollectRecordCSVSerializer` which — consistent with all peer protocols — does not enforce an explicit maximum number of CSV rows, total payload size, or per-cell length cap. An authenticated user could upload a very large CSV to trigger memory/CPU pressure during ingest. This is a pre-existing characteristic of the framework and not a regression introduced by this PR.

**References:** CWE-400 Uncontrolled Resource Consumption, CWE-770 Allocation of Resources Without Limits or Throttling, OWASP API4:2023 Unrestricted Resource Consumption

<sub>Found by: reviewer-security · CR-013</sub>

src/api/submission/validations/validators/invert_obs.py
```diff
@@ -0,0 +1,168 @@
...
+        super().__init__(**kwargs)
+
+    @validator_result
+    def __call__(self, collect_record, **kwargs):
+        observations = self.get_value(collect_record, self.observations_path) or []
+        if not observations:
+            return OK
+        if all(not obs.get("include", True) for obs in observations):
+            return WARN, self.ALL_EXCLUDED
+        return OK
```
gridcell: **Info** · `maintainability` · confidence: medium

**AllObsExcludedValidator exported into shared validators namespace without an Invert* prefix**

Every other class in `invert_obs.py` is prefixed `Invert*` (`InvertCountValidator`, `InvertObsCountHighValidator`, `InvertSizeBinRequiredValidator`, `InvertSizeValidator`). `AllObsExcludedValidator` breaks this convention and is exported through the shared `validators/__init__.py`. If a future protocol adds its own all-observations-excluded check under the same name, the import in `__init__.py` will silently shadow it.

<sub>Found by: reviewer-backend-python · CR-014</sub>

saanobhaai:
@coderabbitai review

PR #690, opened 2026-05-22
[M1978] GFCR changes phase 2

saanobhaai:
## Summary by CodeRabbit

* **New Features**
  * Added finance-solution type, geographical coverage, TAF name, and solutions-count fields plus related choice lists; report export columns updated to include these.
  * Added a GFCR data audit command and a Phase simulation script (pre-deploy checks).

* **Bug Fixes**
  * Backfilled missing sustainable finance mechanism values to empty lists during migration.

* **Chores**
  * Moved finance-solution data into a dedicated FacilitiesSolutions export sheet.

* **Tests**
  * Added serializer, validation, choice-list, and report-export tests.


saanobhaai:
[GFCR API Phase 2 — new fields, validation, choices, report](https://trello.com/c/W77WCHkn/1978-gfcr-api-phase-2-new-fields-validation-choices-report)

saanobhaai:
@coderabbitai review

saanobhaai:
@coderabbitai review

src/tools/management/commands/check_gfcr_data.py
```diff
@@ -0,0 +1,470 @@
...
+                raise argparse.ArgumentTypeError(f"Invalid UUID: {value!r}")
+
+        parser.add_argument(
+            "--project-id", type=_uuid, help="Scope output to a single project UUID"
+        )
+        parser.add_argument("--output", type=str, help="Override default output file path")
+
+    def handle(self, *args, **options):
+        self.stdout.write("Restoring production database...")
+        call_command("dbrestore", "prod")
```
gridcell, line 65: **Warning** · `security` · confidence: high

**check_gfcr_data unconditionally restores the production DB locally with no confirmation guard**

`handle()` begins by calling `call_command("dbrestore", "prod")` as its very first action — no `--confirm` flag, no environment check, no prompt. `dbrestore` then DROPs and recreates the configured database and downloads the production backup into it. Two risks: (1) destructive — an accidental invocation drops local DB state with no opt-out before any audit logic runs; (2) data handling — it pulls a full copy of production data onto the local machine every run, unscoped even when `--project-id` is given.

> **Suggestion:** Gate the restore behind an explicit flag (e.g. require `--restore`/`--confirm`), or prompt before dropping the DB. At minimum print a clear warning and skip restore when data is already present. Document that this command downloads production data locally.

> **Why it matters:** An unguarded destructive DROP DATABASE plus an automatic full-production-data pull is easy to trigger by mistake — both data-loss and sensitive-data-handling risk.

<sub>Found by: reviewer-security · CR-001</sub>

src/tools/management/commands/check_gfcr_data.py
```diff
@@ -0,0 +1,470 @@
...
+                lines.append("---")
+                lines.append("")
+
+        lines.append("*Run `python manage.py check_gfcr_data` to regenerate.*")
+        lines.append("")
+
+        content = "\n".join(lines)
+
+        output_path = options.get("output") or f"gfcr_audit_{today_str}.md"
+        with open(output_path, "w") as f:
```
gridcell, line 215: **Warning** · `security` · confidence: medium

**--output path is written without validation, allowing arbitrary path writes**

`options.get("output")` is passed straight into `open(output_path, "w")` with no normalization or directory restriction. A value like `--output ../../etc/cron.d/x` or an absolute path writes/overwrites anywhere the process user can write. Practical exploitability is low (operator-run command) but undesirable, especially combined with the automatic prod restore.

> **Suggestion:** Resolve and constrain the output path to a known reports directory, or reject absolute paths and `..` components and refuse to overwrite outside an allowed directory.

> **Why it matters:** Unvalidated operator-controlled file path can clobber unrelated files or land in sensitive locations in elevated or CI contexts.

<sub>Found by: reviewer-security · CR-002</sub>

src/api/resources/gfcr.py
```diff
@@ -131,6 +131,59 @@ class Meta:
...
+            data["local_enterprise"] = False
+
+        # gender_smart: only Business and Financial mechanism.
+        if type_val not in ("business", "financial_mechanism"):
+            data["gender_smart"] = False
+
+        # number_of_solutions_supported_by: TAF, CTF, and Financial facility — must be > 0.
+        if type_val not in ("taf", "ctf", "financial_facility"):
+            data["number_of_solutions_supported_by"] = 0
+        elif data.get("number_of_solutions_supported_by", 0) == 0:
```
gridcell: **Warning** · `correctness` · confidence: high

**Explicit JSON null for number_of_solutions_supported_by bypasses the >0 guard and causes a DB IntegrityError**

When a client sends `"number_of_solutions_supported_by": null`, `data.get("number_of_solutions_supported_by", 0)` returns `None` (the key is present, so the default is not used). `None == 0` is `False`, so the error branch is skipped and `None` is passed to a `PositiveIntegerField(default=0)`, raising a DB IntegrityError (500) at save time instead of a clean 400.

> **Suggestion:** Use `(data.get("number_of_solutions_supported_by") or 0) == 0`, or coerce `None` to `0` before the elif.

> **Why it matters:** An unhandled IntegrityError surfaces as a 500 instead of a 400, leaking implementation details and breaking API-contract consistency with the other cross-field checks.

src/api/reports/gfcr.py
```diff
@@ -167,13 +167,17 @@ def f7_data(indicator_sets):
 
 def common_finance_solutions_columns(finance_solution):
     return [
+        finance_solution.get_fs_type_display(),
         finance_solution.name,
         ",".join(finance_solution.get_sustainable_finance_mechanisms_display()),
         finance_solution.get_sector_display(),
+        finance_solution.geographical_coverage,
```
gridcell: **Warning** · `correctness` · confidence: high

**geographical_coverage written as raw stored code in report export, inconsistent with display-value siblings**

`common_finance_solutions_columns` writes `finance_solution.geographical_coverage` (raw code e.g. `"national"`) while adjacent columns use `get_fs_type_display()` and `get_sector_display()`. The report shows `"national"` next to `"Conservation trust fund (CTF)"`. The test asserts `row[8] == "national"`, confirming the raw value ships.

> **Suggestion:** Use `finance_solution.get_geographical_coverage_display()` (auto-generated since `GEOGRAPHICAL_COVERAGE_CHOICES` is defined) and update the test assertion to the human-readable label (e.g. `"National"`).

> **Why it matters:** Report consumers expecting human-readable labels (as for all other choice fields in the same function) receive machine codes for this one column.

<sub>Found by: reviewer-backend-python · CR-004</sub>

src/api/migrations/0116_gfcr_phase2.py
```diff
@@ -0,0 +1,134 @@
+from django.db import migrations, models
+
+
+def convert_sfm_nulls(apps, schema_editor):
+    GFCRFinanceSolution = apps.get_model("api", "GFCRFinanceSolution")
+    GFCRFinanceSolution.objects.filter(sustainable_finance_mechanisms__isnull=True).update(
+        sustainable_finance_mechanisms=[]
```
gridcell, line 7: **Warning** · `correctness` · confidence: high

**Backfill converts NULLs to [] but field stays nullable, allowing new NULLs to be written**

`convert_sfm_nulls` converts existing NULL `sustainable_finance_mechanisms` to `[]`. The model field retains `null=True` (with `default=list`). Code paths that explicitly pass `None` can still write NULL, undoing the invariant. `validate()` coerces SFM=`[]` for non-`financial_mechanism` types but is skipped when `fs_type` is None.

> **Suggestion:** If the invariant is non-NULL, follow the data migration with an `AlterField` dropping `null=True`. If NULL must remain until a later phase, document it explicitly and ensure all write paths default to `[]`.

> **Why it matters:** New NULLs make `get_sustainable_finance_mechanisms_display()` and report iteration raise TypeError or produce blank output.

<sub>Found by: reviewer-database · CR-006</sub>

src/api/migrations/0116_gfcr_phase2.py
```diff
@@ -0,0 +1,134 @@
...
+    )
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("api", "0115_goi_schema_finalize"),
+    ]
+
+    operations = [
+        migrations.RunPython(convert_sfm_nulls, migrations.RunPython.noop),
```
gridcell, line 17: **Warning** · `maintainability` · confidence: high

**noop reverse for convert_sfm_nulls makes rollback silently incomplete**

The reverse function is `RunPython.noop`, so a downgrade will not restore original NULLs. Rolled-back rows keep `[]` rather than NULL, masking 'never set' semantics from pre-migration data and leaving data in a state older app versions do not expect.

> **Suggestion:** Add a real reverse function that sets `[]` back to NULL, or if reversibility is intentionally not required, add an inline comment documenting that decision.

> **Why it matters:** A noop reverse is a one-way data migration disguised as reversible; a rollback leaves data differing from what older app versions expect.

<sub>Found by: reviewer-database · CR-007</sub>

src/api/tests/test_gfcr.py
```diff
@@ -359,24 +360,287 @@ def test_notes_in_report_export(
...
+    assert "FacilitiesSolutions" in [ws.title for ws in wb.worksheets]
+    assert "BusinessesFinanceSolutions" not in [ws.title for ws in wb.worksheets]
+
+    fs_sheet = wb["FacilitiesSolutions"]
+    row = list(fs_sheet.iter_rows(min_row=2, max_row=2, values_only=True))[0]
+    # type is at index 4, geographical_coverage at index 8
+    assert row[4] == "Conservation trust fund (CTF)", f"Expected CTF display, got: {row}"
+    assert row[8] == "national", f"Expected 'national', got: {row}"
 
     os.remove(report_path)
```
gridcell, line 660: **Warning** · `error_handling` · confidence: high

**Report file leaks to disk when an assertion fails before os.remove**

`test_report_new_columns_and_sheet_name` (and the pre-existing `test_notes_in_report_export`) call `create_report()`, write a real file, assert, then `os.remove()` at the end. If an assertion fails, `os.remove()` is never reached and the xlsx leaks. Repeated CI failures accumulate stale files.

> **Suggestion:** Wrap file creation and assertions in `try/finally`, or use a pytest `tmp_path` fixture so cleanup always runs regardless of assertion outcome.

> **Why it matters:** Leaked files accumulate on CI agents and stale files make failures harder to diagnose.

<sub>Found by: reviewer-tests · CR-008</sub>

src/api/tests/test_gfcr.py
```diff
@@ -359,24 +360,287 @@ def test_notes_in_report_export(
...
+
+    report_path = create_report([project1.id])
+    assert report_path is not None
+
+    wb = load_workbook(report_path)
+    assert "FacilitiesSolutions" in [ws.title for ws in wb.worksheets]
+    assert "BusinessesFinanceSolutions" not in [ws.title for ws in wb.worksheets]
+
+    fs_sheet = wb["FacilitiesSolutions"]
+    row = list(fs_sheet.iter_rows(min_row=2, max_row=2, values_only=True))[0]
```
gridcell: **Warning** · `maintainability` · confidence: high

**list(...)[0] used in new test while sibling code was upgraded to next() in this same PR**

`test_notes_in_report_export` was updated in this PR to use `next(sheet.iter_rows(...))` instead of `list(...)[0]`, but the newly-added `test_report_new_columns_and_sheet_name` still uses `list(fs_sheet.iter_rows(...))[0]` — inconsistent with the style fix applied in the same PR.

> **Suggestion:** Replace with `row = next(fs_sheet.iter_rows(min_row=2, max_row=2, values_only=True))`.

> **Why it matters:** Inconsistent style within one PR; `list()[0]` raises a less informative IndexError than `next(..., default)` on an empty sheet.

<sub>Found by: reviewer-tests · CR-009</sub>

gridcell (commented):
## AI Code Review

**Assessment:** A well-structured, well-tested phased migration PR introducing GFCR Phase 2 schema changes, new cross-field serializer validation, a simulation rehearsal script, and a Phase 4 readiness audit command. The phased-rollout tooling is commendable and no blocking issues were found. Actionable polish items: an explicit JSON `null` bypasses the `number_of_solutions_supported_by` guard (500 instead of 400); the audit command unconditionally restores the production DB without a guard; `geographical_coverage` ships as a raw code in the report; the SFM field stays nullable after backfill with a noop reverse; duplicated business-rule constants; and a few test-coverage gaps.

| 0 | 10 | 2 | 7 | 4 |

**Files reviewed:** 11 / 11

> Note: the `gfcr.xlsx` template was independently verified against the PR head commit — it correctly contains the `FacilitiesSolutions` sheet (16 header columns) and updated `Investments`/`Revenues` sheets (15 columns each). Three earlier draft findings claiming the template was stale were false positives (wrong-branch inspection) and have been excluded.

### Cross-Cutting Concerns
- The Phase 2→3→4 rollout depends on out-of-band manual data fixes coordinated with the simulate and audit scripts. Two gaps remain writable in the Phase 2 interim: the `fs_type=None` validation bypass (CR-011) silently skips all cross-field invariants on full PUTs, and the deprecated `fm_*` sector choices (CR-012) remain valid API inputs. Both are intentional scaffolding for Phase 3/4 but should be tracked as explicit tickets with removal criteria.
- The five business-rule constants shared between `simulate_gfcr_phase3.py` and `check_gfcr_data.py` are duplicated verbatim (CR-005). Because the simulation is the rehearsal and the audit is the Phase 4 gate, silent divergence between the two copies is a direct deployment risk.
- The `check_gfcr_data` audit command conflates a read-only data-quality check with a destructive unconditional production DB restore (CR-001). Separating the restore (behind a flag or a separate command) would make the audit safe to invoke repeatedly and in automated contexts.

### General Findings

<details>
<summary><strong>CR-005</strong> &middot; Warning &middot; <code>maintainability</code> — Five module-level constants duplicated between simulation and audit scripts</summary>

`_REMOVED_SECTOR_VALUES`, `_REMOVED_SFM_VALUES`, `_REPORT_TITLES`, `_TARGET_TITLES`, and `_ALL_VALID_TITLES` are defined identically in both `simulate_gfcr_phase3.py` and `check_gfcr_data.py`. These encode Phase 3/4 business rules; editing one copy and not the other makes the simulation and audit silently disagree.

> **Suggestion:** Extract to a single shared module and import in both files for one source of truth.

> **Why it matters:** Silent divergence creates deployment risk: the simulation may mark Phase 4 safe while the audit disagrees, or vice versa.

Applies to: `src/api/scripts/simulate_gfcr_phase3.py`, `src/tools/management/commands/check_gfcr_data.py`

</details>
<details>
<summary><strong>CR-010</strong> &middot; Warning &middot; <code>test_gap</code> — Coercion parametrize cases missing for financial_facility, programmatic_co_financing, and the financial_mechanism SFM-retained path</summary>

`test_finance_solution_validate_coercion` covers `taf` and `business` well but never exercises `financial_facility` or `programmatic_co_financing` as the primary `fs_type`, and there is no positive test that `sustainable_finance_mechanisms` is RETAINED for `financial_mechanism` (the only type that permits it). PCF clears the broadest set of fields including SFM yet no parametrize case targets it.

> **Suggestion:** Add parametrize cases: `financial_facility` (used_an_incubator coerced to None), `programmatic_co_financing` (SFM coerced to []), and a dedicated test that `financial_mechanism` with SFM=['blue_bonds'] returns 200 with SFM intact.

> **Why it matters:** A regression clearing SFM for financial_mechanism, or failing to coerce PCF fields, would not be caught by the existing test suite.

Applies to: `src/api/tests/test_gfcr.py`, `src/api/resources/gfcr.py`

</details>
<details>
<summary><strong>CR-015</strong> &middot; Info &middot; <code>maintainability</code> — Check IDs and category codes scattered as bare string literals in audit command</summary>

Check IDs (`FS-1`..`FS-16`, `IS-1`, `IS-2`, `INV-1`) and category keys (`is_titles`, `fs_type`, `fs_cross`, `fs_pcf`, `fs_sector`, `fs_sfm`, `inv_type`) are bare string literals across yield statements and the summary table, creating a risk of typo drift that silently miscounts categories.

> **Suggestion:** Define check IDs and category keys as an Enum or module-level constants and reference them in both the yields and the counts table.

> **Why it matters:** A typo between a yield and the counts table silently drops a category from the totals.

Applies to: `src/tools/management/commands/check_gfcr_data.py`

</details>
<details>
<summary><strong>CR-017</strong> &middot; Info &middot; <code>correctness</code> — Merge migration 0117 is correct — both 0116_* migrations present and depend on 0115</summary>

`0116_remove_invert_harvest_type` and `0116_gfcr_phase2` both depend on `0115_goi_schema_finalize`; `0117_merge` lists both as dependencies with no operations. The migration graph is valid.

> **Why it matters:** Confirms the parallel-0116 branch is resolved correctly.

Applies to: `src/api/migrations/0117_merge_20260530_1302.py`

</details>
<details>
<summary><strong>CR-019</strong> &middot; Info &middot; <code>maintainability</code> — xl.get_worksheet performs case-insensitive lookup with create=True — sheet name mismatches silently create duplicates</summary>

`get_worksheet` lowercases both the requested and template sheet titles before comparing and falls back to creating a new sheet when not found. A typo in a `sheet_data` key would silently create a duplicate sheet rather than raising an error — a footgun for future template or sheet-name changes. (Note: `xl.py` is not modified in this PR; raised as context for the sheet rename.)

> **Why it matters:** Future sheet renames could silently produce ghost sheets instead of failing fast.

Applies to: `src/api/reports/xl.py`

<sub>Found by: reviewer-general-purpose</sub>
</details>

### Praise
- **CR-020** — Serializer validate() uses safe field coercion with no injection surface (src/api/resources/gfcr.py) <sub>(reviewer-security)</sub>
- **CR-021** — Thorough server-side cross-field enforcement with coercion-before-error pattern (src/api/resources/gfcr.py) <sub>(reviewer-backend-python)</sub>
- **CR-022** — Phased migration strategy with simulation script and audit command is well-structured <sub>(reviewer-database)</sub>
- **CR-023** — Shared _fs_put_payload helper and _FS_BASE constant eliminate repetition across parametrized tests <sub>(reviewer-tests)</sub>

### Agent Coverage
| reviewer-security | security | 11 | 5 | 0 |
| reviewer-backend-python | backend | 6 | 4 | 0 |
| reviewer-python-quality | quality | 10 | 2 | 0 |
| reviewer-database | database | 2 | 6 | 0 |
| reviewer-tests | testing | 2 | 6 | 0 |
| reviewer-general-purpose | general | 1 | 1 | 0 |

src/api/resources/gfcr.py
```diff
@@ -131,6 +131,59 @@ class Meta:
         model = GFCRFinanceSolution
         exclude = []
 
+    def validate(self, data):
+        type_val = data.get("fs_type")
+        if type_val is None:
```
gridcell, line 136: **Question** · `maintainability` · confidence: medium

**Early return when fs_type is None silently bypasses all cross-field invariants on a full PUT — intentional and time-bounded?**

`data.get("fs_type")` is None both when the key is omitted and when `null` is sent; `validate()` then returns immediately, skipping every cross-field check. Because `update()` is a full PUT (no `partial=True`), a client omitting `fs_type` can write any combination of guarded fields with no validation, leaving records Phase 4 will later reject. The inline comment about Phase 4 deferral suggests this is deliberate scaffolding, but scope and exit criteria are not captured in code.

> **Suggestion:** Confirm: (a) Phase 3 will backfill fs_type on all records, (b) Phase 4 removes the bypass or makes fs_type required, and (c) this is tracked in a ticket.

> **Why it matters:** Records can drift into states Phase 4 enforcement rejects, with no signal until the Phase 4 audit runs.

<sub>Found by: reviewer-backend-python · CR-011</sub>

src/api/migrations/0116_gfcr_phase2.py
```diff
@@ -0,0 +1,134 @@
...
+                    ("fm_biodiversity_credits", "Financial Mechanisms - Biodiversity Credits"),
+                    ("fm_blue_carbon_credits", "Financial Mechanisms - Blue Carbon Credits"),
+                    (
+                        "fm_conservation_trust_fund",
+                        "Financial Mechanisms - Conservation Trust Fund",
+                    ),
+                    ("fm_insurance_mechanisms", "Financial Mechanisms - Insurance Mechanisms"),
+                    ("fm_mpa_user_fee", "Financial Mechanisms - MPA User Fee"),
+                    ("fm_resilience_credits", "Financial Mechanisms - Resilience Credits"),
+                    ("fm_other", "Financial Mechanisms - Other"),
```
gridcell, line 46: **Question** · `correctness` · confidence: medium

**sector AlterField retains deprecated fm_* choices — intentional for Phase 2?**

The `sector` `AlterField` (and the model `SECTOR_CHOICES`) preserves the seven `fm_*` choices that `simulate_gfcr_phase3.py` and `check_gfcr_data.py` treat as 'removed in Phase 4'. The API will keep accepting them as valid until Phase 4.

> **Suggestion:** Confirm whether there is a risk of new data being written with deprecated `fm_*` values between Phase 2 and Phase 4, and whether that is acceptable.

> **Why it matters:** New records written with soon-to-be-removed values increase the Phase 4 data-cleanup burden.

<sub>Found by: reviewer-database · CR-012</sub>

src/api/scripts/simulate_gfcr_phase3.py
```diff
@@ -0,0 +1,264 @@
...
+        iset.save(update_fields=["title"])
+        log[iset.project.name].append(f"  IS title: {title!r} → {new_title!r}")
+
+
+# ---------------------------------------------------------------------------
+# Step 2 — finance solution types and field coercion
+# ---------------------------------------------------------------------------
+
+
+def _assign_types(log, errors):
```
gridcell, line 107: **Info** · `security` · confidence: high

**simulate_gfcr_phase3 runscript mutates DB and bypasses serializer auth/permissions by design**

The runscript instantiates `GFCRFinanceSolutionSerializer()` directly and writes via `fs.save()`/`qs.update()`/`qs.delete()` outside any request context, intentionally bypassing DRF auth and `ProjectDataAdminPermission` and performing unscoped irreversible mutations including deleting all PCF revenues. Acceptable for an operator-only 'local DB only' runscript, but this must never be wired into any request-handling path.

> **Suggestion:** Keep this script out of any HTTP/route-reachable code path; the 'local DB only' docstring is good — consider an explicit environment guard.

> **Why it matters:** Direct serializer use bypasses the authorization layer the API viewset enforces.

<sub>Found by: reviewer-security · CR-013</sub>

src/tools/management/commands/check_gfcr_data.py
```diff
@@ -0,0 +1,470 @@
...
+)
+
+_REPORT_TITLES = frozenset({"Baseline", "Mid-year report", "End-year report"})
+_TARGET_TITLES = frozenset({"Phase 1 target", "Mid-term target", "Final target"})
+_ALL_VALID_TITLES = _REPORT_TITLES | _TARGET_TITLES
+
+
+def _md_table(headers, rows):
+    def _cell(v):
+        return str(v).replace("|", "&#124;").replace("\n", "<br>")
```
gridcell, line 39: **Info** · `security` · confidence: high

**Markdown table cells are escaped, preventing markdown/HTML injection from DB-sourced names**

`_md_table` escapes pipe and newlines in every cell, correctly preventing user-controlled DB values (project and finance-solution names) from breaking out of table cells. Positive observation noted for traceability.

> **Why it matters:** Confirms the report-generation path is not injectable via stored names.

<sub>Found by: reviewer-security · CR-014</sub>

src/api/migrations/0116_gfcr_phase2.py
```diff
@@ -0,0 +1,134 @@
...
+class Migration(migrations.Migration):
+    dependencies = [
+        ("api", "0115_goi_schema_finalize"),
+    ]
+
+    operations = [
+        migrations.RunPython(convert_sfm_nulls, migrations.RunPython.noop),
+        migrations.AlterField(
+            model_name="gfcrfinancesolution",
+            name="sector",
```
gridcell, line 20: **Info** · `performance` · confidence: high

**AlterField on sector and AddField operations are metadata-only in PostgreSQL 11+ — no table lock concern**

`AlterField` on `sector` (blank/default) is ORM-only; `AddField` with non-volatile defaults (`geographical_coverage` '', `taf_name` '', `number_of_solutions_supported_by` 0) are catalog-only in PG11+. Only a brief AccessExclusiveLock is taken; the table is not large enough to warrant zero-downtime migration tooling here.

> **Why it matters:** Confirms the migration will not cause a long table-rewrite lock.

<sub>Found by: reviewer-database · CR-016</sub>

src/api/tests/test_gfcr.py
```diff
@@ -359,24 +360,287 @@ def test_notes_in_report_export(
...
+        # geographical_coverage is required for CTF — no value to coerce to
+        ({"fs_type": "ctf", "geographical_coverage": ""}, "geographical_coverage"),
+        # number_of_solutions_supported_by must be > 0 for TAF — no value to coerce to
+        (
+            {"fs_type": "taf", "number_of_solutions_supported_by": 0},
+            "number_of_solutions_supported_by",
+        ),
+    ],
+)
+def test_finance_solution_validate_errors(
```
gridcell, line 503: **Info** · `test_gap` · confidence: medium

**validate_errors missing a ctf/financial_facility case for the number_of_solutions_supported_by requirement**

`test_finance_solution_validate_errors` covers `business` (sector), `ctf` (geographical_coverage), and `taf` (number_of_solutions_supported_by==0). `ctf` and `financial_facility` also require `number_of_solutions_supported_by`>0 but only the `taf` path is tested for that specific error.

> **Suggestion:** Add a ctf and/or financial_facility case with number_of_solutions_supported_by=0 to the error parametrize list.

> **Why it matters:** Tightens confirmation that the >0 requirement applies to all three types.

<sub>Found by: reviewer-tests · CR-018</sub>

src/api/migrations/0116_gfcr_phase2.py
saanobhaai: null=True on SFM is intentional per spec; the migration backfill is a one-time cleanup, not an invariant change.

src/api/resources/gfcr.py
saanobhaai: Phase 2 gate is deliberate and spec-documented; removed in Phase 4 ticket.

src/api/migrations/0116_gfcr_phase2.py
saanobhaai: informational only

alanjleonard (approved):

PR #697, opened 2026-06-03
propose new invert species support

saanobhaai:
## Summary by CodeRabbit

* **New Features**
  * Added a new API endpoint for managing invertebrate species with listing, search, and creation support.

* **Enhancements**
  * Admin UI gains a status filter for invertebrate species.
  * API returns display names, status, and length metadata; supports filtering by genus, family, group, and status.

* **Database**
  * Installed revision-tracking for invert attributes and backfilled revision timestamps.

* **Tests**
  * Added test validating creation of a proposed invertebrate species via the API.


gridcell (commented):
## AI Code Review

**Assessment:** The new InvertSpecies endpoint is structurally consistent with the existing InvertAttribute pattern — permissions, status-forcing, and queryset shape are all correct. However, five warnings across correctness, database, and test coverage warrant attention before merge: most critically, BaseAttributeApiViewSet.perform_create never sets created_by, breaking both the authenticated visibility filter (submitters cannot see their own pending records) and the audit trail (any user can attribute a record to another user). Supporting gaps include a missing UniqueTogetherValidator (duplicate POSTs return 500), an order_by on a FK id rather than name, over-fetching in select_related, and insufficient test coverage for negative paths and filter/search behavior.

| 0 | 9 | 2 | 3 | 3 |

**Files reviewed:** 4 / 4

### Cross-Cutting Concerns
- The root defect behind CR-001 (attribution spoofing + broken submitter visibility) lives in BaseAttributeApiViewSet.perform_create in src/api/resources/base.py, not in InvertSpeciesViewSet specifically. The same gap affects every attribute viewset that inherits from BaseAttributeApiViewSet (InvertAttributeViewSet, and likely fish and benthic equivalents). Fixing it once in the base class corrects all inheriting viewsets simultaneously.
- exclude=[] without explicit read_only_fields is a codebase-wide convention that consistently exposes created_by as a writable serializer field across all attribute serializers. A single base serializer with created_by and updated_by declared as read_only_fields would eliminate the attribution-spoofing surface across all endpoints.
- The test file (test_invert_attributes.py) covers only the happy-path POST. All negative paths — unauthenticated access, status immutability, filter/search correctness, and PUT/DELETE on approved records — are absent. This pattern of single happy-path test coverage is insufficient for an endpoint that enforces a business workflow (proposal/approval cycle).
- CR-003 (order_by on FK id) and CR-005 (over-fetching in select_related) suggest the queryset was copied from a sibling viewset without auditing the field names. A brief audit of the model's Meta.ordering and the serializer's actual field access would catch both issues before review.

### General & Cross-File Findings

<details>
<summary><strong>CR-001</strong> &middot; Warning &middot; <code>correctness</code> — perform_create never sets created_by, breaking submitter visibility and allowing attribution spoofing</summary>

BaseAttributeApiViewSet.perform_create calls serializer.save(status=APPROVAL_STATUSES[-1][0]) but never passes created_by. This has two concrete consequences. First, a correctness gap: get_queryset for authenticated users filters on Q(created_by=request.user.profile), so any InvertSpecies created via the API has created_by=NULL and is never returned to the submitting user until a superuser approves it — the submitter cannot see or manage their own pending submission. Second, a security gap: because InvertSpeciesSerializer uses exclude=[] without overriding created_by as read-only, a client can POST {"name": ..., "genus": ..., "created_by": "<any-profile-uuid>"} and attribute the record to another user. While this is limited to provenance spoofing (no privilege escalation, status is forced to PROPOSED), the audit trail for who proposed a species can be forged by any authenticated user. This is a pre-existing, codebase-wide defect shared with InvertAttributeSerializer and fish/benthic serializers.

> **Suggestion:** Fix BaseAttributeApiViewSet.perform_create to pass created_by=self.request.user.profile (and updated_by) when saving, so all attribute viewsets are covered in one place. As a secondary defence, also declare created_by read-only in the serializer (read_only_fields = ("created_by",)) so the field cannot be supplied by clients regardless of the base class behaviour.

> **Why it matters:** Without server-side assignment, the authenticated get_queryset visibility branch (Q(created_by=request.user.profile)) is functionally dead for all newly created records, and the provenance audit trail can be spoofed by any authenticated user.

Applies to: `src/api/resources/base.py`

<sub>Found by: reviewer-security, reviewer-backend-python</sub>
</details>

<details>
<summary><strong>CR-002</strong> &middot; Warning &middot; <code>error_handling</code> — UniqueConstraint(name, genus) has no DRF validator — duplicate POST returns 500</summary>

The DB enforces uniqueness of (name, genus) via a unique_invertspecies_name_genus constraint (visible in macroinvertebrate.py). InvertSpeciesSerializer declares no UniqueTogetherValidator, so a duplicate POST raises an IntegrityError that surfaces as an unhandled 500 rather than a structured 400 validation error. This leaks DB error details and breaks API clients expecting RFC 7807-style error responses.

> **Suggestion:** Add UniqueTogetherValidator(queryset=InvertSpecies.objects.all(), fields=["name", "genus"]) to InvertSpeciesSerializer.Meta.validators (import from rest_framework.validators).

> **Why it matters:** A 500 leaks DB error details to clients and breaks any consumer that expects a 400 with a structured error body on duplicate submissions.

Applies to: `src/api/resources/invert_attribute.py`, `src/api/models/protocols/macroinvertebrate.py`

</details>

<details>
<summary><strong>CR-006</strong> &middot; Warning &middot; <code>test_gap</code> — No tests for list, retrieve, or any advertised filter/search parameters</summary>

InvertSpeciesViewSet exposes list and retrieve actions plus filters (genus, genus__family, genus__group_of_interest, status) and search (name, genus__name, genus__family__name). None of these paths are tested. A typo in a filter field name or a misconfigured search backend would be silent, and regressions in the read paths would not be caught by CI.

> **Suggestion:** Add at minimum: one list test, one retrieve test, one filter test (e.g., filter by status), and one search test.

> **Why it matters:** The read path is the highest-traffic surface of the endpoint; regressions there would go undetected.

Applies to: `src/api/tests/test_invert_attributes.py`

</details>

### Praise
- **select_related fully covers the genus join needed by get_display_name, preventing N+1 on list** (`src/api/resources/invert_attribute.py`) — CR-015, found by reviewer-backend-python, reviewer-database
- **Three-layer assertion (status code, response body, DB state) is the correct test pattern** (`src/api/tests/test_invert_attributes.py`) — CR-016, found by reviewer-tests
- **status declared read-only at the serializer layer in addition to viewset enforcement — good defense-in-depth** (`src/api/resources/invert_attribute.py`) — CR-017, found by reviewer-security

### Agent Coverage
| reviewer-security | security | 4 | 4 | 0 |
| reviewer-backend-python | backend | 4 | 5 | 0 |
| reviewer-python-quality | code_quality | 4 | 1 | 0 |
| reviewer-tests | testing | 1 | 6 | 0 |
| reviewer-database | database | 2 | 4 | 0 |

src/api/resources/invert_attribute.py
```diff
@@ -107,3 +107,38 @@ class InvertAttributeViewSet(BaseAttributeApiViewSet):
...
+        model = InvertSpecies
+        fields = ["genus", "genus__family", "genus__group_of_interest", "status"]
+
+
+class InvertSpeciesViewSet(BaseAttributeApiViewSet):
+    serializer_class = InvertSpeciesSerializer
+    queryset = InvertSpecies.objects.select_related(
+        "genus__family__order__invert_class",
+        "genus__group_of_interest",
+    ).order_by("genus", "name")
```
gridcell: **Warning** · `correctness` · confidence: high

**order_by("genus", "name") sorts by genus FK integer id, not genus name**

Passing a FK field name ("genus") to order_by sorts by genus_id (the integer primary key), not by genus name. The result set appears in arbitrary internal-id order, which is inconsistent with the display_name format "Genus Species" and with InvertGenus.Meta.ordering=("name",). InvertSpecies.Meta.ordering=("genus","name") on the model has the same defect and will affect any queryset that does not override ordering.

> **Suggestion:** Use .order_by("genus__name", "name") in the viewset queryset.

> **Why it matters:** Clients receive species sorted by an opaque internal id rather than taxonomic name, risking confusing UI display and inconsistency with sibling endpoints.

<sub>Found by: reviewer-database · CR-003</sub>

src/api/resources/invert_attribute.py
```diff
@@ -107,3 +107,38 @@ class InvertAttributeViewSet(BaseAttributeApiViewSet):
...
+        required=False,
+        allow_null=True,
+    )
+
+    class Meta:
+        model = InvertSpecies
+        exclude = []
+
+    def get_display_name(self, obj):
+        return f"{obj.genus.name} {obj.name}"
```
gridcell, line 128: **Warning** · `correctness` · confidence: medium

**get_display_name accesses obj.genus.name without a guard against a missing or deferred genus**

get_display_name returns f"{obj.genus.name} {obj.name}" with no null/attribute guard. The default viewset queryset covers genus via select_related, so this is safe today. However, a ?fields= slimmed serializer, a nested instantiation, or any future override that omits select_related would either cause per-row N+1 queries or raise AttributeError if genus is None, surfacing as a 500 on the list endpoint. InvertAttribute.__str__ already handles this case safely and could be delegated to.

> **Suggestion:** Either delegate to str(obj) (which already handles a missing genus safely) or add an explicit guard: return f"{obj.genus.name} {obj.name}" if obj.genus else obj.name.

> **Why it matters:** A deferred or missing genus would surface as a 500 on the list endpoint rather than graceful degradation, and the fragility is non-obvious to future maintainers.

<sub>Found by: reviewer-backend-python · CR-004</sub>

src/api/resources/invert_attribute.py
```diff
@@ -107,3 +107,38 @@ class InvertAttributeViewSet(BaseAttributeApiViewSet):
...
+        fields = ["genus", "genus__family", "genus__group_of_interest", "status"]
+
+
+class InvertSpeciesViewSet(BaseAttributeApiViewSet):
+    serializer_class = InvertSpeciesSerializer
+    queryset = InvertSpecies.objects.select_related(
+        "genus__family__order__invert_class",
+        "genus__group_of_interest",
+    ).order_by("genus", "name")
+    filterset_class = InvertSpeciesFilterSet
```
gridcell: **Warning** · `performance` · confidence: high

**select_related fetches family/order/invert_class join chains the serializer never reads**

select_related("genus__family__order__invert_class", "genus__group_of_interest") emits LEFT OUTER JOINs for family, order, and invert_class on every list request. InvertSpeciesSerializer only accesses obj.genus.name and InvertSpecies scalar fields. genus__group_of_interest is also unused by the serializer (the filterset references it only by id). The deeper chain adds unnecessary joins and wider rows on every list call with no benefit to the current serializer.

> **Suggestion:** Trim to select_related("genus") and keep group_of_interest only if a downstream consumer traverses the related objects. Also update order_by to order_by("genus__name", "name") (see CR-003).

> **Why it matters:** Unnecessary joins increase DB data transfer and result-set parse time on every list call, with no benefit to the current serializer output.

<sub>Found by: reviewer-database · CR-005</sub>

src/api/tests/test_invert_attributes.py
```diff
@@ -219,6 +220,17 @@ def test_invertbelttransect_filter_size_bin(
...
 
+def test_post_invert_species_creates_proposed(db_setup, api_client1, invert_genus_1):
+    """POST /v1/invertspecies/ creates a proposed InvertSpecies owned by the authenticated user."""
+    payload = {"name": "newspecies", "genus": str(invert_genus_1.pk)}
+    response = api_client1.post("/v1/invertspecies/", data=payload, format="json")
+    assert response.status_code == 201
+    data = response.json()
+    assert data["status"] == PROPOSED
+    assert data["display_name"] == f"{invert_genus_1.name} newspecies"
+    assert InvertSpecies.objects.filter(name="newspecies", genus=invert_genus_1).exists()
```
gridcell, line 231: **Warning** · `test_gap` · confidence: high

**Unauthenticated POST is not tested**

A create action was added to the endpoint but there is no test verifying that an unauthenticated POST returns 401 or 403. Combined with the permission guards being inherited (not explicitly tested here), a missing or mis-configured permission class would silently allow anonymous data injection.

> **Suggestion:** Add a test using an unauthenticated client (no credentials) POSTing to the invert species endpoint and asserting a 401 or 403 response.

> **Why it matters:** A missing permission guard would allow anonymous users to inject unreviewed species records.

<sub>Found by: reviewer-tests, reviewer-security · CR-007</sub>

gridcell, line 231: **Warning** · `test_gap` · confidence: high

**No test verifies that a caller-supplied status is forced to PROPOSED**

The primary business invariant is that any POSTed species is stored with status=PROPOSED regardless of the submitted value. If a client POSTs status=APPROVED, no test confirms the server overrides it. Removal or regression of the status-override logic in perform_create would not be caught, potentially allowing self-approval of species records.

> **Suggestion:** Add a test that POSTs with status=APPROVED (or any non-PROPOSED value) and asserts that the stored record has status==PROPOSED.

> **Why it matters:** Removal of the status-override logic would allow authenticated users to self-approve species proposals, bypassing the review workflow entirely.

<sub>Found by: reviewer-tests, reviewer-security · CR-008</sub>

src/api/resources/invert_attribute.py
```diff
@@ -107,3 +107,38 @@ class InvertAttributeViewSet(BaseAttributeApiViewSet):
...
+        coerce_to_string=False,
+        required=False,
+        allow_null=True,
+    )
+
+    class Meta:
+        model = InvertSpecies
+        exclude = []
+
+    def get_display_name(self, obj):
```
gridcell, line 127: **Warning** · `code_quality` · confidence: medium

**get_display_name lacks type annotations for parameter and return type**

The method get_display_name has no type hints for either its parameter or its return value. The project convention and mypy compliance expect annotations of the form def get_display_name(self, obj: InvertSpecies) -> str:. This is the only annotation gap in the new code; all other methods and fields follow the convention.

> **Suggestion:** Add type annotations: def get_display_name(self, obj: InvertSpecies) -> str:

> **Why it matters:** Missing annotations reduce mypy coverage and make the method signature ambiguous to readers and static analysis tools.

<sub>Found by: reviewer-python-quality · CR-009</sub>

src/api/resources/invert_attribute.py
```diff
@@ -107,3 +107,38 @@ class InvertAttributeViewSet(BaseAttributeApiViewSet):
...
+        max_digits=6,
+        decimal_places=1,
+        coerce_to_string=False,
+        required=False,
+        allow_null=True,
+    )
+
+    class Meta:
+        model = InvertSpecies
+        exclude = []
```
gridcell, line 125: **Question** · `api_contract` · confidence: medium

**exclude=[] exposes all model fields including audit and FK columns — intentional?**

exclude=[] on InvertSpeciesSerializer.Meta serializes every model column: genus_id, status (read-only), created_by, updated_by, created_on, updated_on, and all max_length_* fields. The sibling InvertAttributeSerializer also uses exclude=[] but redeclares many fields as SerializerMethodFields to shape output. InvertSpeciesSerializer does not apply similar shaping. Is exposing raw audit timestamps, the created_by profile UUID, and the genus FK integer (alongside any other internal columns) the intended API contract for this endpoint?

<sub>Found by: reviewer-backend-python · CR-010</sub>

src/api/tests/test_invert_attributes.py
```diff
@@ -219,6 +220,17 @@ def test_invertbelttransect_filter_size_bin(
     assert resp.json()["count"] == 0
 
 
+def test_post_invert_species_creates_proposed(db_setup, api_client1, invert_genus_1):
+    """POST /v1/invertspecies/ creates a proposed InvertSpecies owned by the authenticated user."""
```
gridcell, line 224: **Question** · `docs` · confidence: high

**Test docstring claims ownership but no created_by assertion verifies it, and ownership may not be set**

The docstring for test_post_invert_species_creates_proposed states that the created species is 'owned by the authenticated user', but no assertion verifies the created_by field on the stored record. Given that BaseAttributeApiViewSet.perform_create does not currently set created_by (see CR-001), the ownership claim in the docstring is likely inaccurate. The test may be documenting intended rather than actual behaviour.

<sub>Found by: reviewer-tests · CR-011</sub>

src/api/resources/invert_attribute.py
```diff
@@ -107,3 +107,38 @@ class InvertAttributeViewSet(BaseAttributeApiViewSet):
...
+class InvertSpeciesSerializer(CreateOrUpdateSerializerMixin, BaseAPISerializer):
+    status = serializers.ReadOnlyField()
+    display_name = serializers.SerializerMethodField()
+    max_length = serializers.DecimalField(
+        max_digits=6,
+        decimal_places=1,
+        coerce_to_string=False,
+        required=False,
+        allow_null=True,
+    )
```
gridcell, line 121: **Info** · `correctness` · confidence: high

**DecimalField max_digits=6, decimal_places=1 matches the model column exactly**

InvertSpecies.max_length is DecimalField(max_digits=6, decimal_places=1, null=True, blank=True). The serializer field matches on max_digits, decimal_places, allow_null, and required=False. No precision or nullability mismatch.

<sub>Found by: reviewer-backend-python · CR-012</sub>

src/api/tests/test_invert_attributes.py
```diff
@@ -219,6 +220,17 @@ def test_invertbelttransect_filter_size_bin(
...
+    payload = {"name": "newspecies", "genus": str(invert_genus_1.pk)}
+    response = api_client1.post("/v1/invertspecies/", data=payload, format="json")
+    assert response.status_code == 201
+    data = response.json()
+    assert data["status"] == PROPOSED
+    assert data["display_name"] == f"{invert_genus_1.name} newspecies"
```
gridcell, line 230: **Info** · `test_gap` · confidence: medium

**display_name assertion relies on implicit fixture name format**

The display_name assertion in the test passes as long as the fixture's genus and species names happen to match the format assembled by get_display_name. If get_display_name were changed to use a different field (e.g., genus abbreviation), the assertion could still pass if the fixture value coincidentally matched. Making the expected value explicit (asserting the exact string built from known fixture values) would clarify intent and make the test more resilient.

<sub>Found by: reviewer-tests · CR-013</sub>

src/api/resources/invert_attribute.py
gridcell, line 140: **Info** · `security` · confidence: high

**Status mass-assignment and unauthenticated write are correctly prevented by the inherited permission and serializer pattern**

status cannot be mass-assigned: it is declared as ReadOnlyField in the serializer and perform_create forces the value to PROPOSED regardless of POST body content. Write access is gated by Or(UnauthenticatedReadOnlyPermission, AttributeAuthenticatedUserPermission): anonymous users cannot create; authenticated users can POST and GET; PUT and DELETE are restricted to records in PROPOSED status. GET is intentionally public, matching the sibling InvertAttributeViewSet. No auth-bypass or status-escalation path exists in the current implementation.

<sub>Found by: reviewer-security · CR-014</sub>

src/api/resources/sync/views.py
```diff
@@ -116,6 +119,12 @@
     INVERT_ATTRIBUTES_SOURCE_TYPE: {
         "view": invert_attribute.InvertAttributeViewSet,
         "required_filters": NO_FILTERS,
+        "read_only": True,
+        "visibility_filtered": True,
```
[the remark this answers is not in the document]
saanobhaai: The correct write path for proposed species is the invert_species source (lines 125–130), which uses InvertSpeciesViewSet → InvertSpeciesSerializer with model = InvertSpecies. That's where name and genus are writable, and InvertSpecies is registered in admin. The frontend should push proposals to invert_species, not invert_attributes.

PR #702, opened 2026-06-11
[M1963] submitted and aggregate inverts endpoints

saanobhaai:
## Summary by CodeRabbit

* **New Features**
  * Added macroinvertebrate (belt-invert) data support across Obs/SU/SE with policy-aware fields, summary/cache generation, API endpoints, filters, CSV/Geo exports, and reporting views.
  * Introduced new belt-invert method endpoints and supporting SQL-backed summary layers for observation, sample unit, and sample event.
* **Bug Fixes**
  * N/A
* **Tests**
  * Added unit tests and fixtures for belt-invert transect methods, SU/SE views, group-of-interest summaries, and cross-project access protection.
* **Documentation**
  * Updated CLAUDE.md with macroinvertebrate policy/roles guidance.
* **Chores**
  * Updated .gitignore to exclude the `plan/` directory.


saanobhaai:
[Inverts Phase 5: Aggregate API Endpoints](https://trello.com/c/X6aAipZE/1963-inverts-phase-5-aggregate-api-endpoints)

saanobhaai:
@coderabbitai review

gridcell (commented):
## AI Code Review

**Assessment:** The macroinvertebrate (belt invert) feature is well-structured and faithfully follows established BeltFish patterns for access control, data policy, SQL modelling, and serialization — the IDOR-safe update endpoint is a notable improvement over the BeltFish baseline. The main gaps are test coverage (no aggregate-endpoint, data-policy, or round-trip observation tests), a missing `sample_event_id` index on the new summary tables, and a `data_policy_macroinvertebrate` default-value divergence. No blocking issues.

| 0 | 5 | 3 | 4 | 3 |

**Files reviewed:** 19 / 19

### Cross-Cutting Concerns
- **Test coverage is the largest gap.** Three new aggregate endpoints (Obs/SU/SE), their serializers, filter sets, CSV serializers, and the data-policy access-control logic ship with zero dedicated tests; the two included tests cover only the transect-method PUT at a shallow level. Sibling protocols (BeltFish, BenthicLIT) establish the pattern to follow.
- The new `MACROINVERTEBRATE_PROTOCOL` branches in `summary_cache.py` / `summary_csv_cache.py` would be incidentally covered if the recommended aggregate-endpoint tests exercise the full cache path; otherwise a mis-mapped model pair stays undetected until runtime.
- The missing `sample_event_id` index (CR-004) is cheapest to add now while the tables are empty — once populated it needs `CREATE INDEX CONCURRENTLY` to avoid a table lock.

### General Findings

<details>
<summary><strong>CR-001</strong> &middot; Warning &middot; <code>test_gap</code> — No tests for the three new aggregate endpoints (Obs/SU/SE)</summary>

The PR registers `BeltInvertProjectMethodObsView`, `BeltInvertProjectMethodSUView`, and `BeltInvertProjectMethodSEView` at new URL patterns. Sibling protocols have dedicated tests in `test_sample_unit_method_obs/su/se.py` verifying response count, field presence, computed values, GeoJSON, and CSV columns. Belt invert has none of these.

> **Suggestion:** Add tests following the `test_beltfish_se_view/su_view/csv_view` patterns: assert response count, presence of `density_indha` and `density_indha_group_interest` in SU responses, `density_indha_avg` in SE responses, and correct CSV column count.

> **Why it matters:** The aggregate views, serializers, and filter sets are all new code with no test coverage. Regressions in field exclusion, ordering, or filters would go undetected.

Applies to: `src/api/resources/sampleunitmethods/beltinvertmethod.py`, `src/api/tests/test_sample_unit_method_obs.py`, `src/api/tests/test_sample_unit_method_su.py`, `src/api/tests/test_sample_unit_method_se.py`

</details>

<details>
<summary><strong>CR-002</strong> &middot; Warning &middot; <code>test_gap</code> — No data-policy access-control tests for data_policy_macroinvertebrate</summary>

`BeltInvertProjectMethodSEView` gates SE results through `ProjectPublicSummaryPermission`, and the SE serializer masks count/density fields when `data_policy_macroinvertebrate` is below the public threshold for non-members. Sibling protocols test that sensitive fields are null for anonymous/non-member users. Belt invert has no such test.

> **Suggestion:** Add a test creating a project with `data_policy_macroinvertebrate=10` (private), call the SE endpoint as a non-member, and assert count/density fields are null while `sample_unit_count` is present. Mirror the beltfish SE access-control test.

> **Why it matters:** Without this test, a regression in the CASE expression in `SummarySampleEventSQLModel` or a mis-wired `project_policy` string would silently expose private data.

Applies to: `src/api/resources/sampleunitmethods/beltinvertmethod.py`, `src/api/models/summary_sample_events.py`, `src/api/tests/test_sample_unit_method_se.py`

</details>

<details>
<summary><strong>CR-004</strong> &middot; Warning &middot; <code>performance</code> — New managed summary tables lack sample_event_id indexes</summary>

The three new managed tables (`summary_belt_invert_obs/su/se`) only declare `db_index=True` on `project_id`. The cache-refresh logic and the SE SQL model join these on `sample_event_id`, so without an index each join degrades to a sequential scan. All analogous existing tables follow the same project_id-only pattern, so this extends an existing gap rather than introducing a new regression — but it is best addressed at table-creation time while the tables are empty.

> **Suggestion:** Add `db_index=True` to `sample_event_id` on `BeltInvertObsModel`, `BeltInvertSUModel`, and `BeltInvertSEModel` in `summaries.py`, with the corresponding migration operation.

> **Why it matters:** As macroinvertebrate data accumulates, cache refreshes and summary queries will degrade to full sequential scans.

Applies to: `src/api/models/summaries.py`, `src/api/migrations/0119_beltinvertobssqlmodel_beltinvertsesqlmodel_and_more.py`

</details>

<details>
<summary><strong>CR-008</strong> &middot; Question &middot; <code>correctness</code> — No submission/validation path for the macroinvertebrate protocol — intentionally deferred?</summary>

The PR description references "submitted and aggregate inverts endpoints", but no validation rules appear in `src/api/submission/validations/` for the macroinvertebrate protocol and there is no `parser.py` extension. Without protocol-aware validation, users could submit biologically invalid data (e.g. negative counts) via the collect record endpoint without receiving protocol-specific errors. Is this intentionally deferred to a follow-up?

Applies to: `src/api/submission/validations/`, `src/api/resources/sampleunitmethods/beltinvertmethod.py`

</details>

<details>
<summary><strong>CR-010</strong> &middot; Question &middot; <code>test_gap</code> — No tests for summary-cache / summary-CSV-cache integration paths</summary>

`summary_cache.py` and `summary_csv_cache.py` were extended with new `MACROINVERTEBRATE_PROTOCOL` branches calling `_update_cache` with the new model pairs. Is there an existing test exercising the summary-cache update path that would catch a mis-mapped model pair, or is that coverage expected to come from the aggregate-endpoint tests once added?

Applies to: `src/api/utils/summary_cache.py`, `src/api/utils/summary_csv_cache.py`, `src/api/tests/test_summary_sample_event.py`

</details>

<details>
<summary><strong>CR-012</strong> &middot; Info &middot; <code>security</code> — New SQL models reuse the framework's string-interpolated project_id (pre-existing pattern)</summary>

The new beltinvert SQL models and the added `beltinvert_su` CTE bind `project_id` via Python string interpolation rather than a parameterized query (`sqltables` `SQLTable.as_sql` performs `base_sql = self.sql % self.sql_table_params`). This is the identical convention used by every other protocol and is not newly introduced by this PR. The value is a validated Project UUID further constrained by a `::uuid` cast, so it is not currently exploitable. Flagging for awareness only; no change required in this PR (CWE-89).

Applies to: `src/api/models/sql_models/beltinvert.py`, `src/api/models/summary_sample_events.py`, `src/sqltables/datastructures.py`

</details>

<details>
<summary><strong>CR-014</strong> &middot; Info &middot; <code>maintainability</code> — CSV serializer field list repeated across three levels (Obs, SU, SE)</summary>

`ObsBeltInvertCSVSerializer`, `BeltInvertMethodSUCSVSerializer`, and `BeltInvertMethodSECSVSerializer` each repeat ~20 common `ReportField` definitions. This is consistent with other protocols (BeltFish, BenthicLIT) so it follows established convention, but the duplication is a maintenance burden if shared fields need to change.

Applies to: `src/api/resources/sampleunitmethods/beltinvertmethod.py`

</details>

### Highlights (praise)
- **CR-015** &middot; `security` — `BeltInvertMethodView.update()` resolves the target via the routed `pk` + `project_id` filter and 404s on mismatch, preventing cross-project IDOR (more secure than the BeltFish baseline; locked in by a test).
- **CR-016** &middot; `correctness` — `NULLIF` guards on every density division path prevent divide-by-zero, with `null=True` result columns completing the chain.
- **CR-017** &middot; `maintainability` — The multi-level GOI weight CTEs (`goi_weights_by_family/order/class`) provide taxonomy-agnostic density attribution, splitting coarse-level observations proportionally; clean six-branch `UNION ALL` and `MATERIALIZED` hints.

### Agent Coverage
| reviewer-security | security | 19 | 2 | 0 |
| reviewer-backend-python | backend | 19 | 6 | 0 |
| reviewer-python-quality | code_quality | 19 | 3 | 0 |
| reviewer-database | database | 19 | 5 | 0 |
| reviewer-tests | testing | 19 | 4 | 0 |

src/api/tests/test_sample_unit_method.py
```diff
@@ -28,6 +35,39 @@ def test_sampleunitmethods_list_includes_macroinvertebrate(
...
+def test_update_belt_invert_transect_method(db_setup, api_client1, project1, belt_invert1):
+    belt_invert_id = str(belt_invert1.pk)
+    url_kwargs = {"project_pk": str(project1.pk), "pk": belt_invert_id}
+    detail_url = reverse("beltinverttransectmethod-detail", kwargs=url_kwargs)
+
+    get_response = api_client1.get(detail_url)
+    assert get_response.status_code == 200
+    data = get_response.json()
+
+    put_response = api_client1.put(detail_url, data, format="json")
```
gridcell, line 47: **Warning** · `test_gap` · confidence: high

**Round-trip update test asserts only status code and returned ID — no field values or observation data verified**

`test_update_belt_invert_transect_method` GETs then PUTs the same payload and only asserts `status_code == 200` and the returned `id`. A serializer silently dropping or renaming fields would still pass. Additionally, the `belt_invert1` fixture creates zero `ObsBeltInvert` records, so the `save_one_to_many` path for `obs_belt_inverts` is never exercised with real data.

> **Suggestion:** Add assertions on round-tripped field values after the PUT (e.g. `beltinvert_transect.len_surveyed`, `depth`). Extend the fixture or add a `belt_invert1_with_obs` variant with at least one `ObsBeltInvert` and assert it survives the round-trip.

> **Why it matters:** Shallow round-trip tests miss serializer field renames and missing `source=` kwargs. The observation-update path is completely untested.

<sub>Found by: reviewer-tests · CR-003</sub>

src/api/migrations/0119_beltinvertobssqlmodel_beltinvertsesqlmodel_and_more.py
```diff
@@ -0,0 +1,609 @@
...
+        migrations.AlterField(
+            model_name="invertgenus",
+            name="group_of_interest",
+            field=models.ForeignKey(
+                on_delete=django.db.models.deletion.PROTECT,
+                related_name="genera",
+                to="api.invertgroupofinterest",
+            ),
+        ),
+    ]
```
gridcell, line 610: **Warning** · `maintainability` · confidence: high

**data_policy_macroinvertebrate uses default="" on SummarySampleEventModel while all other data_policy fields use a meaningful sentinel**

`data_policy_macroinvertebrate` is added to `SummarySampleEventModel` with `default=""`. Every other `data_policy_*` field on this model uses a meaningful default (e.g. `data_policy_benthicpqt` uses `default="public summary"`). The two concrete managed tables (restricted/unrestricted) correctly use `default="awaiting refresh"`. Since `SummarySampleEventModel` is SQL-view-backed, the default only affects the `ADD COLUMN` DDL and rows are never written directly — no runtime impact today — but the inconsistency is a maintenance hazard.

> **Suggestion:** Change the `SummarySampleEventModel` default to `"awaiting refresh"` to match the other `data_policy_*` fields, and update the model field default in `summary_sample_events.py`.

> **Why it matters:** An empty string silently diverges from all other data_policy sentinels, making "not yet refreshed" vs "no policy" harder to distinguish in queries and monitoring.

<sub>Found by: reviewer-database, reviewer-backend-python · CR-005</sub>

src/api/resources/sampleunitmethods/beltinvertmethod.py
```diff
@@ -34,3 +79,492 @@ class Meta:
...
+        is_valid = True
+        nested_data = dict(
+            sample_event=request.data.get("sample_event"),
+            beltinvert_transect=request.data.get("beltinvert_transect"),
+            observers=request.data.get("observers"),
+            obs_belt_inverts=request.data.get("obs_belt_inverts"),
+        )
+        belt_invert_data = {k: v for k, v in request.data.items() if k not in nested_data}
+        belt_invert_id = pk
+        belt_invert_data["id"] = belt_invert_id
```
gridcell: **Warning** · `correctness` · confidence: high

**belt_invert_id sourced from URL pk; request body 'id' is silently overwritten — diverges from BeltFish and can mask caller errors**

`belt_invert_id` is set to the URL `pk` parameter and `belt_invert_data["id"]` is forced to that value. `BeltFishMethodView.update()` instead reads `belt_fish_id` from the body. The new approach is more secure and fixes the prior IDOR finding, but it silently discards any `"id"` the caller supplied in the body rather than validating it matches the URL. Mobile clients that supply a mismatched id receive a silent override with no error.

> **Suggestion:** After `belt_invert_id = pk`, add:
> ```python
> body_id = request.data.get("id")
> if body_id and str(body_id) != str(belt_invert_id):
>     return Response({"id": ["ID in request body does not match URL."]}, status=status.HTTP_400_BAD_REQUEST)
> ```

> **Why it matters:** Silent ID substitution can mask integration bugs in mobile clients and makes the API contract ambiguous.

<sub>Found by: reviewer-backend-python · CR-006</sub>

src/api/models/sql_models/beltinvert.py
```diff
@@ -0,0 +1,430 @@
...
+            FROM obs_goi GROUP BY pseudosu_id, goi_name
+        ),
+        su_goi_zeroes AS MATERIALIZED (
+            SELECT a.pseudosu_id, a.goi_name, COALESCE(g.goi_count, 0) AS goi_count
+            FROM (
+                SELECT DISTINCT o.pseudosu_id, igoi.name AS goi_name
+                FROM belt_invert_obs_all o CROSS JOIN invert_group_of_interest igoi
+            ) a
+            LEFT JOIN su_goi g USING (pseudosu_id, goi_name)
+        ),
```
gridcell, line 214: **Warning** · `performance` · confidence: medium

**su_goi_zeroes CROSS JOIN against the entire invert_group_of_interest table may produce unnecessarily large intermediate sets**

The `su_goi_zeroes` CTE performs a `CROSS JOIN invert_group_of_interest` against every `pseudosu_id`, producing a zero-density row for every GOI in the entire table — not just GOIs observed within the project. This grows O(pseudoSUs × totalGOIs) and is `MATERIALIZED`, so it is always evaluated eagerly regardless of whether the output rows are consumed.

> **Suggestion:** Scope the cross join to GOIs reachable from `belt_invert_obs_all` for this project, or derive the GOI universe from the `su_goi` CTE.

> **Why it matters:** As the invert taxonomy grows, cache-refresh query time and memory scale with the product of pseudo-SU count and total GOI count. (If the intent is to always emit all global GOIs as zero-filled entries per SU, the current approach is correct and this can be dismissed.)

<sub>Found by: reviewer-backend-python · CR-007</sub>

src/api/models/sql_models/beltinvert.py
```diff
@@ -0,0 +1,430 @@
...
+        density_indha_group_interest_sd
+
+        FROM belt_invert_su
+
+        LEFT JOIN (
+            SELECT sample_event_id,
+            jsonb_object_agg(goi, ROUND(density_avg::numeric, 2))
+                FILTER (WHERE density_avg > 0) AS density_indha_group_interest_avg,
+            jsonb_object_agg(goi, ROUND(density_sd::numeric, 2))
+                FILTER (WHERE density_avg > 0) AS density_indha_group_interest_sd
```
gridcell: **Question** · `correctness` · confidence: medium

**density_indha_group_interest_sd uses FILTER (WHERE density_avg > 0) — is this the intended contract?**

In `BeltInvertSESQLModel.sql` and the `migoi` subquery in `SummarySampleEventSQLModel`, the SD column uses `FILTER (WHERE density_avg > 0)`. GOIs with average density exactly 0 are excluded from both the average and the standard deviation at the SE level, even where a non-NULL SD would be meaningful. This mirrors the existing BeltFish SE pattern exactly, suggesting it is intentional — but worth confirming the design intent given the edge case.

<sub>Found by: reviewer-backend-python, reviewer-database · CR-009</sub>

src/api/migrations/0121_vw_project_summary_sample_events_macroinvertebrate.py
```diff
@@ -0,0 +1,42 @@
+from django.db import migrations, models
+
+forward_sql = """
+DROP VIEW IF EXISTS vw_project_summary_sample_events;
+CREATE VIEW vw_project_summary_sample_events AS
+SELECT *, 'restricted'::text AS access
+FROM restricted_project_summary_se
+UNION ALL
+SELECT *, 'unrestricted'::text AS access
+FROM unrestricted_project_summary_se;
```
gridcell, line 16: **Question** · `correctness` · confidence: medium

**SELECT * in forward_sql view will silently break if the underlying table column order or count changes**

`forward_sql` recreates `vw_project_summary_sample_events` with `SELECT *, 'restricted'::text AS access FROM restricted_project_summary_se`. `data_policy_macroinvertebrate` is added by the `AddField` earlier in the same migration, so the forward path is safe today. The concern is forward-looking: a future column change to either underlying table without recreating this view would make the two `UNION ALL` branches mismatch positionally. `reverse_sql` already enumerates columns explicitly, which is the safer pattern.

> **Suggestion:** Consider enumerating columns explicitly in `forward_sql` as `reverse_sql` already does, so future schema changes require deliberate view recreation rather than silently producing mismatched columns.

<sub>Found by: reviewer-database · CR-011</sub>

src/api/resources/sampleunitmethods/beltinvertmethod.py
```diff
@@ -34,3 +79,492 @@ class Meta:
...
+    beltinvert_transect = InvertBeltTransectSerializer(source="transect")
+    observers = ObserverSerializer(many=True)
+    obs_belt_inverts = ObsBeltInvertSerializer(many=True, source="beltinvert_observations")
+
+    class Meta:
+        model = BeltInvert
+        exclude = []
+
+
+class BeltInvertMethodView(
```
gridcell, line 95: **Info** · `docs` · confidence: high

**BeltInvertMethodView and its serializer classes lack docstrings**

`BeltInvertMethodView` lacks a class-level docstring. Multiple associated serializer classes (`BeltInvertMethodSerializer`, `BeltInvertMethodObsSerializer`, and others) also lack docstrings. This is consistent with sibling modules so the priority is low, but a brief docstring on the ViewSet would aid discoverability.

<sub>Found by: reviewer-python-quality · CR-013</sub>

gridcell (commented):
## AI Code Review

**Assessment:** Solid, well-structured addition of the macroinvertebrate (belt-invert) summary feature across Obs/SU/SE, faithfully mirroring the established belt-fish patterns. Security is clean (cross-project IDOR closed with a regression test; access keyed to `data_policy_macroinvertebrate`) and the new SU/SE group-of-interest density tests assert exact numeric values. Three warnings are worth addressing before merge; none are blocking.

| 0 | 3 | 0 | 2 | 1 |

**Files reviewed:** 22 / 24
**Previously addressed:** 10 finding(s) suppressed (resolved or acknowledged in prior review)

### Cross-Cutting Concerns
- `summary_sample_events.py` bundles an unrelated change to the existing **BeltFish** SE trophic-group aggregation (switch to `*_zeroes` source + new `FILTER`) into a macroinvertebrate PR; it should be split out or explicitly justified with a belt-fish regression test.

<details>
<summary><strong>CR-004</strong> &middot; Info &middot; <code>testing</code> — No regression test for pseudo-SU width/size-bin separation or include=False exclusion</summary>

The new SU/SE GOI-density tests are strong (exact numeric assertions, family-level proportional weighting). Two gaps remain in light of the recently-resolved pseudo-SU grouping fix: (1) no test that two transects sharing sample_event/depth/number/len_surveyed but differing in `width_id` or `size_bin_id` produce SEPARATE pseudo-SUs with a correct per-width density denominator — the exact bug that fix addressed; and (2) no test exercises an `include=False` observation being excluded from total_abundance/density while still appearing in the obs view.

> **Suggestion:** Add a fixture with two obs at different widths (or size bins) under one sample event and assert two pseudo-SUs with width-specific densities, plus a test that an include=False obs is excluded from SU totals.

> **Why it matters:** The width/size-bin grouping was a critical correctness fix; without a regression test it can silently regress, and the density denominator is the headline metric of this feature.

Applies to: `src/api/tests/fixtures/macroinvertebrate.py`, `src/api/tests/test_sample_unit_method_su.py`, `src/api/tests/test_sample_unit_method_se.py`

</details>

<details>
<summary><strong>CR-005</strong> &middot; Info &middot; <code>security</code> — Raw SQL interpolates project_id via Python %-formatting (inherited pattern)</summary>

The new belt-invert SQL models pass `project_id` through `SQLTableArg(sql=project_where)` where `project_where` is `project.id = '%(project_id)s'::uuid`, executed via Python string interpolation rather than DB-driver parameter binding. The interpolated field lists are class-defined constants (safe); the only externally influenced value is `project_id`, which the permission layer loads as a Project before the SQL runs and the `::uuid` cast further constrains. This is the identical mechanism every existing protocol uses, so it is not a regression — flagged only so the team is aware the macroinvertebrate models inherit the same trust assumption.

> **Suggestion:** Optionally validate `project_pk` as a UUID at the routing/queryset layer so the safety is explicit rather than incidental.

> **Why it matters:** If a future call site reaches the SQL table with an unvalidated, non-UUID project_id bypassing the project-load permission check, the `::uuid` cast is the only barrier before interpolation.

Applies to: `src/api/models/sql_models/beltinvert.py`

</details>

<details>
<summary><strong>CR-006</strong> &middot; Praise &middot; <code>security</code> — Cross-project IDOR closed and locked in by a regression test</summary>

`BeltInvertMethodView.update` fetches the target with `transect__sample_event__site__project_id=project_pk` and returns 404 when the record is outside the routed project — stronger than the analogous `BeltFishMethodView.update`, which fetches by id alone. `test_update_belt_invert_transect_method_cross_project_id_rejected` locks the behavior in.

Applies to: `src/api/resources/sampleunitmethods/beltinvertmethod.py`

</details>

### Agent Coverage
| reviewer-security | security | 24 | 1 | 0 |
| reviewer-backend-python | backend | 14 | 2 | 0 |
| reviewer-python-quality | quality | 18 | 0 | 0 |
| reviewer-database | database | 6 | 0 | 0 |
| reviewer-tests | tests | 4 | 1 | 0 |
| reviewer-docs | docs | 1 | 1 | 0 |
| reviewer-general-purpose | general | 2 | 0 | 0 |

src/api/models/summary_sample_events.py
```diff
@@ -376,7 +403,7 @@ class SummarySampleEventSQLModel(SummarySampleEventBaseModel):
                     SELECT sample_event_id, pseudosu_id, tgdata.key AS tg,
                     SUM(tgdata.value::double precision) AS biomass_kgha
                     FROM beltfish_su,
-                    LATERAL jsonb_each_text(biomass_kgha_trophic_group) tgdata(key, value)
+                    LATERAL jsonb_each_text(biomass_kgha_trophic_group_zeroes) tgdata(key, value)
```
gridcell, line 406: **Warning** · `correctness` · confidence: medium

**Unrelated BeltFish SE behavior change bundled in: trophic-group aggregation now reads `*_zeroes` and adds a FILTER**

This macroinvertebrate PR also changes the existing **BeltFish** sample-event trophic-group aggregation in `SummarySampleEventSQLModel`: the LATERAL source switches from `biomass_kgha_trophic_group` to `biomass_kgha_trophic_group_zeroes` (this line) and both `biomass_kgha_trophic_group_avg` and `_sd` gain `FILTER (WHERE biomass_kgha_avg > 0)` (lines 393, 397). Pulling from the zero-filled per-SU map changes the denominator: a trophic group present in only some SUs is now averaged across ALL pseudo-SUs in the sample event (zeros included) rather than only the SUs where it was observed, and the FILTER then drops groups whose SE-level average rounds to 0. This is plausibly an intentional alignment with the new invert GOI density, but it silently alters production BeltFish SE summary values and is unrelated to the inverts feature.

> **Suggestion:** Either split this BeltFish change into its own PR, or keep it here with (a) a note in the PR description explaining the intended semantics and (b) a BeltFish SE regression test asserting the new trophic-group avg/sd values.

> **Why it matters:** An unannounced averaging-semantics change to an existing protocol can shift published summary numbers for every project with belt-fish data, with no test to catch a regression and no obvious link to the PR's scope.

src/api/resources/sampleunitmethods/beltinvertmethod.py
```diff
@@ -34,3 +79,491 @@ class Meta:
...
+        errors = {}
+        is_valid = True
+        nested_data = dict(
+            sample_event=request.data.get("sample_event"),
+            beltinvert_transect=request.data.get("beltinvert_transect"),
+            observers=request.data.get("observers"),
+            obs_belt_inverts=request.data.get("obs_belt_inverts"),
+        )
+        belt_invert_data = {k: v for k, v in request.data.items() if k not in nested_data}
+        belt_invert_id = belt_invert_data["id"]
```
gridcell, line 118: **Warning** · `error_handling` · confidence: high

**`belt_invert_data["id"]` raises KeyError (500) when the body omits `id`; the URL `pk` is never used for the lookup**

`belt_invert_id = belt_invert_data["id"]` does an unguarded dict subscript — a PUT whose body lacks an `id` key produces an unhandled KeyError → HTTP 500 instead of a clean 400. Separately, although the cross-project IDOR is correctly closed by scoping the lookup to `transect__sample_event__site__project_id=project_pk`, the record is still selected by the body-supplied `id` and the routed `pk` argument is ignored entirely, so the URL detail id and the body id can diverge silently (as long as both resolve within the project).

> **Suggestion:** Prefer the routed pk and validate it, e.g. `belt_invert_id = pk or belt_invert_data.get("id")`, return 400 when missing, and reject (or ignore) a body `id` that does not match `pk`. Use that single trusted value for the project-scoped `get()` and the `save_one_to_many` foreign-key tuples.

> **Why it matters:** Unhandled input shapes degrading to 500s are poor API hygiene, and letting the body id select the record keeps the contract ambiguous for mobile clients.

<sub>Found by: reviewer-backend-python, reviewer-security · CR-002</sub>

CLAUDE.md
```diff
@@ -132,7 +132,7 @@ src/
 
 **1. Protocol-Based Data Collection**
 
-MERMAID supports multiple survey protocols (Belt Fish, Benthic LIT, Benthic PIT, Benthic PQT, Bleaching QC, Habitat Complexity). Each protocol has:
+MERMAID supports multiple survey protocols (Belt Fish, Benthic LIT, Benthic PIT, Benthic PQT, Bleaching QC, Habitat Complexity). Belt Invertebrate (Macroinvertebrate) exists in resources and models but does not yet have SQL summary models. Each protocol has:
```
gridcell: **Warning** · `docs` · confidence: high

**CLAUDE.md claims macroinvertebrate "does not yet have SQL summary models" — contradicted by this PR**

This line is edited in the same PR that ADDS `BeltInvertObsSQLModel`/`BeltInvertSUSQLModel`/`BeltInvertSESQLModel` and the `summary_belt_invert_obs/su/se` tables. The statement "Belt Invertebrate (Macroinvertebrate) exists in resources and models but does not yet have SQL summary models" is false as of this commit.

> **Suggestion:** Add Macroinvertebrate to the list of protocols with full summary support and drop the "does not yet have SQL summary models" clause, e.g. "...Bleaching QC, Habitat Complexity, Macroinvertebrate). Each protocol has:".

> **Why it matters:** CLAUDE.md is the canonical orientation doc; a statement it contradicts within the same PR will mislead future contributors.

<sub>Found by: reviewer-docs · CR-003</sub>

src/api/resources/sampleunitmethods/beltinvertmethod.py
saanobhaai: Not fixing: the cross-project IDOR is already closed by the existing project-scoped .get(id=..., transect__sample_event__site__project_id=project_pk) lookup (404 on mismatch), which was retained from 4a2cea07. The pk-vs-body-id reconciliation was tried in that same commit but reverted in dbc87e79 because no other protocol's update() (beltfish, benthiclit, benthicpit, habitatcomplexity, bleachingqc) does this — they all use a bare xxx_data["id"] subscript with no pk check. Adding it only here would be inconsistent; the bare-subscript/500-on-missing-id issue is pre-existing across all protocols and out of scope for this PR.

See also preexisting discussion in review comments.

gridcell (approved):

PR #703, opened 2026-06-16
tune alarms and project fetch

ms280690:
Tune CloudWatch alarms and reduce /v1/projects/ query count

  Alarms (iac/stacks/constructs/alerts.py, api.py):
  - ALB p99 latency threshold is now env-specific: 5s prod, 10s dev. Dev is
    low-traffic, so p99 equals the single slowest request and a lone slow probe
    was tripping the 5s threshold (mermaid-dev-alb-p99-latency false alarm).
  - Fix EcsTaskStoppedAlarm reading RunningTaskCount from the AWS/ECS namespace,
    which has no such metric — with treat_missing=BREACHING it could never
    correctly detect a stopped service. Now reads from ECS/ContainerInsights
    (enabled on the cluster), matching ApiMemoryAlarm.
  - Only the primary env (prod) owns the RDS alarms via the new
    monitor_shared_rds flag. The RDS instance is shared across envs, so dev and
    prod were both creating alarms on the same instance and double-paging; the
    dev ones are removed.
  - General/Image DLQ alarms now require 2 datapoints instead of 1, so a message
    cleared by a retry/redrive within 5 minutes no longer pages.

  API (src/api/resources/project.py):
  - Remove N+1 queries in the project list serializer. get_num_sites and
    get_num_active_sample_units used .count() and get_project_admins used
    .filter(), all of which bypass the prefetch cache and issue a query per
    project. Use len()/Python filtering over prefetched data and prefetch
    collect_records and profiles__profile.

## Release Notes

* **Bug Fixes**
  * Improved API and alert behavior with environment-specific p99 latency thresholds, optional shared-RDS alarm gating to prevent duplicate paging, and more correct ECS task-stopped metrics via Container Insights.
  * Tightened DLQ alarm evaluation to detect threshold changes faster.
  * Optimized project serializer computations using pre-fetched data and queryset annotations to avoid extra database work.

* **Tests**
  * Added coverage for project serializer member/admin fields on the project details endpoint.


ms280690:
@saanobhaai When running locally via docker 
```
=========================== 415 passed, 750 warnings in 192.73s (0:03:12) ===========================
```

saanobhaai (approved):

PR #704, opened 2026-06-17
[M1999] obs agg view enhancements

saanobhaai:
## Release Notes

* **New Features**
  * Macroinvertebrate observations now support a detailed taxonomy breakdown as separate fields (taxon, group of interest, class, order, family, genus, species).
  * Added new fields for density (`density_indha`) and `observation_notes`.
  * CSV export now includes the new taxonomy, density, and notes columns.
  * Macroinvertebrate species Excel worksheet title updated to **“Macroinvertebrate Species”**.

* **Improvements**
  * Updated macroinvertebrate protocol label to **“Macroinvertebrate Belt”**.
  * Observation listings and exports now sort by **taxon**.


saanobhaai:
[invert obs aggregate view enhancements](https://trello.com/c/X6JqCRva/1999-invert-obs-aggregate-view-enhancements)

src/api/models/view_models/__init__.py
```diff
@@ -1,3 +1,4 @@
 # view_models should only create postgres views that other db entities do not depend on
 
 from .fish_attributes import *  # noqa: F403
+from .invert_attributes import *  # noqa: F403
```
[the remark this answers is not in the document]
saanobhaai: Not valid — ruff passes without F401 on the invert line, same as the existing fish line.

src/api/signals/attributes.py
```diff
@@ -40,6 +66,10 @@ def refresh_attribute_views(sender, instance, **kwargs):
         logger.info("refresh fish")
         call_command("refresh_view", "mv_fish_attributes")
 
+    if sender in invert_models:
+        logger.info("refresh invert")
+        call_command("refresh_view", "mv_invert_attributes")
```
gridcell, line 75: **Warning** · `correctness` · confidence: high

**REFRESH MATERIALIZED VIEW CONCURRENTLY fires inside a Django transaction block, which PostgreSQL forbids**

call_command('refresh_view', 'mv_invert_attributes') runs REFRESH MATERIALIZED VIEW CONCURRENTLY synchronously inside post_save/post_delete. Django signals fire within the active transaction. PostgreSQL raises ERROR: REFRESH MATERIALIZED VIEW CONCURRENTLY cannot run inside a transaction block. The fish_attributes signal uses the identical structural pattern; admin simple saves often run in autocommit which may explain why it has not surfaced in production, but any save wrapped in atomic() — management commands, bulk imports, ATOMIC_REQUESTS — will hit this hard runtime error. This is a replicated pre-existing pattern, not a newly introduced regression, but the new invert signal is an opportunity to fix both receivers correctly.

> **Suggestion:** Wrap the refresh in transaction.on_commit(lambda: call_command('refresh_view', 'mv_invert_attributes')) and apply the same fix to the fish_attributes refresh signal.

> **Why it matters:** Any invert or fish attribute save inside an atomic block will raise a PostgreSQL exception and fail the surrounding transaction.

<sub>Found by: reviewer-database · CR-001</sub>

src/api/signals/attributes.py
```diff
@@ -20,6 +26,14 @@
...
 benthic_models = [BenthicAttribute, GrowthForm, Region]
 fish_models = [FishGrouping, FishFamily, FishGenus, FishSpecies, Region]
+invert_models = [
+    InvertGroupOfInterest,
+    InvertClass,
+    InvertOrder,
+    InvertFamily,
+    InvertGenus,
+    InvertSpecies,
+]
```
gridcell, line 38: **Warning** · `correctness` · confidence: high

**Base InvertAttribute model missing from signal receiver list — direct operations on parent MTI table bypass matview refresh**

invert_models contains the six subclasses but not InvertAttribute itself. Deleting via the base queryset (e.g. InvertAttribute.objects.filter(...).delete()) sends a signal only for InvertAttribute, not for a subclass, so the materialized view can serve stale data after direct base-table operations. The matview will not be refreshed until an unrelated subclass edit triggers the signal.

> **Suggestion:** Add InvertAttribute to the imports and to the invert_models list in the signal receiver.

> **Why it matters:** Direct base-queryset deletes leave the matview referencing rows that no longer exist until an unrelated save triggers a refresh.

<sub>Found by: reviewer-database · CR-002</sub>

src/api/resources/sampleunitmethods/beltinvertmethod.py
```diff
@@ -212,13 +212,20 @@ class Meta(BaseSUViewAPISerializer.Meta):
...
+                "invert_order",
+                "invert_family",
+                "invert_genus",
+                "invert_taxon",
+                "size_bin",
                 "count",
                 "size",
+                "density_indha",
+                "observation_notes",
+                "invert_group_of_interest",
```
gridcell: **Warning** · `api_contract` · confidence: high

**invert_group_of_interest placed after observation_notes, breaking taxonomy field grouping in CSV/JSON column order**

header_order groups invert_class/order/family/genus/taxon, then jumps to size_bin/count/size/density_indha/observation_notes, and only then appends invert_group_of_interest as the final column — separated from the taxonomy block. ObsBeltInvertCSVSerializer has the same misplacement. Observation-export column order is part of the public data contract for researchers and is cheaper to fix before the field ships.

> **Suggestion:** Move invert_group_of_interest adjacent to the other taxonomy fields (e.g. immediately after invert_taxon) in both header_order and the CSV field list.

> **Why it matters:** Downstream researchers and tools that rely on stable column positions in CSV exports will receive a disorganised column layout that is a breaking change to fix later.

src/api/resources/sampleunitmethods/beltinvertmethod.py
```diff
@@ -199,7 +199,7 @@ class BeltInvertMethodObsSerializer(BaseSUViewAPISerializer):
     class Meta(BaseSUViewAPISerializer.Meta):
         model = BeltInvertObsModel
         exclude = BaseSUViewAPISerializer.Meta.exclude.copy()
-        exclude.extend(["location", "include"])
+        exclude.extend(["location", "include", "invert_species"])
```
gridcell, line 202: **Question** · `maintainability` · confidence: high

**invert_species added to model and matview but silently excluded from all API and CSV output with no explanation**

invert_species is projected into BeltInvertObsSQLModel and BeltInvertObsModel, then excluded from BeltInvertMethodObsSerializer, BeltInvertMethodObsGeoSerializer, and ObsBeltInvertCSVSerializer. The field exists in the database result set but never reaches a consumer. No comment explains whether this is intentional (invert_taxon already renders 'Genus species') or an oversight. Future developers will spend time re-investigating why the field exists but is never surfaced.

> **Suggestion:** If intentional, add an inline comment in the serializer Meta class explaining the suppression. If accidental, add invert_species to header_order and the CSV field list.

> **Why it matters:** Analysts get 'Genus epithet' in invert_taxon but no standalone species column; the intent is ambiguous without documentation.

<sub>Found by: reviewer-database · CR-005</sub>

src/api/models/sql_models/beltinvert.py
```diff
@@ -109,9 +104,17 @@ class BeltInvertObsSQLModel(BaseSUSQLModel):
...
+    invert_taxon = models.CharField(max_length=200, null=True, blank=True)
+    invert_group_of_interest = models.CharField(max_length=200, null=True, blank=True)
+    invert_class = models.CharField(max_length=200, null=True, blank=True)
+    invert_order = models.CharField(max_length=200, null=True, blank=True)
+    invert_family = models.CharField(max_length=200, null=True, blank=True)
+    invert_genus = models.CharField(max_length=200, null=True, blank=True)
+    invert_species = models.CharField(max_length=200, null=True, blank=True)
     count = models.PositiveIntegerField(null=True, blank=True)
     size = models.DecimalField(max_digits=5, decimal_places=1, null=True, blank=True)
+    density_indha = models.DecimalField(max_digits=11, decimal_places=2, null=True, blank=True)
```
gridcell, line 116: **Info** · `correctness` · confidence: high

**density_indha DecimalField(max_digits=11) overflow constraint is inert on managed=False view model**

density_indha is declared as DecimalField(max_digits=11, decimal_places=2) on the managed=False BeltInvertObsSQLModel and BeltInvertObsModel. Because the field is SQL-computed and the model is read-only, the DecimalField max_digits validator never fires at the application layer. The underlying column is numeric (unbounded) in the view. Real-world counts make overflow impossible in practice. Documentation note only — no change required.

<sub>Found by: reviewer-backend-python · CR-011</sub>

gridcell (commented):
## AI Code Review

**Assessment:** Well-structured expansion of the belt invert observation aggregation view introducing mv_invert_attributes and taxonomy rollups. No security vulnerabilities or blocking defects. Four warnings require attention before merge: the synchronous REFRESH MATERIALIZED VIEW CONCURRENTLY inside post_save signals (a pre-existing pattern in fish_attributes that risks failure inside any atomic block), a missing base InvertAttribute model in the signal receivers, a misplaced invert_group_of_interest column in the export field ordering, and duplicated materialized view SQL across four files. Two addressed findings were suppressed.

| 0 | 4 | 3 | 3 | 2 |

**Files reviewed:** 12 / 12

**Previously addressed:** 2 finding(s) suppressed (resolved or acknowledged in prior review)

### Cross-Cutting Concerns
- The synchronous REFRESH MATERIALIZED VIEW CONCURRENTLY pattern is shared between the new invert signal and the existing fish_attributes signal. The on_commit() fix in CR-001 should be applied to both receivers together to avoid a two-class maintenance inconsistency.
- The new mv_invert_attributes materialized view introduces a second matview in the codebase without a documented staleness contract. The DROP-prefix divergence flagged in CR-007 means the migration path and the refresh_view path are not fully equivalent — worth confirming once against the fish_attributes setup procedure.
- The invert_species field travels through migration SQL, the matview, the SQL model, and the ORM model, but is deliberately excluded from all serializers. This project-then-suppress pattern creates a discoverability hazard; a short comment on the serializer Meta stating the reason would eliminate future investigation cost.
- No new tests were added alongside the new signal receivers or the new serializer fields. A signal smoke-test and a CSV column-order assertion would match the existing test coverage patterns for similar belt-fish features.

### Additional Findings

<details>
<summary><strong>CR-004</strong> &middot; Warning &middot; <code>maintainability</code> — Materialized view SQL definition duplicated across four locations</summary>

The full CREATE MATERIALIZED VIEW mv_invert_attributes DDL (~50 lines) is copy-pasted into migration 0122 forward, migration 0124 forward, migration 0124 reverse (correctly restoring the 0122 body), and InvertAttributeView.sql used by refresh_view and model_view_migrations. Four co-equal sources of truth can drift independently. If the definition evolves, all copies must be updated in sync; a missed copy causes subtle bugs during rollbacks or when applying migrations in different environments.

> **Suggestion:** Extract the SQL DDL to a single authoritative location (e.g. a .sql file or a constant in model_view_migrations) and reference it from both migration files and the view model. Follow the existing fish_attributes pattern if one exists.

> **Why it matters:** SQL drift across migrations and the live view model can produce incorrect rollback behaviour or silent data differences depending on which code path created the view.

Applies to: `src/api/migrations/0122_mv_invert_attributes.py`, `src/api/migrations/0124_mv_invert_attributes_goi_family_order_class.py`, `src/api/models/view_models/invert_attributes.py`, `src/api/models/view_models/model_view_migrations.py`

<sub>Found by: reviewer-database, reviewer-python-quality</sub>
</details>

<details>
<summary><strong>CR-006</strong> &middot; Question &middot; <code>correctness</code> — name_goi correlated subqueries exclude genera with NULL group_of_interest_id — intentional?</summary>

The order/class name_goi correlated subqueries INNER JOIN invert_genus to invert_group_of_interest, so genera with NULL group_of_interest_id are excluded from the aggregated string. This probably matches the design intent, but it differs from individual genus rows (which return NULL). Confirm the design intent is correct before shipping.

Applies to: `src/api/migrations/0124_mv_invert_attributes_goi_family_order_class.py`, `src/api/models/view_models/invert_attributes.py`

</details>

<details>
<summary><strong>CR-007</strong> &middot; Question &middot; <code>correctness</code> — Migration 0124 forward_sql and InvertAttributeView.sql differ by a DROP prefix — confirm refresh_view path is correct</summary>

Migration 0124 forward_sql prepends DROP MATERIALIZED VIEW IF EXISTS before the CREATE; InvertAttributeView.sql (used by refresh_view and model_view_migrations) starts directly with CREATE ... IF NOT EXISTS. On a clean database the refresh_view-created view relies on IF NOT EXISTS to avoid replacing the definition, while only the migration path explicitly drops and recreates it. Confirm this matches the fish_attributes pattern and the intended behaviour.

Applies to: `src/api/migrations/0124_mv_invert_attributes_goi_family_order_class.py`, `src/api/models/view_models/invert_attributes.py`, `src/api/models/view_models/model_view_migrations.py`

</details>

<details>
<summary><strong>CR-008</strong> &middot; Info &middot; <code>security</code> — Signal refreshes materialized view via a management command that builds SQL with str.format() on the view name</summary>

call_command('refresh_view', 'mv_invert_attributes') ultimately builds SQL with 'REFRESH MATERIALIZED VIEW CONCURRENTLY {}'.format(viewname). The view name is always a hardcoded literal at all current call sites, so there is no injection risk here. fish_attributes uses the identical safe pattern. Noted only because formatting an identifier directly into raw SQL is fragile if a future caller passes non-constant input.

> **Why it matters:** str.format() into raw SQL is safe only while every caller supplies a trusted constant (CWE-89). No change required today.

Applies to: `src/api/signals/attributes.py`, `src/tools/management/commands/refresh_view.py`

</details>

<details>
<summary><strong>CR-009</strong> &middot; Info &middot; <code>security</code> — New observation fields broaden export surface but expose no new sensitive data</summary>

Adds taxonomy rollups, density_indha, and observation_notes to BeltInvertObsModel, serializers, and CSV output. All new fields are governed by the existing data_policy_macroinvertebrate authorization path; no new authz bypass or PII is introduced. invert_species is correctly excluded from JSON and GeoJSON serializers. observation_notes is free-text rendered only into CSV/JSON (no HTML render), so there is no stored-XSS surface. Confirm that observation_notes is within the existing macroinvertebrate data-sharing policy scope.

> **Why it matters:** Adding fields widens returned data; free-text notes should be confirmed in policy scope (CWE-200).

Applies to: `src/api/models/summaries.py`, `src/api/models/sql_models/beltinvert.py`, `src/api/resources/sampleunitmethods/beltinvertmethod.py`

</details>

<details>
<summary><strong>CR-010</strong> &middot; Info &middot; <code>performance</code> — Full REFRESH MATERIALIZED VIEW CONCURRENTLY on every attribute save may degrade at scale</summary>

refresh_view is called synchronously on every post_save/post_delete for invert attributes, unlike report generation which is offloaded via submit_job (SQS). CONCURRENTLY still rebuilds the entire snapshot including correlated subqueries. Acceptable for a small reference dataset in its current state, but architecturally inconsistent with heavier async tasks and a latency risk if taxonomy grows significantly.

Applies to: `src/api/signals/attributes.py`

</details>

<details>
<summary><strong>CR-012</strong> &middot; Praise &middot; <code>correctness</code> — density_indha correctly guards divide-by-zero with NULLIF on the len*width product</summary>

ROUND(o.count::numeric / NULLIF(su.len_surveyed * w.val, 0) * 10000, 2) applies NULLIF to the full product and casts to numeric before dividing. Zero-length or zero-width transects return NULL rather than raising a division-by-zero error or producing integer truncation. Correctly handled.

Applies to: `src/api/models/sql_models/beltinvert.py`

</details>

<details>
<summary><strong>CR-013</strong> &middot; Praise &middot; <code>maintainability</code> — Clean refactor: replacing seven direct LEFT JOINs with a single mv_invert_attributes join</summary>

Replacing seven taxonomy LEFT JOINs with a single LEFT JOIN mv_invert_attributes in BeltInvertObsSQLModel meaningfully reduces SQL complexity, centralises taxonomy resolution, and aligns with the fish_attributes materialized view pattern. Good structural improvement.

Applies to: `src/api/models/sql_models/beltinvert.py`

</details>

### Agent Coverage
| reviewer-security | security | 12 | 2 | 0 |
| reviewer-database | database | 12 | 7 | 0 |
| reviewer-backend-python | backend | 12 | 4 | 0 |
| reviewer-python-quality | code_quality | 12 | 1 | 0 |

src/api/resources/sampleunitmethods/beltinvertmethod.py
saanobhaai: I actually agree but this was requested deliberately, to be consistent with fish obs handling

src/api/signals/attributes.py
saanobhaai: REFRESH MATERIALIZED VIEW CONCURRENTLY was restricted inside transaction blocks in older PostgreSQL, but that restriction was lifted in PostgreSQL 14. This codebase runs PostgreSQL 16, where it works fine inside Django's atomic() and in the test transaction wrapper. Wrapping with transaction.on_commit was unnecessary and broke all 8 tests by deferring the refresh past the point where tests read the matview.

PR #706, opened 2026-06-19
summary cache waiting for recently triggered reports

saanobhaai:
## Release Notes

* **New Features**
  * Report emails now optionally include a “stale summary data” notice when the underlying summary cache may not be fully up to date at generation time, so recipients better understand data completeness.

* **Tests**
  * Added new pytest coverage for summary-cache freshness waiting, race-condition handling, timeout behavior, and correct propagation of the staleness flag through report generation to email delivery.

### DIFF
--- a/src/api/tests/test_summary_cache_wait.py
+++ b/src/api/tests/test_summary_cache_wait.py
@@ -1,0 +1,70 @@
+import uuid
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+from api.models import SummaryCacheQueue
+from api.utils.reports import create_sample_unit_method_summary_report
+from api.utils.summary_cache import wait_for_summary_cache
+
+
+def test_wait_skips_unrelated_projects_and_returns_false():
+    """Only entries for the requested project_ids trigger a wait; others are ignored."""
+    SummaryCacheQueue.objects.create(project_id=uuid.uuid4())
+
+    with patch("api.utils.summary_cache.time") as mock_time:
+        mock_time.monotonic.return_value = 0.0
+        result = wait_for_summary_cache([uuid.uuid4()])
+
+    assert result is False
+    mock_time.sleep.assert_not_called()
+
+
+def test_wait_polls_until_entry_cleared():
+    project_id = uuid.uuid4()
+    SummaryCacheQueue.objects.create(project_id=project_id)
+    sleep_calls = [0]
+
+    def fake_sleep(_):
+        sleep_calls[0] += 1
+        if sleep_calls[0] >= 2:
+            SummaryCacheQueue.objects.filter(project_id=project_id).delete()
+
+    with (
+        patch("api.utils.summary_cache.SUMMARY_CACHE_MAX_WAIT", 30),
+        patch("api.utils.summary_cache.time.sleep", side_effect=fake_sleep),
+    ):
+        result = wait_for_summary_cache([project_id])
+
+    assert result is False
+
+
+def test_wait_times_out_logs_warning_and_returns_true():
+    project_id = uuid.uuid4()
+    SummaryCacheQueue.objects.create(project_id=project_id)
+    with (
+        patch("api.utils.summary_cache.logger") as mock_logger,
+        patch("api.utils.summary_cache.SUMMARY_CACHE_MAX_WAIT", -1),
+    ):
+        result = wait_for_summary_cache([project_id])
+    assert result is True
+    mock_logger.warning.assert_called_once()
+    assert "Timed out" in mock_logger.warning.call_args[0][0]
+
+
+@pytest.mark.parametrize("stale", [True, False])
+def test_report_passes_stale_flag_to_email(project1, stale):
+    mock_wb = MagicMock()
+    with (
+        patch("api.utils.reports.wait_for_summary_cache", return_value=stale),
+        patch("api.utils.reports.create_protocol_report", return_value=mock_wb),
+        patch("api.utils.reports.email_report") as mock_email,
+    ):
+        create_sample_unit_method_summary_report(
+            project_ids=[project1.pk],
+            protocol="fishbelt",
+            send_email="test@example.com",
+            request=MagicMock(user=MagicMock(profile=MagicMock(email="test@example.com"))),
+            wait_for_cache=True,
+        )
+    assert mock_email.call_args.kwargs.get("data_may_be_stale") is stale
--- a/src/api/tests/test_summary_cache_wait.py
+++ b/src/api/tests/test_summary_cache_wait.py
@@ -5,7 +5,7 @@ import pytest
 
 from api.models import SummaryCacheQueue
 from api.utils.reports import create_sample_unit_method_summary_report
-from api.utils.summary_cache import wait_for_summary_cache
+from api.utils.summary_cache import SUMMARY_CACHE_POLL_INTERVAL, wait_for_summary_cache
 
 
 def test_wait_skips_unrelated_projects_and_returns_false():
@@ -33,6 +33,7 @@ def test_wait_polls_until_entry_cleared():
     with (
         patch("api.utils.summary_cache.SUMMARY_CACHE_MAX_WAIT", 30),
         patch("api.utils.summary_cache.time.sleep", side_effect=fake_sleep),
+        patch("api.utils.summary_cache.connection"),
     ):
         result = wait_for_summary_cache([project_id])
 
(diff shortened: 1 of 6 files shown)


src/api/utils/reports.py
```diff
@@ -55,6 +56,8 @@ def create_sample_unit_method_summary_report_background(
         protocol,
         request=req,
         send_email=send_email,
+        wait_for_cache=True,
+        visibility_timeout=120,
```
gridcell: **Warning** · `correctness` · confidence: high

**One-shot 120s SQS visibility is not renewed; wait + report generation can exceed it and cause re-delivery**

`extend_job_visibility()` is called exactly once before `job.run()` (`simpleq/workers.py:49-51`), setting the message visibility to 120s from that moment. `wait_for_summary_cache` can itself consume up to `SUMMARY_CACHE_MAX_WAIT` (80s), after which report generation + S3 upload + email still has to run. The in-code comment budgets ~30s for that (80 + 30 = 110 < 120), but for large projects report generation alone can exceed the remaining headroom. If total job time crosses 120s, SQS makes the message visible again and re-delivers it, producing a duplicate report email.

> **Suggestion:** Either set `visibility_timeout` to `SUMMARY_CACHE_MAX_WAIT` plus a conservative generation budget (e.g. 180s) and document the headroom, or split the wait and the report generation into two jobs so the wait can't eat into the generation window.

> **Why it matters:** Duplicate job execution emails a second report and repeats expensive S3/DB work; the two emails may disagree on the stale-data notice.

src/api/utils/summary_cache.py
```diff
@@ -303,6 +305,40 @@ def add_project_to_queue(project_id, skip_test_project=False):
...
+        if not pending_ids:
+            return False
+        if time.monotonic() >= deadline:
+            logger.warning(
+                "Timed out waiting for summary cache to update for projects %s. "
+                "Report may contain stale data.",
+                pending_ids,
+            )
+            return True
+        time.sleep(SUMMARY_CACHE_POLL_INTERVAL)
```
gridcell, line 343: **Warning** · `performance` · confidence: high

**DB connection held open (idle) across every sleep in the poll loop**

Each loop iteration checks out a Django DB connection for the `SummaryCacheQueue` query, and Django does not release it during `time.sleep()`. Over an 80s wait the connection stays checked out but idle through up to ~15 sleep gaps. Under concurrent report jobs this can occupy several pooled connections and, on a small `max_connections`, contribute to pool pressure.

> **Suggestion:** Close the connection before sleeping and let Django reopen it on the next query:
> ```python
> from django.db import connection as db_connection
> ...
>     db_connection.close()
>     time.sleep(SUMMARY_CACHE_POLL_INTERVAL)
> ```

> **Why it matters:** Holding idle DB connections for tens of seconds under concurrency can exhaust the pool and surface as connection-timeout errors elsewhere.

<sub>Found by: reviewer-backend-python · CR-002</sub>

src/api/utils/reports.py
```diff
@@ -73,6 +77,8 @@ def create_sample_unit_method_summary_report(
     if isinstance(project_ids, list) is False:
         project_ids = [project_ids]
 
+    data_may_be_stale = wait_for_summary_cache(project_ids) if wait_for_cache else False
```
gridcell, line 81: **Question** · `correctness` · confidence: medium

**TOCTOU window: a project queued after the wait returns False is reported as fresh**

When `wait_for_summary_cache` returns `False` the queue was empty at the moment of the final check, but there is no lock between that check and `create_protocol_report`. A submission that queues a project in that window will be included in the report with `data_may_be_stale=False` (no warning). This is inherent to a polling design, but worth acknowledging.

> **Suggestion:** Document the race in the docstring, and optionally re-check `SummaryCacheQueue.objects.filter(project_id__in=project_ids).exists()` immediately before generating and force `data_may_be_stale=True` if non-empty.

> **Why it matters:** A user who just submitted data could get a report that looks fresh but silently omits their latest submission.

src/api/tests/test_summary_cache_wait.py
```diff
@@ -0,0 +1,70 @@
...
+    with (
+        patch("api.utils.summary_cache.SUMMARY_CACHE_MAX_WAIT", 30),
+        patch("api.utils.summary_cache.time.sleep", side_effect=fake_sleep),
+    ):
+        result = wait_for_summary_cache([project_id])
+
+    assert result is False
+
+
+def test_wait_times_out_logs_warning_and_returns_true():
```
gridcell, line 43: **Warning** · `test_gap` · confidence: high

**Timeout test exits on the first iteration; the multi-poll-then-timeout path is never exercised**

`SUMMARY_CACHE_MAX_WAIT=-1` puts the deadline in the past before the call, so on the first iteration the deadline check is already true and the function returns `True` without ever sleeping. The realistic path — poll several times, then time out — is not covered. A regression that broke the sleep/loop would still pass this test.

> **Suggestion:** Add a variant with a positive `MAX_WAIT` and a `time.monotonic` side_effect that advances past the deadline only after N iterations, asserting `time.sleep` was called >= 1 time before the warning fires. Keep the current test for the return-value/log assertions.

> **Why it matters:** Without exercising multiple polls, the test can't catch a loop that never sleeps or never re-queries.

<sub>Found by: reviewer-tests · CR-004</sub>

src/api/tests/test_summary_cache_wait.py
```diff
@@ -0,0 +1,70 @@
...
+    sleep_calls = [0]
+
+    def fake_sleep(_):
+        sleep_calls[0] += 1
+        if sleep_calls[0] >= 2:
+            SummaryCacheQueue.objects.filter(project_id=project_id).delete()
+
+    with (
+        patch("api.utils.summary_cache.SUMMARY_CACHE_MAX_WAIT", 30),
+        patch("api.utils.summary_cache.time.sleep", side_effect=fake_sleep),
```
gridcell, line 35: **Warning** · `test_gap` · confidence: medium

**The sleep interval (SUMMARY_CACHE_POLL_INTERVAL) passed to time.sleep is never asserted**

`test_wait_polls_until_entry_cleared` attaches a side_effect to `time.sleep` to clear the queue but never checks the argument. If the implementation slept on a hardcoded value instead of `SUMMARY_CACHE_POLL_INTERVAL`, the tests would still pass.

> **Suggestion:** Capture the mock and assert all sleep calls receive the constant: `assert all(c.args[0] == SUMMARY_CACHE_POLL_INTERVAL for c in mock_sleep.call_args_list)`.

> **Why it matters:** The poll interval governs DB load and responsiveness; a silent regression would go undetected.

<sub>Found by: reviewer-tests · CR-005</sub>

src/templates/emails/report.txt
```diff
@@ -3,8 +3,11 @@
...
     Your {{title}} report is ready for download. Please click the link below to download the report.
-    
+
     {{file_url|safe}}
 
     Link will be valid for 7 days.
+{% if data_may_be_stale %}
+    Note: Summary data was still being updated when this report was generated. Some records may not reflect the most recent submissions.
+{% endif %}
 {% endblock %}
```
gridcell, line 13: **Info** · `style` · confidence: high

**report.txt has no trailing newline at EOF (pre-existing)**

The file ends with `{% endblock %}` and no trailing newline. This predates the PR (the endblock line is unchanged context), so it's not introduced here — but since the file is being touched it's a cheap cleanup. Some tooling and pre-commit hooks flag missing-final-newline.

> **Suggestion:** Add a single trailing newline after `{% endblock %}`.

<sub>Found by: reviewer-general-purpose · CR-006</sub>

gridcell (commented):
## AI Code Review

**Assessment:** Focused, well-tested change that makes summary-report jobs wait for the `SummaryCacheQueue` to drain before generating, and surfaces a "data may be stale" notice in the email when the wait times out. The design is sound and the tests are clean. Two backend concerns are worth addressing: the SQS visibility timeout is extended only once (one-shot 120s, not renewed) so a long report after an 80s wait could exceed it and be re-delivered; and the DB connection is held open across the sleep loop. Neither is a hard blocker given this runs on a background worker with a documented time budget.

| 0 | 4 | 2 | 2 | 3 |

**Files reviewed:** 6 / 6

### Cross-Cutting Concerns
- The 80s `SUMMARY_CACHE_MAX_WAIT` and the one-shot 120s SQS `visibility_timeout` are coupled by a hand-budgeted ~30s allowance for report generation/upload/email. That invariant lives only in a code comment; if report generation grows, the message can be re-delivered. Consider centralizing or asserting the relationship.

<details>
<summary><strong>CR-007</strong> &middot; Question &middot; <code>correctness</code> — Does the stale-flag test actually need the DB-backed project1 fixture?</summary>

`wait_for_summary_cache` and `create_protocol_report` are both mocked in `test_report_passes_stale_flag_to_email`, and `project1` is only used for `project1.pk`. If nothing reaches the DB, a raw `uuid.uuid4()` would suffice and remove the fixture dependency (and its DB setup cost). If a real `Project` is genuinely required, confirm `django_db` access is in effect for both parametrized variants.

> **Suggestion:** Replace `project1` with a plain UUID if the mocks make the DB unnecessary; otherwise note why the real fixture is needed.

Applies to: `src/api/tests/test_summary_cache_wait.py`

</details>

<details>
<summary><strong>CR-008</strong> &middot; Info &middot; <code>security</code> — Keep wait_for_cache background-only so the blocking loop can't tie up a web worker</summary>

`wait_for_summary_cache` blocks for up to 80s. In this PR it is only enabled (`wait_for_cache=True`) from `create_sample_unit_method_summary_report_background()`, which runs on the SQS worker — correct. As a guardrail, document/assert that `wait_for_cache` must never be set on a synchronous web-request path, where an 80s block would pin a web worker and could be abused to exhaust capacity.

No XSS/injection risk in the change: `project_id__in` is ORM-parameterized and `data_may_be_stale` is a server-side boolean; the pre-existing `{{file_url|safe}}` renders a server-generated presigned URL and is untouched.

Applies to: `src/api/utils/summary_cache.py`, `src/api/utils/reports.py`

</details>

### Praise
- TESTING guard (`if settings.TESTING: return False`) cleanly keeps the 80s blocking loop out of the test suite.
- The `fake_sleep` closure that deletes the queue entry after N calls is a clean way to drive the polling loop without real time.
- The stale-data notice is well-worded and appropriately de-emphasized in both the HTML and text templates.

### Agent Coverage
| reviewer-security | security | 6 | 1 | 0 |
| reviewer-backend-python | backend | 4 | 3 | 0 |
| reviewer-tests | testing | 1 | 3 | 0 |
| reviewer-general-purpose | general | 2 | 1 | 0 |
| reviewer-python-quality | quality | 4 | 0 | 0 |

gridcell (approved):

PR #707, opened 2026-06-22
[M1933] migrate to UniqueConstraint everywhere

saanobhaai:
## Summary by CodeRabbit

* **Chores**
  * Updated database uniqueness rules by moving from legacy composite uniqueness settings to explicit unique constraints across multiple models, including label mappings, observer records, photo quadrats, and summary cache processing.
  * Added a new database migration to apply the constraint changes consistently.


saanobhaai:
[unique_together → UniqueConstraint](https://trello.com/c/zK58I7xz/1933-uniquetogether-%E2%86%92-uniqueconstraint)

saanobhaai:
@coderabbitai review

src/api/models/base.py
```diff
@@ -193,10 +193,6 @@ class AuthUser(BaseModel):
 
     class Meta:
         db_table = "authuser"
-        unique_together = (
-            "profile",
-            "user_id",
-        )
```
JBurkinshaw, line 199: `unique_together = ("profile", "user_id")` is removed from from `AuthUser` and `("profile", "client_id")` from `Application` (and the migration emits `AlterUniqueTogether(..., set())` for both), but unlike the other models, no replacement `UniqueConstraint` is added.

In practice nothing breaks because `user_id` and `client_id` each have field-level `unique=True`, which is stronger than the old profile-scoped composite, so no duplicate can exist. The concern is that this is now a silent, undocumented assumption. If `unique=True` is relaxed on those fields, the per-profile uniqueness guarantee disappears with no warning. Since the ticket says "replace if possible," this is a case where it's possible but was not included.

Can you either add the explicit constraint, or document why field level only was kept with a comment?

src/api/migrations/0125_unique_together_to_unique_constraints.py
```diff
@@ -0,0 +1,150 @@
...
+        migrations.AddConstraint(
+            model_name="obsbenthicphotoquadrat",
+            constraint=models.UniqueConstraint(
+                fields=(
+                    "benthic_photo_quadrat_transect",
+                    "quadrat_number",
+                    "attribute",
+                    "growth_form",
+                ),
+                name="unique_obsbenthicphotoquadrat_pqt_quadrat_attribute_growthform",
```
JBurkinshaw, line 114: I think there's a lock risk here. On large tables reads/writes would be blocked during deploy.

`AddConstraint` with a `UniqueConstraint` builds the unique index non-concurrently, holding an `ACCESS EXCLUSIVE` lock for the duration of the build. On the large observation tables like `obs_benthic_photo_quadrat`, `obs_benthicpit`, and `obs_habitatcomplexity`, it could block all API reads and writes against those tables while the migration runs. The other tables (`observer`, `covariate`, `class_label_mapping`, etc.) are small so likely ok.

If this is a valid concern, we could run this migration during a low-traffic time or split the three large-table constraints into a separate `atomic = False` migration that pre-builds an index with `CREATE UNIQUE INDEX CONCURRENTLY` then attaches it via `ADD CONSTRAINT ... USING INDEX`.

If any of these tables already contains duplicate rows, `AddConstraint` will fail, so a pre-flight dedup check is worth doing (the ticket flags this too: "some will require data deduplication or checking first").

Not blocking the code, but worth thinking about before this hits prod.

JBurkinshaw (changes requested):

src/api/migrations/0125_unique_together_to_unique_constraints.py
saanobhaai: Good eye -- but yeah claude and I did consider this, and I decided the delay would be fine for our users. I think it is very unlikely to cause a block long enough to really impede anybody. I also checked for dups and there are none.

JBurkinshaw (approved):

PR #708, opened 2026-06-22
[M551] system email fixes

saanobhaai:
## Summary by CodeRabbit

* **New Features**
  * Contact form submissions now pass source information through to admin email notifications.
  * Added stronger safeguards to prevent removing or downgrading the last project admin during sync.

* **Bug Fixes**
  * Email admin notifications now send only when there are resolved recipient addresses.
  * Admin email templates now conditionally display the “from” source details.

* **Tests**
  * Expanded coverage for sync push scenarios involving project-admin role changes and protections.


saanobhaai:
[fix "from" text in contact us email from datamermaid.org](https://trello.com/c/svAVKYOn/551-fix-from-text-in-contact-us-email-from-datamermaidorg)

src/api/resources/sync/push.py
```diff
@@ -93,7 +94,22 @@ def apply_changes(request, serializer, record, force=False):
                 Revision.create_from_instance(instance, deleted=True)
                 return 202, "Project has been flagged for deletion", None
 
-            instance.delete()
```
JBurkinshaw, line 96: The last-admin guard here only fires on DELETE. But there's a second way to remove admin status from someone - a sync PUT that changes their role from Admin to Collector/Read-only. If a client sends that kind of update for the project's only admin, the code would take a different path and this check never runs. That would leave the project with no admin.

Is this scenario already blocked somewhere on the update path (e.g. in the serializer or a separate validation)? If not, it might be worth covering here too.

JBurkinshaw (commented):
Looking good. One question in the code

JBurkinshaw (approved):

PR #713, opened 2026-06-24
[M2010] unapproved attribute and obs include property fixes

saanobhaai:
## Summary by CodeRabbit

* **New Features**
  * Belt-invert size validation is now size-bin aware when transect size bins are available.
  * Observation data entry/admin and ingest schemas no longer include the legacy “include” option across affected workflows.

* **Bug Fixes**
  * Belt invert totals and densities now sum all recorded counts (no longer filter by the removed “include” flag).
  * Observation retrieval now skips the extra ownership/status filtering that applied to other actions.

* **Tests**
  * Updated fixtures and assertions to remove “include” coverage; added/adjusted belt-invert validation tests for missing-size scenarios.


saanobhaai:
[unapproved attribute and obs include property fixes](https://trello.com/c/i6xG0LrE/2010-unapproved-attribute-and-obs-include-property-fixes)

src/api/resources/base.py
```diff
@@ -615,10 +615,11 @@ class BaseAttributeApiViewSet(BaseApiViewSet):
...
 
     def get_queryset(self):
-        # GET requests have no authenticators (method_authentication_classes = {"GET": []}),
-        # so is_authenticated is always False on the public REST API — only sync pull (POST)
-        # reaches the created_by branch.
         qs = super().get_queryset()
+        if getattr(self, "action", None) == "retrieve":
+            # UUID lookups bypass the status filter so a user can resolve attributes
+            # referenced in submitted records they have access to but didn't create.
+            return qs
```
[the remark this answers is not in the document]
saanobhaai: Deliberate in this case.

src/api/signals/attributes.py
```diff
@@ -25,6 +27,15 @@
 from ..utils.reports import update_attributes_report
 
 logger = logging.getLogger(__name__)
+
+
+def _schedule_refresh(view_name):
+    if "pytest" in sys.modules:
+        call_command("refresh_view", view_name)
+        return
+    transaction.on_commit(lambda: call_command("refresh_view", view_name))
```
gridcell, line 36: **Warning** · `maintainability` · confidence: high

**Production signal branches on `"pytest" in sys.modules`; the `settings.TESTING` flag conftest sets is unused**

`_schedule_refresh` decides between a synchronous refresh and a `transaction.on_commit` refresh by inspecting `"pytest" in sys.modules`. Meanwhile conftest.py now sets `settings.TESTING = True`, but nothing reads it — the two mechanisms are disconnected, so the new settings flag is effectively dead code. Keying production behavior off interpreter state (whether pytest happens to be imported) is fragile and surprising.

> **Suggestion:** Branch on the Django setting instead: `from django.conf import settings; if getattr(settings, "TESTING", False):`. That makes the conftest assignment meaningful and keeps test-mode detection in configuration rather than interpreter introspection.

> **Why it matters:** Anything that imports pytest in a non-test context would silently switch to synchronous refresh, and the unused `settings.TESTING` line invites confusion about how test mode is actually detected.

<sub>Found by: reviewer-backend-python, reviewer-python-quality, reviewer-tests · CR-001</sub>

src/api/migrations/0127_refresh_mv_invert_attributes.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("api", "0126_invertattributeview"),
+    ]
+
+    operations = [
+        migrations.RunSQL(
+            "REFRESH MATERIALIZED VIEW mv_invert_attributes",
```
gridcell: **Question** · `database` · confidence: medium

**REFRESH MATERIALIZED VIEW (non-CONCURRENTLY) takes an ACCESS EXCLUSIVE lock during deploy**

A plain `REFRESH MATERIALIZED VIEW mv_invert_attributes` acquires an ACCESS EXCLUSIVE lock for the duration of the refresh, blocking all reads of the view until it completes. In a migration run during deploy this is usually acceptable, but on a large view it can briefly stall request paths that read `mv_invert_attributes`.

> **Suggestion:** If the view has (or can get) a unique index, `REFRESH MATERIALIZED VIEW CONCURRENTLY` avoids blocking reads. If a brief block at deploy time is acceptable, leave as-is — just confirm it is intentional.

> **Why it matters:** Unexpected read blocking during deploy can surface as transient errors/timeouts for clients querying invert attributes.

<sub>Found by: reviewer-database · CR-004</sub>

gridcell (commented):
## AI Code Review

**Assessment:** Solid, well-scoped change. The bulk is a clean mechanical removal of the `include` observation flag across models, serializers, parser, SQL aggregation, ingest, and tests, plus a new size-bin-aware invert size validator and a deferred (`on_commit`) materialized-view refresh. No correctness bugs or security vulnerabilities found. Two follow-ups worth addressing: the new size-bin remapping logic has no direct test, and the signal's test-detection branch is disconnected from the `settings.TESTING` flag conftest now sets.

| 0 | 2 | 3 | 1 | 2 |

**Files reviewed:** 23 / 23
**Previously addressed:** 3 finding(s) suppressed (resolved or acknowledged in prior review)

### Cross-Cutting Concerns
- Dropping the `include` flag is a data-semantics change: observations previously recorded with `include=False` are now counted in all aggregations/totals/densities, and migration 0125 removes the column permanently. Confirm this is intended for already-collected production data.
- The reviewed diff was obtained via `gh pr diff` (authoritative) because the local working tree was on a different branch and did not contain these changes. Confirm the merged commit matches the reviewed diff.

### Additional Findings

<details>
<summary><strong>CR-002</strong> &middot; Warning &middot; <code>test_gap</code> — The new size-bin remapping path in InvertSizeValidator is never exercised by a test</summary>

The central new behavior is that `check_invert_size` remaps `comparison_size` to a matching `InvertSize` bin's `min_val` before the `max_length * 1.5` check. The two added tests (`test_belt_invert_partial_size_ok`, `test_belt_invert_no_size_no_bin_ok`) only hit early-return OK paths — neither provides an `InvertSize` row for the transect's `size_bin`, so `InvertSize.objects.filter(invert_bin_size_id=size_bin_id)` returns empty and the remapping loop is never entered. `test_belt_invert_size_exceeds_max_warn` would still pass even if the entire bin-remapping block were deleted.

> **Suggestion:** Add a fixture creating `InvertSize` bins (with realistic `min_val`/`max_val`, including a NULL-bound bin) for the transect's `size_bin`, then assert: (1) a size exceeding max*1.5 on its own but whose bin `min_val` is below threshold → OK; (2) a size whose bin `min_val` is still above threshold → WARN; (3) a size matching no bin → raw size used.

> **Why it matters:** Without coverage, the size-bin logic could regress or be removed silently — and it is the feature this PR exists to add.

Applies to: `src/api/submission/validations/validators/invert_obs.py`, `src/api/tests/test_validations.py`

<sub>Found by: reviewer-tests, reviewer-backend-python</sub>
</details>

<details>
<summary><strong>CR-003</strong> &middot; Question &middot; <code>data_integrity</code> — Previously-excluded observations are now counted; the flag is permanently dropped</summary>

Removing the `WHERE include = TRUE` CTE and the `FILTER (WHERE include = TRUE)` clauses means any observation a user had marked `include=False` is now included in `total_abundance`, density, and all summary outputs. Migration 0125 then drops the column irreversibly, so the prior exclusion intent cannot be recovered. The PR description frames this as intended.

> **Suggestion:** Confirm with the data team that no production records rely on `include=False` to be excluded, and that recomputing affected summaries post-deploy is acceptable. If any audit/rollback need exists, archive the dropped values before 0125 runs.

> **Why it matters:** This silently changes aggregated values for any existing survey that used `include=False`, and the source signal is destroyed by the column drop.

Applies to: `src/api/migrations/0125_remove_include_from_observation_models.py`, `src/api/models/sql_models/beltinvert.py`

<sub>Found by: reviewer-database, reviewer-backend-python</sub>
</details>

<details>
<summary><strong>CR-005</strong> &middot; Question &middot; <code>security</code> — Confirm attribute serializer exposes no per-user-sensitive fields via the unauthenticated retrieve-by-UUID bypass</summary>

The `retrieve` action now returns the unfiltered queryset (all statuses/owners) on a viewset whose GET path has no authenticators. The maintainer confirmed this is intentional for resolving attributes referenced in submitted records — this is a distinct follow-up, not a re-raise of that decision. Since non-approved (PROPOSED) attribute rows are now readable by anyone who knows a UUID, please spot-check that the attribute serializer emits only taxonomic/reference data and nothing user-identifying (e.g. proposer identity beyond an opaque UUID, internal notes).

> **Suggestion:** Review the FishSpecies/InvertAttribute serializer field list; exclude any sensitive field from the public retrieve representation if present. Enumeration risk itself is low (v4 UUID lookups).

> **Why it matters:** An unauthenticated GET-by-UUID that bypasses the status filter is fine for reference data but would be a data-exposure issue if rows embed user-identifying fields.

Applies to: `src/api/resources/base.py`

</details>

<details>
<summary><strong>CR-006</strong> &middot; Info &middot; <code>test_gap</code> — No test covers an uploaded CSV that still contains the removed Include column</summary>

The Include column and its serializer field were removed and the two `include` ingest assertions deleted. There is no test asserting how the ingest pipeline handles an older CSV template that still includes an Include column (rejected vs silently ignored). Clients with cached templates may submit such files.

> **Suggestion:** Add a small ingest test feeding a CSV with a stray Include column and assert the intended behavior (graceful ignore or clear validation error).

> **Why it matters:** Backward-compatibility of CSV ingest is undefined by tests after this change.

Applies to: `src/api/tests/test_ingest.py`, `src/api/ingest/macroinvertebrate_serializer.py`

</details>

<details>
<summary><strong>CR-007</strong> &middot; Praise &middot; <code>authorization</code> — retrieve bypass is correctly scoped and does not weaken write-path authorization</summary>

The status-filter bypass is gated strictly on `action == "retrieve"`, so list/create/update/delete are unaffected. The DELETE/PUT authorization in `AttributeAuthenticatedUserPermission` calls `get_queryset().get(...).status` for update/destroy actions, which still receive the filtered queryset — so the bypass cannot be leveraged to mutate others' attributes.

</details>

<details>
<summary><strong>CR-008</strong> &middot; Praise &middot; <code>testing</code> — New no-size fixture exercises diverse attribute types</summary>

`valid_belt_invert_collect_record_no_size` covers species, genus, class, and two groups-of-interest in one record with `size_bin=None`, giving the no-size/no-bin path breadth across InvertAttribute subtypes rather than a single-species case.

</details>

### Agent Coverage
| reviewer-security | security | 23 | 2 | 0 |
| reviewer-backend-python | backend | 17 | 2 | 0 |
| reviewer-database | database | 4 | 2 | 0 |
| reviewer-tests | testing | 5 | 3 | 0 |
| reviewer-python-quality | quality | 8 | 1 | 0 |

src/api/signals/attributes.py
saanobhaai: Switching to settings.TESTING would reintroduce an ordering bug we fixed previously: Testing.__exit__ resets settings.TESTING = False, so after any test using with Testing(): runs, subsequent tests lose the flag and _schedule_refresh takes the deferred on_commit path. "pytest" in sys.modules is the deliberate choice. The settings.TESTING = True in conftest is also not dead — summary_cache.py reads it.

gridcell (approved):

PR #714, opened 2026-06-24
Enforce SSL for RDS connections

ms280690:
https://trello.com/c/cEujjilH

## Summary by CodeRabbit

* **New Features**
  * Added secure storage for inference container images with vulnerability scanning, immutable tags, and automatic cleanup of unused images.
  * Enforced SSL connections for PostgreSQL databases in development and production environments.

* **Bug Fixes**
  * Simplified database connection settings for local and CI environments by avoiding SSL configuration where it is not required.


src/app/settings.py
```diff
@@ -186,6 +186,27 @@
...
+elif DB_SSLROOTCERT:
+    DB_SSLMODE = "verify-full"
+elif ENVIRONMENT in ("dev", "prod"):
+    DB_SSLMODE = "require"
+else:
+    DB_SSLMODE = "prefer"
+
+_db_options = {"sslmode": DB_SSLMODE}
+if DB_SSLROOTCERT:
+    _db_options["sslrootcert"] = DB_SSLROOTCERT
```
michaelconnor00: This is a lot of flexibility for something we define and can be explicit about. No need to customize via env vars.

Why can the db options not be `"OPTIONS": {"sslmode": "verify-full"}`?

michaelconnor00 (changes requested):

src/app/settings.py
```diff
@@ -186,6 +186,27 @@
...
+#      verifies the server identity against the CA (MITM protection), else
+#   3. "require" in dev/prod (encrypt only), else
+#   4. "prefer" locally/CI (no SSL on those Postgres).
+DB_SSLROOTCERT = os.environ.get("DB_SSLROOTCERT")
+if os.environ.get("DB_SSLMODE"):
+    DB_SSLMODE = os.environ["DB_SSLMODE"]
+elif DB_SSLROOTCERT:
+    DB_SSLMODE = "verify-full"
+elif ENVIRONMENT in ("dev", "prod"):
+    DB_SSLMODE = "require"
```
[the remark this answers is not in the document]
ms280690: skipped in this PR, reason: verify-full is not a settings-only change. It requires (a) shipping the RDS CA bundle into the image, (b) wiring DB_SSLROOTCERT in iac/stacks/api.py container env, (c) settings using it. None exist today — the old verify-full branch was dead code because DB_SSLROOTCERT is set nowhere in iac/. It also can't be validated locally (no TLS on local/CI Postgres), and a wrong cert path means prod cannot reach the DB. rds.force_ssl=1 + require already guarantees encryption; verify-full adds MITM/hostname verification — real hardening, but it deserves its own PR that can be deployed to dev and verified first.

src/app/settings.py
ms280690: 20 lines → 4. No env-var customization, one explicit branch. dev/prod get {"sslmode": "require"}, local/CI get {}. No stray DB_SSLMODE/DB_SSLROOTCERT refs anywhere.

"why not just {"sslmode": "verify-full"}": two blockers, both concrete —
1. Local/CI Postgres serves no TLS, so an unconditional require/verify-full breaks make test and the CI test job. Hence the one env branch (not env-var config).
2. verify-full needs sslrootcert. The Amazon RDS root CA is not in Debian's ca-certificates, and no bundle is shipped in the Dockerfile — so verify-full would fail to connect today.

PR #721, opened 2026-06-26
[M2032] change GoI weighting to equal split

saanobhaai:
## Summary by CodeRabbit

* **Bug Fixes**
  * Updated group-of-interest attribution for family, order, and class to use equal-split weighting across distinct matching names, improving consistency in reported densities.
  * Fixed an edge case where higher-level taxa could be omitted from group-of-interest density results while related abundance totals were still present.

* **Tests**
  * Expanded fixtures and revised expectations to cover equal-split scenarios and families with no genus GoI mapping.
  * Added a new test to verify correct handling when group-of-interest density should be absent.

### DIFF
--- a/src/api/models/sql_models/beltinvert.py
+++ b/src/api/models/sql_models/beltinvert.py
@@ -140,32 +140,35 @@ class BeltInvertSUSQLModel(BaseSUSQLModel):
             {BeltInvertObsSQLModel.sql}
         ),
         goi_weights_by_family AS MATERIALIZED (
-            SELECT ig.family_id, goi.name AS goi_name,
-                COUNT(ig.invertattribute_ptr_id)::float /
-                SUM(COUNT(ig.invertattribute_ptr_id)) OVER (PARTITION BY ig.family_id) AS weight
-            FROM invert_genus ig
-            JOIN invert_group_of_interest goi ON ig.group_of_interest_id = goi.invertattribute_ptr_id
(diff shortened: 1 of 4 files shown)


saanobhaai:
[change GoI density weighting](https://trello.com/c/ovp8OLA3/2032-change-goi-density-weighting)

src/api/models/sql_models/beltinvert.py
```diff
@@ -183,6 +186,10 @@ class BeltInvertSUSQLModel(BaseSUSQLModel):
             JOIN invert_genus g ON o.invert_attribute_id = g.invertattribute_ptr_id
             JOIN invert_group_of_interest goi ON g.group_of_interest_id = goi.invertattribute_ptr_id
             UNION ALL
+            -- NOTE: If a family/order/class taxon has no genera with GoI assignments in the
+            -- taxonomy DB, the JOIN below produces no rows and the observation's count is
+            -- silently absent from density_indha_group_interest while still counted in
+            -- total_abundance and density_indha, making those fields inconsistent.
```
JBurkinshaw, line 192: I like the addition of this note, but I don't think what it describes isn't tested, so it's currently just an worry in a comment. Could we add a fixture with a family-level obs whose genera have no GoI assignment, asserting the count stays in `total_abundance`/`density_indha` but is absent from `density_indha_group_interest`? That makes the current behaviour more intentional (unless you have a plan to change it in the future.

JBurkinshaw (approved):

PR #724, opened 2026-06-30
[M2036] fix invert_size_exceeds_maximum not firing bug

saanobhaai:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved size validation for fish and invert observations so open-ended size bins no longer overwrite the entered size, while bounded bins use a conservative lower bound.
  * Added regression coverage for belt invert records using configured size bins, ensuring oversized observations continue to return a warning as expected.


saanobhaai:
[size validation not returning proper warning when using size bins > 1](https://trello.com/c/gCbDAhg5/2036-size-validation-not-returning-proper-warning-when-using-size-bins-1)

JBurkinshaw (approved):

PR #727, opened 2026-07-02
[M1923] reject ingest rows with unknown Site/MR

saanobhaai:
- CSV ingest (via mermaidr or the API directly) previously let a row through even when its Site/Management name didn't exist in the target project — the row saved with a blank Site/MR instead of failing, contrary to the documented "must be defined before ingestion" contract.
- Adds validate_data__sample_event__site/management to CollectRecordCSVSerializer, mirroring the existing validate_data__observers check, applying to all 7 protocols (fishbelt, benthiclit, benthicpit, benthicpqt, habitatcomplexity, bleaching, macroinvertebrate).
- Fixes a latent whitespace-matching bug in the name-resolution helper (_lower → _normalize, now strips too) so the new strict check doesn't false-reject legitimate names with incidental leading/trailing spaces.
- Error messages now include the offending value; valid-ID lookups are cached per serializer instance (O(1) per row instead of O(n)).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CSV ingest matching so incoming values are trimmed and compared more reliably.
  * When an imported value can’t be matched, it now keeps the original input instead of altering it unexpectedly.
  * Added project-specific validation for sample event site and management values, with clear errors when they don’t exist for the current project.

* **Tests**
  * Added coverage for ingesting records with missing site and management values to verify validation errors are returned.

### DIFF
--- a/src/api/ingest/serializers.py
+++ b/src/api/ingest/serializers.py
@@ -39,14 +39,17 @@ class CollectRecordCSVListSerializer(ListSerializer):
 
     def assign_choices(self, row, choices_sets):
         for name, field in self.child.fields.items():
-            val = field.get_value(row)
+            raw_val = field.get_value(row)
             choices = choices_sets.get(name)
             if choices is None:
                 continue
             try:
-                val = self._lower(val)
-                choices = {label.lower(): value for label, value in choices.items()}
-                row[name] = choices.get(val) or val
+                val = self._normalize(raw_val)
+                choices = {self._normalize(label): value for label, value in choices.items()}
+                # Fall back to the original (unnormalized) value, not the
+                # lowercased/stripped one, so an unresolved value still shows
+                # the user's actual input if it surfaces in an error message.
+                row[name] = choices.get(val, raw_val)
             except (ValueError, TypeError):
                 row[name] = None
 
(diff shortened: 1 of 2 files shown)


saanobhaai:
[Successfully ingested data with Site and MR not in the project](https://trello.com/c/HPrA1GeX/1923-successfully-ingested-data-with-site-and-mr-not-in-the-project)

src/api/ingest/serializers.py
```diff
@@ -335,6 +342,29 @@ def validate_data__observers(self, val):
...
+        if val not in valid_ids:
+            raise ValidationError('{} "{}" does not exist in this project'.format(label, val))
+        return val
+
+    def validate_data__sample_event__site(self, val):
+        return self._validate_project_choice("data__sample_event__site", val, "Site")
+
+    def validate_data__sample_event__management(self, val):
+        return self._validate_project_choice("data__sample_event__management", val, "Management")
+
```
JBurkinshaw, line 367: These live on live on the shared base `CollectRecordCSVSerializer` and apply to all 7 protocols. The added test `test_fishbelt_ingest_missing_site_and_management` only tests FishBelt. The other protocols have no regression test. Would it be worth testing them too or would it just be unnecessary repetition?

JBurkinshaw (approved):
Approved but left one comment

src/api/ingest/serializers.py
saanobhaai: It's always been just fish, and I guess I think that's still enough since we're accumulating a ton of tests (this became a problem on webapp if you recall).

PR #728, opened 2026-07-06
Classifier Schema changes and model registration

gnieuwenhuis:
## Summary
Prepares the mermaid-api `Classifier` record to describe its type, config, and
label set from the model's own `model.json` manifest, and adds
`Classifier.register(version)` to ingest that manifest from S3. **No inference
cutover** — the existing in-process classification path keeps working unchanged.
API-side groundwork for the Lambda classifier lane.

## High-level changes
- **Schema (migration `0134`):** `Classifier` gains `classifier_type` and a
  `config` JSON field; drops the `num_points` and `patch_size` columns
  (`patch_size` moves into `config`, still surfaced flat by the image serializer);
  `version` is now unique.
- **`register(version)`:** reads the version's `model.json` from S3, validates
  `config` against a per-type pydantic schema, resolves each `ba_uuid::gf_uuid`
  into the BA+GF label set, and upserts the classifier atomically (a malformed
  manifest raises and writes nothing). Replaces the manual CSV import.
- **`register_classifier` management command** replaces `set_classifier_bagfs`.
- **`num_points`** now comes from a new `INFERENCE_DEFAULT_NUM_POINTS` setting
  (= 25) instead of the model — behavior unchanged.
- **Classifier read endpoint** exposes `classifier_type`/`config` (no `num_points`).

## Deploying this — read before merging

`0134` drops two columns that the currently-deployed code still declares, and
migrations run at container start while old tasks are still serving
(`min_healthy_percent=100`). **For the length of the rolling deploy, every
`Classifier` query from an old task raises `ProgrammingError`.** That is:

- `GET /v1/classification/classifiers/` (and `/latest/`) — unauthenticated
- the `assign_classifier` pre-save signal on `CollectRecord`
- the image worker's `_get_classifier_and_weights`
- the BenthicPQT annotation validators, so image-classification collect records
  cannot be validated or submitted
- image detail responses, via `ImageSerializer.get_patch_size`
- the Django admin classifier list

This is accepted rather than overlooked. An expand/contract split across two
releases was implemented and then reverted: it bought a few minutes of uptime at
the cost of a `SeparateDatabaseAndState` block, a nullable-column concession and
a follow-up release to chase. Everything that breaks is inside the
image-classification workflow, the window is bounded by the deploy, and it
recovers on its own. No data is lost — `settings.py` sets no `ATOMIC_REQUESTS`,
so a failing record in a `/v1/push/` batch does not poison its siblings.

One thing to watch: `ApiService` runs with
`DeploymentCircuitBreaker(rollback=True)`. If the new task fails for an unrelated
reason after the migration commits, ECS reverts to old code against the new
schema and stays there. Recovery is `manage.py migrate api 0133`, which this
migration supports — the reverse re-adds both columns while they are still
nullable, backfills `patch_size` from `config`, then restores `NOT NULL`.

Two preconditions guard the round trip and raise rather than guessing: a
`num_points` that is not the field default cannot be recovered from anywhere, and
a row whose `config` carries no `patch_size` cannot be restored at all. Both name
the offending versions. Dev and prod each hold a single row at `num_points=25`,
so neither fires today.

### If you already migrated this branch locally

`0134` shipped on this branch between 2026-09-10 and today in a form that kept
both columns. Django records applied migrations by name, so on a database
migrated from that commit the new `RemoveField` operations are a silent no-op —
`migrate` reports nothing to do, the columns are still there, and
`makemigrations --check` cannot see the divergence. It only surfaces on reverse,
as `column "patch_size" of relation "class_classifier" already exists`.

Either `make dbrestore`, or `python manage.py migrate api 0133 --fake` and drop
the two columns by hand, before running `migrate` again. Dev and prod are
unaffected: this branch has never been merged to `dev`.

Closes: https://github.com/data-mermaid/mermaid-classifier/issues/52

## Summary by CodeRabbit

- **New Features**
  - Classifier metadata now supports classifier types and validated configuration.
  - Added classifier registration from model manifests, including dry-run validation.
  - Added admin filtering by classifier type.
  - Added configurable default inference point counts.

- **Bug Fixes**
  - Improved classifier and label validation, with safer handling of malformed registrations.
  - Reused annotation rows during validation runs for more consistent processing.

- **Changes**
  - Classifier versions are now unique.
  - Classifier responses expose configuration and type details; legacy point-count fields are no longer exposed.


src/api/admin/classification.py
```diff
@@ -49,7 +49,7 @@ class BenthicAttributeGrowthFormAdmin(BaseAdmin):
 
 @admin.register(Classifier)
 class ClassifierAdmin(BaseAdmin):
-    list_display = ["version", "name", "patch_size", "num_points"]
+    list_display = ["version", "name", "classifier_type", "patch_size"]
```
saanobhaai, line 52: Might as well add `list_filter = ["classifier_type"]`

src/api/models/classification.py
```diff
@@ -123,10 +159,90 @@ class Classifier(BaseModel):
...
+
+        Validates config via the per-type pydantic schema and resolves each
+        `ba_uuid::gf_uuid` class into the BA+GF M2M. Raises ClassifierRegistrationError
+        on any malformed/mismatched manifest, applying nothing.
+        """
+        key = f"classifier/{version}/model.json"
+        manifest = s3.read_json_object(settings.AWS_CONFIG_BUCKET, key)
+
+        schema_version = manifest.get("schema_version")
+        if schema_version != SUPPORTED_MANIFEST_SCHEMA_VERSION:
```
saanobhaai, line 182: So only one `SUPPORTED_MANIFEST_SCHEMA_VERSION` at a time -- reasonable, just checking on intended flow when we update. Would we handle any cases of needing to re-register with an updated schema?

src/api/migrations/0130_classifier_config.py
```diff
@@ -0,0 +1,60 @@
...
+                max_length=20,
+            ),
+        ),
+        migrations.AddField(
+            model_name="classifier",
+            name="config",
+            field=models.JSONField(blank=True, default=dict),
+        ),
+        migrations.RunPython(patch_size_to_config, config_to_patch_size),
+        migrations.RemoveField(model_name="classifier", name="patch_size"),
```
saanobhaai, line 49: Reversing RemoveField(patch_size) re-adds patch_size as NOT NULL with no default, and that happens before the RunPython reverse function backfills it

src/api/models/classification.py
```diff
@@ -123,10 +159,90 @@ class Classifier(BaseModel):
...
+    @classmethod
+    def register(cls, version, *, name=None, description=None):
+        """Ingest s3://<config-bucket>/classifier/<version>/model.json into this row.
+
+        Validates config via the per-type pydantic schema and resolves each
+        `ba_uuid::gf_uuid` class into the BA+GF M2M. Raises ClassifierRegistrationError
+        on any malformed/mismatched manifest, applying nothing.
+        """
+        key = f"classifier/{version}/model.json"
+        manifest = s3.read_json_object(settings.AWS_CONFIG_BUCKET, key)
```
saanobhaai, line 179: Should we catch read errors here?

src/api/models/classification.py
```diff
@@ -123,10 +159,90 @@ class Classifier(BaseModel):
...
+            if not created:
+                classifier.classifier_type = classifier_type
+                classifier.config = config
+                if name is not None:
+                    classifier.name = name
+                if description is not None:
+                    classifier.description = description
+                classifier.save()
+
+            classifier.benthic_attribute_growth_forms.set(resolved)
```
saanobhaai, line 242: This is unconditional... would wipe all BA/GFs if the manifest's `classes` list is empty or missing

src/api/models/classification.py
```diff
@@ -123,10 +159,90 @@ class Classifier(BaseModel):
     class Meta:
         db_table = "class_classifier"
 
+    @property
+    def patch_size(self):
```
saanobhaai, line 163: `patch_size` changed from a required, non-null `IntegerField` to a property that can return `None`, but nothing outside `register()` enforces that config actually contains a `patch_size` — breaking the field's previous non-null guarantee for API consumers

src/api/migrations/0130_classifier_config.py
```diff
@@ -0,0 +1,60 @@
...
+        config = dict(classifier.config or {})
+        config["patch_size"] = classifier.patch_size
+        classifier.config = config
+        classifier.save(update_fields=["config"])
+
+
+def config_to_patch_size(apps, schema_editor):
+    Classifier = apps.get_model("api", "Classifier")
+    for classifier in Classifier.objects.all():
+        classifier.patch_size = (classifier.config or {}).get("patch_size") or 0
```
saanobhaai, line 24: fallback fabricates a bogus patch_size=0 instead of failing loudly when config lacks `patch_size`, corrupting data on migration rollback

src/api/models/classification.py
```diff
@@ -123,10 +159,90 @@ class Classifier(BaseModel):
...
+            raise ClassifierRegistrationError(
+                f"Unsupported model.json schema_version {schema_version!r} for {version}"
+            )
+
+        task = manifest.get("task")
+        classifier_type = TASK_TO_CLASSIFIER_TYPE.get(task)
+        if classifier_type is None:
+            raise ClassifierRegistrationError(f"Unknown task {task!r} in model.json for {version}")
+
+        config_schema = CONFIG_SCHEMAS[classifier_type]
```
saanobhaai, line 192: direct dict index outside the try/except, so a `classifier_type` present in `TASK_TO_CLASSIFIER_TYPE` but missing from `CONFIG_SCHEMAS` raises an uncaught `KeyError` instead of `ClassifierRegistrationError`

src/api/resources/classification/classifier.py
```diff
@@ -26,7 +26,17 @@ class ClassifierSerializer(BaseAPISerializer):
...
+            "id",
+            "name",
+            "version",
+            "classifier_type",
+            "config",
+            "description",
+            "benthic_attribute_growth_forms",
+            "created_on",
+            "updated_on",
+        ]
```
saanobhaai, line 39: This (silently) drops `updated_by` and `updated_by` which have been present... Could be a good idea I guess.

src/api/models/classification.py
```diff
@@ -123,10 +159,90 @@ class Classifier(BaseModel):
...
+        except Exception as e:
+            raise ClassifierRegistrationError(
+                f"Invalid config in model.json for {version}: {e}"
+            ) from e
+        config = validated_config.model_dump()
+
+        with transaction.atomic():
+            resolved = []
+            for label in manifest.get("classes", []):
+                ba_uuid, _, gf_uuid = label.partition("::")
```
saanobhaai, line 204: Reimplements the existing "ba_id::gf_id" label-parsing convention with different edge-case semantics than the established implementation in utils/classification.py, creating two divergent parsers for the same string format

src/api/models/classification.py
```diff
@@ -123,10 +159,90 @@ class Classifier(BaseModel):
...
+                try:
```
saanobhaai, line 205: These two try/except blocks could potentially be DRYed up, but may not be worth it for readability

saanobhaai (changes requested):
Looks great! In addition to inline comments, I'd just say there are a lot of tests and am not sure they're all necessary (I didn't look at those in much detail though).

src/api/admin/classification.py
gnieuwenhuis: Fixed in 1b71f1b0 — `ClassifierAdmin` now carries `list_filter = ["classifier_type"]`.

src/api/models/classification.py
gnieuwenhuis: `register()` is an idempotent upsert keyed on `version`, so updating the schema is just re-running `register_classifier <version>` — it overwrites `classifier_type`, `config` and the BA+GF set on the existing row.

- Nothing persists `schema_version` on the `Classifier` row, so bumping `SUPPORTED_MANIFEST_SCHEMA_VERSION` doesn't invalidate what's already registered — it only means a manifest still written at the old version can't be re-ingested until it's regenerated.
- That's the intended coupling: the manifest is the source of truth and lives next to the model in S3, so a schema bump is a model-side republish followed by a re-register, not a data migration here.

src/api/migrations/0134_classifier_config.py
```diff
@@ -0,0 +1,60 @@
...
+                max_length=20,
+            ),
+        ),
+        migrations.AddField(
+            model_name="classifier",
+            name="config",
+            field=models.JSONField(blank=True, default=dict),
+        ),
+        migrations.RunPython(patch_size_to_config, config_to_patch_size),
+        migrations.RemoveField(model_name="classifier", name="patch_size"),
```
gnieuwenhuis: Fixed in d268a6dd — `0134` no longer drops the column, so the reverse cannot re-add it as NOT NULL.

- `patch_size` was an `IntegerField` with no default, so the reverse `AddField` applied `NOT NULL` before the `RunPython` had backfilled anything. Against a database with a single `Classifier` row, `migrate api 0133` aborted with `IntegrityError: column "patch_size" of relation "class_classifier" contains null values`.
- The two `RemoveField`s are now `state_operations` inside `SeparateDatabaseAndState` with empty `database_operations`: the fields leave Django's state, the columns stay, and `sqlmigrate` renders the operation as `(no-op)`. There is no reverse `AddField` left to fail.
- The column drop moves to its own release — see the PR comment for why it could not stay here.

src/api/models/classification.py
gnieuwenhuis: Fixed in 82bde9cb — the read is wrapped, so a missing or unreadable `model.json` and a non-JSON body both surface as `ClassifierRegistrationError` with the original chained.

That is what the docstring already claimed, so this was a real gap rather than a style point.

src/api/models/classification.py
gnieuwenhuis: Fixed in 82bde9cb — `register()` now rejects an empty or missing `classes` list before it reaches the M2M.

- The test covers the consequence rather than the raise: it registers a classifier with real labels, attempts an empty re-register, and asserts the label set is still intact. A test that only asserted the exception would have passed even with the wipe still happening.

src/api/models/classification.py
gnieuwenhuis: Fixed in d268a6dd — the guarantee is enforced by the database again, not by `clean()`.

- `Classifier.Meta.constraints` carries a `CheckConstraint` requiring `config` to hold a `patch_size` key when `classifier_type == "pyspacer"`, imposing nothing on other types: `CHECK ((NOT ("classifier_type" = 'pyspacer') OR "config" ? 'patch_size'))`.
- Presence is all a constraint can assert; `PyspacerConfig` still enforces the type. Direct ORM construction is now rejected by Postgres rather than slipping through — which `clean()` could not achieve, since `BaseModel` has no `save()` override and so it never runs on a plain `save()`.

src/api/migrations/0134_classifier_config.py
```diff
@@ -0,0 +1,60 @@
...
+        config = dict(classifier.config or {})
+        config["patch_size"] = classifier.patch_size
+        classifier.config = config
+        classifier.save(update_fields=["config"])
+
+
+def config_to_patch_size(apps, schema_editor):
+    Classifier = apps.get_model("api", "Classifier")
+    for classifier in Classifier.objects.all():
+        classifier.patch_size = (classifier.config or {}).get("patch_size") or 0
```
gnieuwenhuis: Fixed in d268a6dd — with the column retained, the reverse has nothing to fabricate.

- It still raises, naming the offending version, when `config` has no `patch_size`, and now also fills `num_points` where it is NULL — the only way a row reaches the reverse without one.
- Existing `num_points` values survive untouched, because the column is never dropped.

src/api/models/classification.py
gnieuwenhuis: Fixed in 82bde9cb — `CONFIG_SCHEMAS.get()` with an explicit `ClassifierRegistrationError` when a type has no schema.

Worth noting this is live the moment `segmentation` gets a task mapping, since it is already a declared `classifier_type` with no schema entry.

src/api/resources/classification/classifier.py
gnieuwenhuis: Fixed in 1b71f1b0 — both are back on the field list. Dropping them made sense in the moment, but was not asked for; the PR was not meant to change that part of the response shape.

src/api/models/classification.py
gnieuwenhuis: Fixed in 7d10099b — extracted `parse_bagf_label` and pointed both call sites at it.

- The divergence was real: `(label.split("::", 1) + [None])[:2]` turns `"ba::"` into `growth_form_id=""` where `partition()` yields `None`. Worth being accurate about the impact though — that difference was never observable, because `ForeignKey.get_db_prep_save` maps `""` to `None` for a UUID target, so both spellings already wrote SQL NULL. This unifies the semantics rather than fixing a live defect.
- They are the same strings — `response_message.classes` at inference time is the same class list `model.json` carries — so having one parser is the point.
- Grepped for other `"::"` parsing in non-test code; those two were the only sites.

src/api/models/classification.py
gnieuwenhuis: Fixed in 7d10099b — folded into a single `_resolve_label_part(model, pk, label, kind)`.

You were right that it was marginal on its own. Extracting the shared label parser in the same commit is what tipped it, since both blocks then reduced to one call each.

gnieuwenhuis:

Answered all 11 inline threads plus the three body-only findings: 12 accepted, 1 pushed back on, 3 answered. Range c9b1f02...d268a6dd. The full suite passes at 83.66% coverage, and the migration round trip was run both directions against a populated `class_classifier`.

A later pass reshaped migration `0134` and revised three of those answers — the two reverse-path threads and the `patch_size` guarantee thread. `0134` was dropping `patch_size` and `num_points`, which is unsafe on this service: `ApiService` has no `command=` override, so it migrates through `docker-entry.sh` at container start, and `min_healthy_percent=100` keeps the old task serving until the new one is healthy. The columns therefore disappeared while code that still declared them was live, breaking `GET /v1/classification/classifiers/`, the image endpoints, collect-record saves, BenthicPQT validation and the classify worker for the length of the rollout; `circuit_breaker(rollback=True)` would then revert the code but not the schema. The queue workers never migrate at all, since they use a `command=` override.

**Not covered by an inline reply**

- `0134` is now the additive half of an expand/contract pair: both columns stay, both become nullable, and the fields are removed from Django's state only. Dropping them moves to its own release, tracked in data-mermaid/mermaid-classifier#98 (d268a6dd).
- CodeRabbit's `classes`-must-be-a-nonempty-array finding is accepted and fixed (d268a6dd) — it was the one finding from that review still open. Iterating a JSON object yields its keys, which are strings, so the per-element check could not distinguish a mapping of valid `ba_uuid::gf_uuid` keys from a list.
- Verified: `makemigrations --check --dry-run` reports nothing pending, and `sqlmigrate` renders the state-only removal as `(no-op)` and the new `CheckConstraint` with its literals inlined.
- Still needing a human before merge: the forward/reverse rehearsal against a restored prod snapshot, since CI runs `--no-migrations` and that is the only evidence; and confirmation that no webapp or third-party consumer reads top-level `patch_size`/`num_points` from `GET /v1/classification/classifiers/`, which this PR drops.
- Trimmed six tests that only an intentional decision could break — `test_classifier_config_schema.py` in full, plus the constant check and two default/property checks (30af7bc1). A follow-up review caught that this went one test too far: `extra="forbid"` on `PyspacerConfig` was left unguarded, and `test_register_rejects_invalid_config` passes with or without it. Re-covered behaviourally in 74ce15b1 — a manifest carrying a key outside the config schema now has to be rejected rather than silently dropped from the stored `config`.
- CodeRabbit's suggestion of a direct test for `read_json_object`: not taken. It forwards to `get_object` and calls `json.loads`, so a test there would assert the standard library works against a mocked body; it is already exercised through `Classifier.register()`.
- CodeRabbit's `num_points_per_quadrat` fixture finding was already fixed — that review ran against ce43940, and c9b1f02 set it explicitly.
- `register()` and `clean()` also had raw `TypeError`/`AttributeError` escaping for a manifest or `config` that parsed as valid JSON but not an object (74ce15b1). Same class of gap as the inline findings, so it was fixed alongside them.

src/api/models/classification.py
```diff
@@ -123,10 +180,113 @@ class Classifier(BaseModel):
...
+            validated_config = config_schema(**(manifest.get("config") or {}))
+        except Exception as e:
+            raise ClassifierRegistrationError(
+                f"Invalid config in model.json for {version}: {e}"
+            ) from e
+        config = validated_config.model_dump()
+
+        classes = manifest.get("classes") or []
+        if not classes:
+            raise ClassifierRegistrationError(f"model.json for {version} has no classes")
```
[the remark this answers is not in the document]
gnieuwenhuis: Fixed in d268a6dd — `classes` is rejected unless it is a list.

- The per-element `isinstance(label, str)` check could not catch a JSON object: iterating one yields its keys, which are strings, so a mapping of valid `ba_uuid::gf_uuid` keys would have registered a label set.
- Covered by a test that fails when the guard is removed.

src/api/models/classification.py
```diff
@@ -109,24 +159,147 @@ def __str__(self):
...
 
+    @classmethod
+    def register(cls, version, *, name=None, description=None):
+        """Ingest s3://<config-bucket>/classifier/<version>/model.json into this row.
+
+        Validates config via the per-type pydantic schema and resolves each
+        `ba_uuid::gf_uuid` class into the BA+GF M2M. Raises ClassifierRegistrationError
+        on any malformed/mismatched manifest, applying nothing.
+        """
+        key = f"classifier/{version}/model.json"
```
saanobhaai: This should use CLASSIFIER_CONFIG_S3_PATH rather than being hardcoded, I think (utils/classification.py).

src/api/models/classification.py
```diff
@@ -109,24 +159,147 @@ def __str__(self):
...
+    def clean(self):
+        super().clean()
+        config_schema = CONFIG_SCHEMAS.get(self.classifier_type)
+        if config_schema is None:
+            return
+        config = self.config or {}
+        if not isinstance(config, dict):
+            raise ValidationError({"config": "config must be a JSON object."})
+        try:
+            config_schema(**config)
```
saanobhaai, line 208: Claude points out that pydantic does lax coercion by default, so `PyspacerConfig(patch_size="224")` doesn't raise, but silently produces `patch_size=224` (an int) on the returned model instance. clean() never captures that instance, so `self.config["patch_size"]` is still the string "224" that gets written to the DB by save(). register() doesn't have this problem. Could address in one of two ways:

- Reassign (self.config = config_schema(**config).model_dump()) — accepts "224", silently normalizes it to 224 before saving. Mirrors what register() already does, so admin-edited and S3-manifest-registered classifiers end up byte-for-byte consistent in the DB.
- Strict validation (model_config = ConfigDict(extra="forbid", strict=True) on PyspacerConfig) — rejects "224" outright with a validation error in the admin form, since CONFIG_SCHEMAS is shared between clean() and register(), so this tightens both paths identically.

src/app/settings.py
```diff
@@ -458,6 +458,7 @@ def _sentry_before_send(event, hint):
 AWS_QUERYSTRING_AUTH = False
 AUTOCONFIRM_THRESHOLD = 1.0
 CLASSIFIED_THRESHOLD = 0.5
+INFERENCE_DEFAULT_NUM_POINTS = int(os.environ.get("INFERENCE_DEFAULT_NUM_POINTS") or 25)
```
saanobhaai: INFERENCE_DEFAULT_NUM_POINTS is read from os.environ but isn't added to DjangoSettings or the ECS environment dict, so it can't actually be overridden per-environment despite that being its apparent purpose.

src/api/migrations/0134_classifier_config.py
```diff
@@ -0,0 +1,100 @@
...
+        ),
+        migrations.AddField(
+            model_name="classifier",
+            name="config",
+            field=models.JSONField(blank=True, default=dict),
+        ),
+        # patch_size and num_points remain in the database: code that declares them must
+        # keep reading them while two versions of the app run side by side. They are
+        # nullable because the model does not declare them and supplies no value on
+        # insert. Dropping the columns is a separate release (mermaid-classifier#98).
```
saanobhaai: How will we remember to follow up on this?

src/tools/management/commands/register_classifier.py
```diff
@@ -0,0 +1,49 @@
+from django.core.management.base import BaseCommand
+from django.db import transaction
+
+from api.models import Classifier
+
+
+class Command(BaseCommand):
```
saanobhaai, line 7: Just thinking about the QA stage, on dev -- it might be helpful to generate a list of things to try (external to this codebase, but based on this PR), including one or two examples of running this command, on dev.

saanobhaai (changes requested):
Looking good -- just a few outstanding questions...

src/api/migrations/0134_classifier_config.py
gnieuwenhuis: After our talk this morning I think I'm going to drop this change entirely - it was meant to avoid downtime during deploy, but it only adds complexity, and I think the short downtime for classification during the deployment is not worth this complexity.

src/api/models/classification.py
gnieuwenhuis: Fixed in 07db03e6 — both call sites share the constant now.

- Moved rather than imported: `utils/classification.py` already imports `get_image_storage_config` and `parse_bagf_label` from `..models.classification`, so pulling the constant the other way is a cycle. It now sits in `models/classification.py` beside `SUPPORTED_MANIFEST_SCHEMA_VERSION`, and `utils/classification.py` imports it — `_fetch_and_cache_classifier_config` is untouched.
- Covered by a test asserting `register()` reads `AWS_CONFIG_BUCKET` at `classifier/<version>/model.json`. The expected key is a hand-written literal rather than one built from the constant, so it pins the S3 layout instead of restating whatever the constant happens to say.

src/api/models/classification.py
gnieuwenhuis: Fixed in bae76a06 — took the second option, strict validation.

- `PyspacerConfig` now carries `ConfigDict(extra="forbid", strict=True)`. A quoted `patch_size` is rejected in the admin form instead of normalized, and because `CONFIG_SCHEMAS` is shared, `register()` rejects a manifest carrying one too.
- Chose rejection over the reassign because `clean()` mutating the instance it was handed is a surprising side effect to leave for the next reader, and a manifest with a quoted `patch_size` is a producer bug worth failing on rather than absorbing.

src/app/settings.py
gnieuwenhuis: Fixed in 77387904 — the `os.environ` read is gone; it is `INFERENCE_DEFAULT_NUM_POINTS = 25`.

- You were right that it could not be overridden. It is absent from the `DjangoSettings` dataclass, from the ECS `environment` dict in `iac/stacks/api.py`, from `.env.sample`, and from every workflow.
- Went with the literal rather than wiring it through, makes more sense for clarity and maintainibility, and there was no need for a per-environment override.

src/api/migrations/0134_classifier_config.py
gnieuwenhuis: Settled in 164c3892 — there is nothing left to remember, because the deferred drop is gone. `0134` drops both columns itself.

src/tools/management/commands/register_classifier.py
gnieuwenhuis: Good call — opened #766 as a place we can record things for QA in dev.

saanobhaai (approved):

saanobhaai:
[api changes to support classifier abstraction](https://trello.com/c/iqBiszXf/2107-api-changes-to-support-classifier-abstraction)

PR #730, opened 2026-07-07
Classifier Lambda lane behind a toggle (dev cutover)

gnieuwenhuis:
## Summary
Adds the ability to run image classification in a separate, on-demand cloud function (AWS Lambda) instead of inside the main API — controlled by a single setting and instantly reversible. With the setting off, classification behaves exactly as it does today. This is the "turn it on in dev" step; the old in-process path is **not** removed.

## High-level changes
- **Reversible toggle:** a setting selects either the new cloud-function path or the existing in-process path (empty setting = today's behavior, a safe fallback).
- **Cloud-function lane:** the API sends the image location and its sample points to the function and records the returned classifications the same way as before — results are identical to the in-process path.
- **Points-per-image is now a per-upload option** (defaults to 25, unchanged from today).
- **Classifier info endpoint:** can be looked up by version and now flags which version is the active default.
- **Infrastructure:** grants the background worker permission to call the function, and passes the function's name and active version to it.

## Notes
- The model version is not chosen per request; the active version is a single setting kept in sync with the deployed function.
- Stacked on #728 (merges into that branch).

Closes: https://github.com/data-mermaid/mermaid-classifier/issues/54


gnieuwenhuis:
Closing this because we decided to deploy the Beta model directly, without a toggle, simplifying what we started here.

PR #731, opened 2026-07-07
[M2040] allow 0 for number_of_solutions_supported_by

saanobhaai:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved finance-solution validation so certain solution types can now accept a zero value without errors.
  * Kept blank or missing values invalid where required, while allowing valid zero values to be saved.
  * Updated data checks to better flag cases where the value should be zero for other solution types.


saanobhaai:
['Number of solutions supported by' field does not accept zero as a value in 'Technical assistance facility', 'Conservation trust fund' and 'Financial facility' in GFCR page](https://trello.com/c/1tVnPH6Z/2040-bug-number-of-solutions-supported-by-field-does-not-accept-zero-as-a-value-in-technical-assistance-facility-conservation-trust-f)

JBurkinshaw (approved):
Looks good to me

PR #734, opened 2026-07-08
[M2002] GFCR phase 4

saanobhaai:
## Summary by CodeRabbit

- **New Features**
  - Added support for programmatic co-financing finance solutions.
  - Added stricter validation for finance-solution types, indicator-set titles, sustainable-finance mechanisms, investment types, and revenue relationships.
  - Added a `--no-restore` option to audit the current database without restoring production data.

- **Bug Fixes**
  - Prevented unsupported solution counts and invalid field combinations from being accepted.
  - Standardized indicator-set titles and rejected deprecated choices.
  - Made finance-solution type selection required.


saanobhaai:
[GFCR API phase 4](https://trello.com/c/ArT1q3yX/2002-gfcr-api-phase-4)

saanobhaai:
@coderabbitai review

src/api/tests/test_gfcr.py
JBurkinshaw, line 168: With the changes made, I don't think this testing anything any more. The fixture and the PUT payload now both use `title="Baseline"`, so the "update" is a same-value no-op. Previously it changed the title to a distinct value. As written it would pass even if the update path silently dropped `title`.
To fix you could set the payload to a different valid report title like `"Mid-year report"`.

src/api/migrations/0130_gfcr_phase4.py
```diff
@@ -0,0 +1,245 @@
+import django.contrib.postgres.fields
+from django.db import migrations, models
+
+import api.models.gfcr
+
+
+def assert_phase4_preconditions(apps, schema_editor):
```
JBurkinshaw, line 7: Should this be tested?
The SQL checks are the condition for a big schema change, but none are tested. An SQL bug wouldn't appear until run against real data. Worth a test that seeds a violating row per check and asserts `RuntimeError`, plus a clean-data pass.

JBurkinshaw (commented):
Looks good and thorough.
One thing that predates this PR (not a blocker) - the PCF-with-revenues rule is only enforced by the migration precondition and the audit script, not on writes.
After this change is applied, I think a client could still POST a revenue against a `programmatic_co_financing` solution and quietly re-introduce it. A fix could go in `IndicatorSetViewSet._save` (gfcr.py, just after `fin_sol_record` is saved) since the serializer can't see revenues there.
We're likely handling this in Collect though, so we could call it out of scope.

JBurkinshaw (approved):
Looks good. The comments I made have been addressed

saanobhaai:
@coderabbitai review

PR #740, opened 2026-07-16
[M2050] add depth_avg/depth_sd to summary SE views

saanobhaai:
## Summary by CodeRabbit

* **New Features**
  * Added average depth and depth standard deviation to sample event and project summary results.
  * Included the new depth statistics in CSV and report exports.

* **Bug Fixes**
  * Improved depth aggregation across observations for more accurate summary results.

* **Tests**
  * Added coverage validating calculated depth averages and standard deviations.


saanobhaai:
[add depth_avg/depth_sd to summary SE views](https://trello.com/c/XzvKicJu/2050-add-depthavg-depthsd-to-summary-se-views)

saanobhaai:
@coderabbitai review

src/api/models/summary_sample_events.py
```diff
@@ -47,6 +47,16 @@ class SummarySampleEventBaseModel(models.Model):
...
+    depth_avg = models.DecimalField(
+        max_digits=4, decimal_places=2, null=True, blank=True, verbose_name="depth mean (m)"
+    )
+    depth_sd = models.DecimalField(
+        max_digits=4,
+        decimal_places=2,
+        null=True,
+        blank=True,
+        verbose_name="depth standard deviation (m)",
+    )
```
JBurkinshaw, line 59: The ticket says this is SQL-view-backed so no migration needed. `depth_avg`/`depth_sd` has been added to the abstract `SummarySampleEventBaseModel`, which `SummarySampleEventModel` (the managed `summary_sample_event` table) also inherits. That's the table the viewset reads and `summary_cache` writes to. Don't we need a migration adding these two columns? Tests wouldn't catch it since they run `--no-migrations`.

JBurkinshaw (commented):
The project-level aggregate (`_update_project_summary_sample_events` building `BaseProjectSummarySampleEvent.records`) - does that need depth too, or is it out of scope here? Ticket only calls out the per-SE view, but `mermaid_get_summary_sampleevents()` consumers might expect it on both.

saanobhaai:
Re: project-level agg depth: _update_project_summary_sample_events() (summary_cache.py:206) builds records via SummarySampleEventSerializer(qs, many=True).data — the exact same serializer used by the per-SE viewset (resources/summary_sample_event.py:28, imported at summary_cache.py:70). That serializer is exclude-based (exclude = ["project_status", "sample_event_notes"], resources/base.py:171), not an explicit field whitelist, so any new field on SummarySampleEventBaseModel — including depth_avg/depth_sd — flows into both outputs automatically. Nothing to fix, but did add a test guarding against that regressing.

JBurkinshaw (approved):

PR #744, opened 2026-07-22
[M2060] fix image/status delete race condition

saanobhaai:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved image deletion reliability, including correct “not found” behavior when an image is removed concurrently.
  * Project deletion now also cleans up associated quadrat images and ensures their storage cleanup is scheduled.
  * Standardized the validation error code for cases where surveyed length is non-positive.
* **Tests**
  * Added regression coverage for project deletion cleaning up quadrat images.
  * Added coverage for image delete returning 404 when deletion happens concurrently.


saanobhaai:
[image delete race condition bug](https://trello.com/c/UeEdClan/2060-image-delete-race-condition-bug)

src/api/utils/project.py
```diff
@@ -1103,7 +1103,13 @@ def delete_project(pk):
         with transaction.atomic():
             sid = transaction.savepoint()
             try:
+                # Collect PQT image IDs before the cascade-delete removes the
+                # ObsBenthicPhotoQuadrat rows that link back to them.
+                pqt_image_ids = collect_project_pqt_image_ids(instance)
                 delete_instance_and_related_objects(instance)
+                # Now that ObsBenthicPhotoQuadrat rows are gone (PROTECT lifted),
+                # delete the orphaned Image records and schedule S3 cleanup.
+                delete_collected_pqt_images(pqt_image_ids)
```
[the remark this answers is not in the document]
saanobhaai: Images are never referenced by observations outside the project.

JBurkinshaw, line 1112: `delete_collected_pqt_images` logs `"Failed to delete S3 file during demo project cleanup"` (`utils/project.py:408`). This function now runs on every project deletion, not just demo cleanup, so the message will not be accurate failures. Worth generalizing to "…during project deletion".

src/api/tests/test_image.py
```diff
@@ -269,6 +270,30 @@ def test_upload_image_exceeding_size_limit(
     assert "size limit" in response.json()["error"]
 
 
+def test_destroy_returns_404_when_image_deleted_concurrently(
```
JBurkinshaw: This test only covers `Image.DoesNotExist → NotFound` (deletion before the locked re-fetch). I don't think it verifies that a concurrent `ClassificationStatus` insert no longer breaks the delete.
If that the concurrency is hard to test, you could update the test name to reflect that

JBurkinshaw (changes requested):

JBurkinshaw (approved):

PR #746, opened 2026-07-22
[M2054] Add duplicate checks to Site/MR creation

saanobhaai:
## Summary by CodeRabbit

* **New Features**
  * Added duplicate detection when creating sites and management records.
  * Duplicate warnings include an error code and matching record IDs.
  * Added an option to bypass duplicate warnings during creation.
  * Duplicate checks apply only to new records, not updates.

* **Bug Fixes**
  * Improved sync error handling for nested validation responses.
  * Standardized duplicate detection across API and submission workflows.


saanobhaai:
[Add duplicate checks to Site/MR creation](https://trello.com/c/5EQmAJIS/2054-add-duplicate-checks-to-site-mr-creation)

src/api/utils/duplicates.py
```diff
@@ -0,0 +1,121 @@
...
+    )
+    if exclude_id is not None:
+        qry = qry.exclude(id=exclude_id)
+    qry = (
+        qry.filter(location__distance_lt=(location, Distance(m=SITE_BUFFER_M)))
+        .annotate(similarity=TrigramSimilarity("name", name))
+        .filter(similarity__gte=SITE_NAME_MATCH_PERCENT)
+    )
+
+    return list(qry.distinct())
```
JBurkinshaw, line 88: The query annotates `similarity` but ends with `list(qry.distinct())` and no `.order_by()`, so results fall back to alphabetical, not similarity-descending.

Both consumers then slice `[:3]` expecting the closest matches (`_reject_if_duplicate` in mixins.py; `UniqueSiteValidator.__call__` in site.py). With 4+ qualifying duplicates, a user is shown an effectively arbitrary 3

src/api/resources/mixins.py
```diff
@@ -69,6 +75,106 @@ def create(self, validated_data):
...
+            # code/matches are nested under "duplicate" (rather than sibling
+            # keys next to "name") so DRF's error-detail normalization -- which
+            # wraps any non-list/dict value in a list -- doesn't turn code into
+            # a single-item list; nesting them in a dict value sidesteps that.
+            raise exceptions.ValidationError(
+                {
+                    "name": [message],
+                    "duplicate": {"code": code, "matches": [str(d.id) for d in duplicates[:3]]},
+                }
+            )
```
JBurkinshaw: The `"name"` key is hardcoded here. It works for Site and Management since both have a `name` field, but this is a shared mixin - if it's ever reused on a model without `name`, the error lands on the wrong key without giving an obvious reason why. Might be worth passing the field name in, or at least a note flagging the assumption.

src/api/utils/duplicates.py
```diff
@@ -0,0 +1,121 @@
...
+NOT_UNIQUE_SITE_CODE = "not_unique_site"
+SIMILAR_NAME_CODE = "similar_name"
+
+
+def _has_submitted_sample_unit_q(prefix):
+    """
+    Q object matching rows with at least one submitted sample unit,
+    reachable via the reverse-FK lookup path `prefix` to SampleEvent (e.g.
+    "sample_events" from Site, "sampleevent" from Management).
+    """
```
JBurkinshaw, line 42: This mentions a `"sampleevent"` prefix for Management, but the helper is only ever called with `"sample_events"` (Site) - Management uses the raw-SQL path instead. Could drop the Management example so a future reader doesn't assume there's an ORM path here that's actually used.

JBurkinshaw (changes requested):

JBurkinshaw (approved):

PR #747, opened 2026-07-23
[M2061] project/project_profile/cache cleanup issues

saanobhaai:
## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented deleting or changing a role in ways that would leave a project without an administrator.
  * Enforced these protections transactionally across the admin UI, API, and sync workflows.
  * Improved project authorization checks when updating project data.
  * Corrected permission/profile lookups when project or profile data is missing.

* **Improvements**
  * Project and profile permission data refreshes more accurately after changes via targeted cache invalidation.
  * Project deletion handling is more reliable for related protected relationships.


saanobhaai:
[project/project_profile/cache cleanup issues](https://trello.com/c/kYIYO4el/2061-project-projectprofile-cache-cleanup-issues)

src/api/admin/mermaid.py
```diff
@@ -120,13 +120,20 @@ def tag_list(self, obj):
...
+        # even though delete_project() below already resolves those
+        # relations itself, atomically. Skip the generic protected gate here.
+        to_delete = [str(obj) for obj in objs]
+        model_count = {self.model._meta.verbose_name_plural: len(objs)}
+        perms_needed = {
+            self.model._meta.verbose_name
+            for obj in objs
+            if not self.has_delete_permission(request, obj)
+        }
+        return to_delete, model_count, perms_needed, []
```
JBurkinshaw, line 147: Skipping the protected gate makes sense. But the code which creates the tuple also drops `perms_needed` for cascaded models — it's now computed only against `Project`. Since `Site`/`Management`/`SampleEvent` are admin-registered, a staff user with just `delete_project` can now cascade-delete all their survey data (superusers unaffected).

Assuming the minimal tuple was to dodge the `NestedObjects` collector cost and the huge confirmation page. Have you considered keeping that intent but restoring the check? Either (A) delegate to `super()` and return its `perms_needed`/`model_count` with `protected=[]` (simplest, but pays the full collection), or (B) keep the cheap page and run a collector purely to derive `perms_needed` from `collector.data`. A unless the page size is a real problem. As it is it silently weakens least-privilege.

src/api/permissions.py
```diff
@@ -84,6 +88,36 @@ def get_project_profile(project, profile, request=None):
...
+    sync push writes a project's own fields."""
+    cache = getattr(request, PROJECT_CACHE_ATTR, None)
+    if cache is None:
+        return
+    try:
+        pk = check_uuid(pk)
+    except ParseError:
+        return
+    cache.pop(pk, None)
+
```
JBurkinshaw, line 104: The project cache is keyed by raw ID: `get_project` stores under whatever `check_uuid` returns (a string, for current callers), and `invalidate_project` pops using the same string form, so they match today. But a dict treats `"abc-123"` and `UUID("abc-123")` as different keys. If any future caller passes `get_project` a `UUID` instead of a string, the entry gets stored under a `UUID` key while `invalidate_project` still looks for the string one, silently missing and leaving stale project data cached for the rest of the request.

Not a bug right now, just fragile. Normalizing the key to `uuid.UUID` in both `get_project` and `invalidate_project` removes the coincidence. (The profile cache already keys on real `.pk` UUIDs, so this only affects the project cache.)

JBurkinshaw (changes requested):

JBurkinshaw (approved):

PR #754, opened 2026-08-07
[M1718] Add org id to export user metrics report

gridcell:
Add organization id to the exported CSV file for user metrics.

## Summary by CodeRabbit

* **New Features**
  * Project metrics now include associated tag IDs alongside tag names.
  * Added support for storing project tag IDs in user metrics.
  * Existing projects without tags continue to work with an empty value.


src/tools/metrics.py
```diff
@@ -115,6 +115,7 @@ def get_project_lookup(log_events):
             "project_name": p.name,
             "project_status": p.get_status_display(),
             "project_tags": ",".join(t.name for t in p.tags.all()),
+            "project_tag_ids": ",".join(str(t.id) for t in p.tags.all()),
```
gridcell: **Warning** · `readability` · confidence: high

**p.tags.all() is iterated twice to build project_tags and project_tag_ids**

`p.tags.all()` is evaluated twice on consecutive lines (117-118) to build two comma-joined strings (tag names and tag ids). Since `tags` is prefetched via `prefetch_related("profiles", "tags", "sites")`, the second call reuses Django's prefetch cache rather than issuing an extra query, so this is not an N+1 problem — but it is a redundant/less-clear pattern that reads as an oversight or incomplete refactor.

> **Suggestion:** Hoist the queryset to a local variable once, e.g. `tags = list(p.tags.all())`, then build both joined strings from that list.

> **Why it matters:** Iterating the same collection twice for near-identical joins makes the intent less clear and can look like incomplete refactoring during future maintenance.

<sub>Found by: reviewer-backend-python, reviewer-python-quality · CR-001</sub>

src/tools/models.py
```diff
@@ -49,6 +49,7 @@ class UserMetrics(models.Model):
     )
     role = models.CharField(max_length=10, null=True, blank=True)
     project_tags = models.TextField()
+    project_tag_ids = models.TextField(default="", blank=True)
```
gridcell, line 52: **Info** · `style` · confidence: high

**Keyword argument ordering differs between migration and model definition**

The migration defines the field with kwargs in order `blank=True, default=""`, while the model definition reverses them to `default="", blank=True`. Django migrations are immutable snapshots, so this has no functional effect, but it's a trivial inconsistency worth normalizing if the migration is ever regenerated.

<sub>Found by: reviewer-python-quality · CR-005</sub>

src/tools/metrics.py
gridcell: **Info** · `security` · confidence: high

**Added tag UUIDs are already public data; no new exposure in the admin-only metrics export**

UserMetrics rows are written only by the `update_metrics` management command and read only through `UserMetricsAdmin`, which subclasses `BaseAdmin` and inherits `export_model_display_as_csv` / `export_model_all_as_csv` from `src/api/admin/base.py`. `export_model_all_as_csv` enumerates all concrete model fields, so the new column appears in the 'all fields' CSV export automatically — no downstream serializer change is needed. That surface requires an authenticated staff account. The values themselves are `api.Tag` primary keys, and `ProjectTagViewSet` serves the full Tag record to unauthenticated GET requests, so these identifiers are already public. The export row already carried `project_tags` (the names) plus far more sensitive PII (first_name, last_name, email, role). Net risk of this change is effectively zero.

<sub>Found by: reviewer-security · CR-007</sub>

gridcell (commented):
## AI Code Review

**Assessment:** This is a small, low-risk PR that adds a project_tag_ids TextField to UserMetrics, populated alongside the existing project_tags names field in get_project_lookup, with a safe metadata-only migration. No functional bugs or security issues were found — the migration is reversible and cannot lock the table, and the newly exported tag UUIDs are already public data with no new exposure through the admin-only CSV export. Remaining feedback is minor: a redundant queryset iteration worth tidying, a couple of style/consistency nits, an unaddressed gap in historical data for existing rows, and a PR title that does not match the actual diff (it adds project_tag_ids, not an org id) and should be confirmed with the author.

| 0 | 1 | 1 | 5 | 0 |

**Files reviewed:** 3 / 3

### Cross-Cutting Concerns
- No tests were added or modified for this change, and no existing test suite exercises src/tools/metrics.py or src/tools/models.py — the new field and its population logic (get_project_lookup) are entirely untested.
- The migration itself is a well-formed, safe pattern: a single AddField with a constant default (metadata-only on Postgres 11+, no table rewrite/lock), correct dependency chaining onto 0003 with no branching migration heads, and model/migration kwargs that match exactly.
- The new project_tag_ids field duplicates information derivable from the existing tag relations, but security review traced the actual consumption path (Django admin CSV export auto-includes all concrete model fields) and confirmed no downstream code needs to change to expose it — this appears to be an intentional, low-cost addition rather than dead/duplicated logic.

### General Findings

<details>
<summary><strong>CR-002</strong> &middot; Question &middot; <code>maintainability</code> — Comma-joined TextField vs ArrayField for project_tag_ids</summary>

`project_tag_ids` is stored as a comma-joined TextField, mirroring the existing `project_tags` naming convention. This is internally consistent, but since the values are semantically structured data (a list of UUIDs) rather than free text, an ArrayField would be more queryable (e.g. `__contains` filters) and avoid string-splitting for any future downstream consumer.

Applies to: `src/tools/models.py`, `src/tools/metrics.py`

</details>

<details>
<summary><strong>CR-003</strong> &middot; Info &middot; <code>docs</code> — PR title says "Add org id" but the diff adds project_tag_ids, not an org id</summary>

The PR title references adding an organization id to the export user metrics report, but the actual diff only adds a `project_tag_ids` field (a comma-joined list of tag ids, mirroring the existing `project_tags` names field). There is no org_id/organization field anywhere in `src/tools`. This may be a stale/copy-pasted title, or it may indicate the org id portion of ticket M1718 was not actually implemented in this PR.

> **Suggestion:** Confirm whether an organization id was intended as part of M1718. If so, it appears to be missing from this diff and should be added (or split into a follow-up PR); if not, update the PR title/description to accurately describe the change (adds project_tag_ids).

> **Why it matters:** A mismatched title/description can cause reviewers and future maintainers to believe functionality (org id export) exists when it doesn't, and could mean the ticket's actual requirement was missed.

Applies to: `src/tools/metrics.py`, `src/tools/models.py`

</details>

<details>
<summary><strong>CR-004</strong> &middot; Info &middot; <code>style</code> — project_tags lacks the default/blank kwargs that the new project_tag_ids has</summary>

`project_tags` has no `default`/`blank` specified, while the new `project_tag_ids` includes `default="", blank=True`. Both fields store comma-separated values and are always initialized to empty strings in `PROJECT_DEFAULT`, so in practice this asymmetry causes no observed bug — `project_tags` is always explicitly populated by application code, not left to its field default. This is a minor style inconsistency rather than a functional issue.

> **Suggestion:** Optional: align `project_tags` with `default="", blank=True` for consistency, but note this is out of scope for this PR and would require its own migration — not worth doing unless a broader model cleanup is planned.

> **Why it matters:** Inconsistent field definitions make the model slightly harder to scan, though it has no functional impact given current usage.

Applies to: `src/tools/models.py`, `src/tools/metrics.py`

</details>

<details>
<summary><strong>CR-006</strong> &middot; Info &middot; <code>correctness</code> — Existing UserMetrics rows will report an empty project_tag_ids while project_tags (names) remains populated</summary>

`UserMetrics` rows are append-only, written daily by `update_metrics.py` via `metrics.agg_log_events()` / `get_project_lookup()`. The new `project_tag_ids` field is only populated going forward — all historically existing rows will carry `project_tag_ids=""` while their `project_tags` column is already populated. Any consumer expecting `project_tag_ids` to correlate with `project_tags` for a given row will see this asymmetry for all pre-existing dates.

> **Suggestion:** Consider either (a) documenting that `project_tag_ids` is only reliable from the deploy date forward, or (b) a one-off backfill that re-resolves tag names to current tag IDs (accepting that tags renamed/deleted since won't map perfectly).

> **Trade-off:** A backfill by name-matching is best-effort only — tags could have been renamed, merged, or deleted since, so it could introduce incorrect IDs rather than just leaving them blank.

> **Why it matters:** Reports/dashboards built on this table for historical date ranges could show missing tag IDs even though tag names are present, which could look like a bug rather than expected 'field didn't exist yet' behavior.

Applies to: `src/tools/models.py`, `src/tools/metrics.py`, `src/tools/migrations/0004_usermetrics_project_tag_ids.py`

</details>

### Agent Coverage
| reviewer-backend-python | backend | 3 | 3 | 0 |
| reviewer-python-quality | code_quality | 3 | 3 | 0 |
| reviewer-database | database | 2 | 4 | 0 |
| reviewer-security | security | 3 | 1 | 0 |

PR #756, opened 2026-08-12
Support  .sql and .dump backup files in the dbrestore

gridcell:
## Summary by CodeRabbit

* **New Features**
  * Project metrics now include associated project tag names and IDs.
  * Database restore supports both `.dump` and `.sql` backup files.
  * Restore errors provide clearer details about searched locations and supported file types.

* **Bug Fixes**
  * Improved discovery and validation of local and cloud-hosted restore files.
  * Added storage for project tag IDs in user metrics.


src/tools/management/commands/dbrestore.py
```diff
@@ -55,37 +57,37 @@ def handle(self, *args, **options):
...
-                expected_suffix = f".{BACKUP_EXTENSION}"
+                expected_marker = "mermaid_backup_"
                 for f in os.listdir(tmpdir):
                     localfile = os.path.join(tmpdir, f)
                     if (
                         os.path.isfile(localfile)
-                        and f.startswith(expected_prefix)
-                        and f.endswith(expected_suffix)
+                        and expected_marker in f
+                        and any(f.endswith(ext) for ext in RESTORABLE_EXTENSIONS)
```
gridcell: **Severe** · `security` · confidence: high

**Local backup file matching lost environment scoping — dbrestore -n can silently restore a different (e.g. production) environment's data**

The local-file predicate changed from `f.startswith(f"{self.restore}_mermaid_backup_")` to a bare substring test: `expected_marker = "mermaid_backup_"` / `expected_marker in f`. The environment/target component of the filename is no longer checked at all, so every file in /tmp/mermaid containing that substring is a candidate, and the newest by mtime wins regardless of which environment it belongs to. This is directly reachable in normal workflows: files downloaded from S3 by this same command are named `{LastModified}_{Key with '/' replaced by '_'}` (e.g. `20250612093000_prod_mermaid_backup_123.dump`), and `src/tools/management/commands/check_gfcr_data.py:65` runs `call_command("dbrestore", "prod")`, leaving exactly such a prod dump behind in the shared /tmp/mermaid directory. A subsequent `python manage.py dbrestore local -n` will select that prod file (it contains the marker and is newest), drop the local database, and restore production data into it. The error message immediately below still claims the search was scoped (`No local {self.restore} backup files found`), so the mismatch between intent and behavior is silent. One agent (general-purpose) rated this a warning rather than severe; the other three agents rated it severe given the destructive, irreversible nature of the operation and the demonstrated production-data exposure path — the merged severity reflects that.

> **Suggestion:** Keep the environment component in the match while tolerating the timestamp prefix that S3 downloads add, e.g.:
```python
import re
pattern = re.compile(rf"^(\d+_)?{re.escape(self.restore)}_mermaid_backup_.+$")
...
if (
    os.path.isfile(localfile)
    and pattern.match(f)
    and f.endswith(RESTORABLE_EXTENSIONS)
):
```
Any equivalent that requires `{self.restore}_mermaid_backup_` (not just `mermaid_backup_`) as a delimited component of the filename is fine.

> **Why it matters:** Restoring a production dump into a developer or staging database moves production data (including any PII in project/profile tables) into an environment with weaker access controls, and it happens after the target database has already been dropped, so there is no undo. This is a real data-safety regression, not a theoretical one, given the demonstrated call path through check_gfcr_data.py.

<sub>Found by: spk-reviewer-security, spk-reviewer-backend-python, spk-reviewer-database, spk-reviewer-general-purpose · CR-001</sub>

src/tools/management/commands/dbrestore.py
```diff
@@ -156,9 +158,11 @@ def _psql_restore_db(self, file_name):
...
-            **params
-        )
+        cmd_str = "pg_restore -O -x --jobs=4 "
+        if file_name.lower().endswith(DUMP_EXT):
+            cmd_str += "-F c "
+
+        cmd_str += "-U {db_user} -h {db_host} -d {db_name} {sql_loc}".format(**params)
         self.stdout.write("$> %s" % cmd_str)
 
         command = shlex.split(cmd_str)
```
gridcell: **Severe** · `correctness` · confidence: high

**Legacy .sql backups are still restored with pg_restore, which cannot read plain-text SQL dumps — this is the core feature the PR claims to add, and it does not work**

For files ending in `.sql`, `_psql_restore_db` now omits `-F c` but still invokes `pg_restore`. Dropping the flag does not make pg_restore able to parse a plain-text SQL script — pg_restore only understands archives produced by `pg_dump -F c|t|d` (custom/tar/directory) and explicitly refuses plain text with 'input file appears to be a text format dump. Please use psql.' Git history confirms the correct tool: prior to the switch to custom-format dumps (commit 9c322a67, 'switch to pg_restore'), legacy restores used `psql -U {db_user} -h {db_host} -d {db_name} -q -f {sql_loc}`. `dbbackup.py` only ever produces `.dump` files via `pg_dump -F c`, so the `.sql` branch exists solely to support old archived backups — and as written it will fail every single time it is exercised, defeating the stated purpose of this PR.

> **Suggestion:** Branch on extension and invoke a different tool for the legacy format:
```python
if file_name.lower().endswith(DUMP_EXT):
    cmd_str = "pg_restore -O -x -F c --jobs=4 -U {db_user} -h {db_host} -d {db_name} {sql_loc}".format(**params)
else:
    cmd_str = "psql -v ON_ERROR_STOP=1 -U {db_user} -h {db_host} -d {db_name} -f {sql_loc}".format(**params)
```

> **Why it matters:** This is the core feature of the PR. Without it, restoring any legacy .sql backup fails at the pg_restore invocation — after the target database has already been dropped and recreated by `_init_db()` — leaving it empty with no automatic recovery.

<sub>Found by: spk-reviewer-backend-python, spk-reviewer-database, spk-reviewer-general-purpose · CR-002</sub>

src/tools/management/commands/dbrestore.py
```diff
@@ -112,7 +114,7 @@ def handle(self, *args, **options):
             if (
                 not download_file_name
                 or not os.path.isfile(download_file_name)
-                or not download_file_name.endswith(f".{BACKUP_EXTENSION}")
+                or not download_file_name.endswith(RESTORABLE_EXTENSIONS)
             ):
                 raise ValueError("No database dump file to restore")
             self._init_db()
```
gridcell, line 123: **Severe** · `error_handling` · confidence: high

**Destructive DROP DATABASE runs before the archive is validated as restorable, guaranteeing data loss whenever the .sql/pg_restore mismatch (or any malformed file) is hit**

The pre-flight check only verifies the path exists and ends in `.dump`/`.sql`; `self._init_db()` (terminate connections, `DROP DATABASE IF EXISTS`, `CREATE DATABASE`) then runs unconditionally before `_psql_restore_db()` is ever invoked. There is no validation that the selected file is actually a well-formed, restorable archive before the existing database is destroyed. Accepting `.sql` widens what passes that extension-only check without adding any real verification: given CR-002 (pg_restore cannot read plain SQL), selecting any legacy `.sql` file will always hit this failure path — the database is dropped and recreated empty, and pg_restore then immediately errors with 'did not find magic string in file header', leaving an empty database behind with no undo. One agent (security) rated this a warning; database rated it severe given the guaranteed, irreversible data loss it produces in combination with CR-002 — the merged severity reflects the latter.

> **Suggestion:** Validate the archive before the destructive step — e.g. run `pg_restore -l <file>` (or an equivalent header check for `.sql`) and require success before calling `self._init_db()`. Abort with a clear error if validation fails, before anything is dropped.

> **Why it matters:** This command is irreversibly destructive: once `_init_db()` has run, a failed restore leaves the target database dropped and empty with no rollback path.

<sub>Found by: spk-reviewer-security, spk-reviewer-database · CR-003</sub>

src/tools/management/commands/dbrestore.py
```diff
@@ -55,37 +57,37 @@ def handle(self, *args, **options):
...
+                        and any(f.endswith(ext) for ext in RESTORABLE_EXTENSIONS)
                     ):
                         if download_file_name is None or os.path.getmtime(
                             localfile
                         ) > os.path.getmtime(download_file_name):
                             download_file_name = localfile
 
                 if download_file_name is None:
+                    raise ValueError(f"No local {self.restore} backup files found in {tmpdir}")
                     raise ValueError(f"No local files for {self.env} found")
```
gridcell: **Warning** · `correctness` · confidence: high

**Unreachable duplicate `raise ValueError` left over from the edit, and it still references a stale `self.env` attribute**

Two consecutive `raise` statements appear back-to-back: `raise ValueError(f"No local {self.restore} backup files found in {tmpdir}")` always fires, making the following line — `raise ValueError(f"No local files for {self.env} found")` — permanently unreachable dead code. It also still references `self.env` rather than `self.restore`, suggesting the old message was meant to be deleted, not kept, when the new one was added. All five agents flagged this; severity ratings ranged from info to severe across agents (this was ultimately the only 'severe' rating among the five, from the code-quality reviewer, on the grounds that dead/unreachable code from an incomplete refactor is itself a correctness defect) — the merged severity reflects that highest rating, though the practical runtime impact is limited to code-cleanliness rather than an active bug, since the line can never execute.

> **Suggestion:** Delete the unreachable second `raise ValueError(...)` line, keeping only the new, more informative message.

> **Why it matters:** Dead code with an inconsistent variable reference is confusing for future maintainers and signals the refactor wasn't fully completed; the unreachable message will never be seen by users, which can mislead anyone searching the codebase for its text.

<sub>Found by: spk-reviewer-security, spk-reviewer-backend-python, spk-reviewer-database, spk-reviewer-python-quality, spk-reviewer-general-purpose · CR-004</sub>

src/tools/management/commands/dbrestore.py
gridcell: **Warning** · `security` · confidence: medium

**Filename interpolated into a command string that is then shlex.split — argument injection into pg_restore**

The command construction still builds one string and splits it: `cmd_str += "-U {db_user} -h {db_host} -d {db_name} {sql_loc}".format(**params)` then `shlex.split(cmd_str)`. `sql_loc` is not quoted and is derived from data outside the process's control — the S3-downloaded filename only has `/` neutralized (line ~101), and the local `-n` path accepts any matching entry in /tmp/mermaid verbatim. A name containing whitespace is split by shlex into multiple argv entries, letting whoever can write to the backup bucket or to /tmp/mermaid inject additional pg_restore options; a name containing an unbalanced quote makes `shlex.split` raise `ValueError` after the database has already been dropped. There is no shell involved (`run_subprocess` calls `subprocess.run(command, ...)` without `shell=True`), so this is argument injection, not full command injection.

> **Suggestion:** Build argv directly instead of round-tripping through a string:
```python
command = ["pg_restore", "-O", "-x", "--jobs=4"]
if file_name.lower().endswith(DUMP_EXT):
    command += ["-F", "c"]
command += ["-U", params["db_user"], "-h", params["db_host"], "-d", params["db_name"], "--", file_name]
self.stdout.write("$> %s" % shlex.join(command))
run_subprocess(command, to_file="/tmp/mermaid/std_out_restore.log")
```
The `--` terminator also prevents a filename beginning with `-` from being read as a flag.

> **Why it matters:** The restore runs with privileged database credentials against a database that was just dropped and recreated. Attacker-influenced argv means influence over what gets loaded into that fresh database, and malformed names cause an opaque ValueError after the DB is already destroyed.

src/tools/management/commands/dbrestore.py
gridcell, line 70: **Warning** · `style` · confidence: high

**Redundant any() generator instead of the native tuple-accepting endswith()**

The local-file predicate uses `any(f.endswith(ext) for ext in RESTORABLE_EXTENSIONS)` to check file extensions. However, `str.endswith()` natively accepts a tuple. This is inconsistent with the S3-listing branch a few lines below in the same function, which correctly uses `obj["Key"].endswith(RESTORABLE_EXTENSIONS)` with the tuple directly.

> **Suggestion:** Replace with `and f.endswith(RESTORABLE_EXTENSIONS)` to match the pattern used for the S3 listing check. This is both simpler and more consistent.

> **Why it matters:** The generator-based approach creates unnecessary overhead and is inconsistent with an identical check a few lines below, which hurts readability.

<sub>Found by: spk-reviewer-python-quality · CR-006</sub>

gridcell (commented):
## AI Code Review

> **Verdict: blocking — changes requested.** Posted as a COMMENT review because GitHub does not allow REQUEST_CHANGES on your own pull request.

**Assessment:** The PR's stated goal — letting dbrestore accept legacy plain-text .sql backups alongside the current pg_restore custom-format .dump files — is not actually achieved: the .sql branch still shells out to pg_restore, which cannot parse plain-text SQL and will fail every time. Independently, the local-file (`-n`) selection predicate lost its environment-name scoping, so `dbrestore local -n` can silently pick and restore a different environment's (including production's) backup file, and the unconditional DROP DATABASE in `_init_db()` still runs before any validation that the chosen archive is actually restorable, so hitting either bug destroys the target database and leaves it empty with no rollback. A leftover unreachable `raise` statement (referencing a now-stale `self.env`) also indicates the refactor was left incomplete. Notably, two prior CodeRabbit review threads on this PR already raised the pg_restore/.sql mismatch and the unscoped filename match and were marked resolved, but both defects are still present in head commit 1e353a14 — the threads were closed without the underlying code being fixed. This PR should not merge until the restore-tool branching, the environment-scoped file matching, and the pre-flight validation are corrected.

| 3 | 3 | 0 | 3 | 0 |

**Files reviewed:** 1 / 1

### Cross-Cutting Concerns

- Two prior CodeRabbit review threads on this PR already raised the exact two core issues found here — the pg_restore/.sql format mismatch (CR-002) and the unscoped local-file match (CR-001) — and both threads are marked 'resolved' on GitHub, yet both defects remain present in head commit 1e353a14. The threads appear to have been closed without the underlying code being fixed; this review re-raises them because they still affect the shipped diff.
- No automated test coverage exists for dbrestore or dbbackup (CR-009). A single unit test asserting the constructed restore command per file extension and per environment would have caught CR-001 and CR-002 before they reached review.
- The pattern of building a command as a single interpolated string and then `shlex.split`-ing it (flagged in CR-005) is not unique to this change — the same pattern exists in `_init_db()` in this file and in `dbbackup.py`. If CR-005 is addressed, consider a follow-up PR to apply the same argv-list treatment consistently across all subprocess invocations in these two commands.
- Both CR-001 and CR-003 compound each other: because the target database is unconditionally dropped before any validation of the selected file, either a wrong-environment file (CR-001) or an unrestorable format (CR-002/CR-003) results in irreversible data loss rather than a safe failure. Fixing only one of the three without the others still leaves a destructive-failure path open.

### Additional Findings

<details>
<summary><strong>CR-009</strong> &middot; Info &middot; <code>test_gap</code> — No test coverage for dbrestore (or its sibling dbbackup) command, before or after this change</summary>

There is no test file exercising `manage.py dbrestore`. The extension-matching logic and the `.dump` vs `.sql` branching in `_psql_restore_db` are exactly the kind of small, easy-to-get-wrong logic that would benefit from a unit test mocking `s3.list_objects`/`os.listdir` and asserting the constructed command string per extension. Such a test would very likely have caught both CR-001 (env-scoping regression) and CR-002 (pg_restore/.sql mismatch) before merge.

> **Suggestion:** Add a focused test (mocking `api.utils.s3` and `run_subprocess`) that verifies: (1) `.dump` and `.sql` S3 objects are both discovered, (2) the local no-download path only picks files matching the requested environment, and (3) the constructed restore command differs correctly between `.dump` and `.sql` inputs.

Applies to: `src/tools/management/commands/dbrestore.py`, `src/tools/management/commands/dbbackup.py`

<sub>Found by: spk-reviewer-general-purpose</sub>
</details>

<details>
<summary><strong>CR-008</strong> &middot; Info &middot; <code>correctness</code> — str.endswith(tuple) usage confirmed correct for the S3-listing branch — no change needed</summary>

`RESTORABLE_EXTENSIONS` is a tuple, and `str.endswith()` natively accepts a tuple of suffixes, so `obj["Key"].endswith(RESTORABLE_EXTENSIONS)` works as intended in the S3-listing branch. Noted for completeness; see CR-006 for the inconsistent generator-based variant used in the local-file branch.

<sub>Found by: spk-reviewer-backend-python</sub>
</details>

### Agent Coverage

| spk-reviewer-security | security | 1 | 5 | 1 |
| spk-reviewer-backend-python | backend | 1 | 4 | 2 |
| spk-reviewer-database | database | 1 | 4 | 3 |
| spk-reviewer-python-quality | code_quality | 1 | 2 | 1 |
| spk-reviewer-general-purpose | general | 1 | 4 | 1 |

src/tools/management/commands/dbrestore.py
```diff
@@ -55,37 +57,37 @@ def handle(self, *args, **options):
...
+                    for obj in s3.list_objects(settings.AWS_BACKUP_BUCKET, prefix=prefix)
+                    if not obj["Key"].endswith("/") and obj["Key"].endswith(RESTORABLE_EXTENSIONS)
                 ]
 
                 if not backup_objects:
                     raise ValueError(
-                        f"{settings.AWS_BACKUP_BUCKET} does not exist or is not listable"
+                        f"No backups found at s3://{settings.AWS_BACKUP_BUCKET}/{prefix} "
+                        f"(expected keys ending in {' or '.join(RESTORABLE_EXTENSIONS)})"
                     )
```
gridcell, line 94: **Info** · `security` · confidence: high

**Bucket name in the new error message is acceptable disclosure here — no action needed**

The improved error text surfaces `settings.AWS_BACKUP_BUCKET` and the key prefix, wrapped in a `CommandError`. This is an operator-only management command whose output goes to a console/CI log, never to an HTTP response, and the same bucket name is already printed on the happy path by `dbbackup`. No credentials, presigned URLs, or account IDs are exposed. The added specificity is a net operability win; flagging only so the disclosure is documented as a deliberate decision.

<sub>Found by: spk-reviewer-security · CR-007</sub>

src/tools/management/commands/dbrestore.py
```diff
@@ -159,6 +166,7 @@ def _psql_restore_db(self, file_name):
         cmd_str = "pg_restore -O -x -F c --jobs=4 -U {db_user} -h {db_host} -d {db_name} {sql_loc}".format(
             **params
         )
+
         self.stdout.write("$> %s" % cmd_str)
 
         command = shlex.split(cmd_str)
```
gridcell, line 170: **Severe** · `correctness` · confidence: high

**pg_restore cannot restore plain-text .sql dumps; legacy .sql support is non-functional end-to-end**

`_psql_restore_db` unconditionally builds and runs `pg_restore -O -x -F c --jobs=4 -U {db_user} -h {db_host} -d {db_name} {sql_loc}` regardless of which extension (`.dump` vs `.sql`) was selected upstream. `pg_restore` only understands archive-format dumps (custom/tar/directory produced by `pg_dump -F c/t/d`) and explicitly refuses plain-text SQL scripts, directing the user to `psql` instead. Git history confirms `pg_dump` switched to custom-archive format (`-F c`) in commit 9c322a67 (2020), while the `.sql` extension itself was only renamed to `.dump` in commit b43d1036; any genuinely legacy `.sql`-named backup predating that switch is real plain-text SQL and was historically restored with `psql -q -f {sql_loc}`, not `pg_restore`. This PR widens file selection (`RESTORABLE_EXTENSIONS`) to accept `.sql` files but never branches the restore command on actual format, so a legitimate legacy `.sql` backup — exactly what the PR claims to support, and the same 'local' backup keyname that is exempt from age-based cleanup because it's an old hand-curated dataset — will be selected, pass validation, and then fail when handed to `pg_restore`. The unused `PGDMP_MAGIC = b"PGDMP"` constant added in this same PR strongly suggests format-sniffing logic was intended but never wired up. ⚠️ This issue was previously flagged and discussed but remains present in the code.

> **Suggestion:** Branch the restore command on the real archive format rather than file extension: read the first bytes of the downloaded file and check for the `PGDMP` magic (the added-but-unused `PGDMP_MAGIC` constant) to decide between `pg_restore -F c ...` for custom-format archives and `psql -v ON_ERROR_STOP=1 -U {db_user} -h {db_host} -d {db_name} -f {sql_loc}` for plain-text `.sql` dumps. Perform this check before `_init_db()` runs (see CR-002).

> **Why it matters:** As written, attempting to restore a genuine legacy plain-text `.sql` backup — the explicit purpose of this PR — will drop/recreate the target database via `_init_db()` and then fail during `pg_restore`, leaving the database empty with no automatic rollback.

<sub>Found by: spk-reviewer-backend-python, spk-reviewer-general-purpose · CR-001</sub>

src/tools/management/commands/dbrestore.py
```diff
@@ -112,7 +119,7 @@ def handle(self, *args, **options):
             if (
                 not download_file_name
                 or not os.path.isfile(download_file_name)
-                or not download_file_name.endswith(f".{BACKUP_EXTENSION}")
+                or not download_file_name.endswith(RESTORABLE_EXTENSIONS)
             ):
                 raise ValueError("No database dump file to restore")
             self._init_db()
```
gridcell, line 123: **Severe** · `error_handling` · confidence: high

**Destructive _init_db() (DROP DATABASE) runs before any validation that the selected archive can actually be restored**

`handle()` only checks the file's name/extension (lines 123-128) before calling `self._init_db()` at line 129, which runs `DROP DATABASE IF EXISTS ...; CREATE DATABASE ...`. Only after the database has been destroyed does `_psql_restore_db()` (line 130) attempt the actual restore. If the selected file is not a format `pg_restore` can read (see CR-001) — or is otherwise corrupt or truncated — `run_subprocess` raises `CalledProcessError`, which is caught by the outer `try/except Exception` and re-raised as a `CommandError`, but by then the original database is already gone and nothing has replaced it. ⚠️ This issue was previously flagged and discussed but remains present in the code.

> **Suggestion:** Validate the downloaded archive before calling `_init_db()` — e.g. peek at the file's leading bytes for the `PGDMP` magic, or at minimum run `pg_restore --list {sql_loc}` (which reads the TOC without touching any database) to confirm the archive is well-formed prior to dropping the live database.

> **Why it matters:** A restore command whose only job is to bring back a working database should never leave the target database in a worse, unusable state than before it ran. This ordering turns a recoverable 'bad backup file' error into an unplanned outage requiring manual intervention.

<sub>Found by: spk-reviewer-general-purpose · CR-002</sub>

src/tools/management/commands/dbrestore.py
```diff
@@ -1,12 +1,17 @@
...
 from django.core.management.base import BaseCommand, CommandError
 
 from api.utils import run_subprocess, s3
 
-BACKUP_EXTENSION = "dump"
+DUMP_EXT = ".dump"
+LEGACY_DUMP_EXT = ".sql"
+RESTORABLE_EXTENSIONS = (DUMP_EXT, LEGACY_DUMP_EXT)
+# Leading bytes of a `pg_dump -F c` archive.
+PGDMP_MAGIC = b"PGDMP"
```
gridcell: **Warning** · `security` · confidence: high

**PGDMP_MAGIC constant added but never wired up — backup content validation still relies solely on filename/extension**

The constant `PGDMP_MAGIC = b"PGDMP"` (the leading bytes of a `pg_dump -F c` archive) is added but never read, compared, or referenced anywhere in the file. Selection and validation of the backup file to restore rely solely on filename/extension matching (`RESTORABLE_EXTENSIONS`, plus the `-n` mode's regex and `any(f.endswith(ext) ...)` check), so a mislabeled, renamed, or planted file (e.g., a custom-format dump saved with a `.sql` extension, a corrupt file, or — in `-n` mode — a file placed by another process in the shared `/tmp/mermaid` directory, which is created with `exist_ok=True` and reused across runs) can win the newest-mtime selection and be handed straight to `pg_restore`, which then executes its contents against the target database as the configured DB user. All four reviewers independently flagged this same unused constant, which strongly suggests format-sniffing/branching logic (see CR-001) was intended but never completed, and leaves a misleading impression that content is verified when it is not.

> **Suggestion:** Wire the constant into an actual check — read the first `len(PGDMP_MAGIC)` bytes of the candidate file and compare against `PGDMP_MAGIC` to determine whether it is a custom-format archive before invoking `pg_restore`, and use that result to select the restore code path (see CR-001) rather than trusting the extension alone. If content-sniffing is not going to be implemented, remove the unused constant and its comment so it doesn't imply a safety check that doesn't exist.

> **Why it matters:** Restoring an unverified archive runs arbitrary SQL/DDL against the target database; relying purely on filename/extension is weak when the source is a shared, predictable directory (`-n` mode) or an S3 listing. Dead code describing unimplemented safety logic also misleads future readers into false confidence.

<sub>Found by: spk-reviewer-security, spk-reviewer-backend-python, spk-reviewer-python-quality, spk-reviewer-general-purpose · CR-003</sub>

src/tools/management/commands/dbrestore.py
```diff
@@ -55,14 +60,16 @@ def handle(self, *args, **options):
                 download_file_name = None
                 tmpdir = os.path.join(os.path.sep, self.local_file_location)
 
-                expected_prefix = f"{self.restore}_mermaid_backup_"
-                expected_suffix = f".{BACKUP_EXTENSION}"
+                # Match "{env}_mermaid_backup_*" as written by dbbackup, optionally with the
+                # timestamp prefix an S3 download adds. Anchoring on the env keeps a dump from
+                # another environment (eg prod) from being restored over this one.
+                pattern = re.compile(rf"^(\d+_)?{re.escape(self.restore)}_mermaid_backup_.+$")
```
gridcell, line 64: **Question** · `correctness` · confidence: medium

**Confirm the env-anchored regex intentionally allows any digits-prefixed match after 'env_mermaid_backup_'**

The new pattern `rf"^(\d+_)?{re.escape(self.restore)}_mermaid_backup_.+$"` correctly anchors on `self.restore` and tolerates the S3-download timestamp prefix, matching filenames produced by `dbbackup.py` (`{backup}_mermaid_backup_{id}.{ext}`) and by this command's own S3-download naming (`{ts}_{backup}_mermaid_backup_{id}.{ext}`). One edge case worth confirming: since `simpleflake()` ids and the S3-download timestamp are both purely numeric and the `(\d+_)?` group is optional and unbounded in length, is there any scenario (e.g. an env name that itself starts with digits) where a file meant for a different environment could still satisfy the pattern? Given this feeds directly into a `DROP DATABASE`, it is worth a second look even though current default envs (local/dev/staging/prod) don't trigger it.

<sub>Found by: spk-reviewer-general-purpose · CR-005</sub>

gridcell, line 64: **Info** · `security` · confidence: high

**Comment claims env anchoring prevents cross-environment restores, but the anchor is the user-supplied argument, not ENV**

The comment states "Anchoring on the env keeps a dump from another environment (eg prod) from being restored over this one," but the pattern is built from `self.restore`, which is overwritten by the positional CLI argument at line 51 (`self.restore = restore_name`) and only equals `self.env` when no argument is passed. `python manage.py dbrestore prod -n` in a dev container matches and restores a prod dump — and `src/tools/management/commands/check_gfcr_data.py:65` does exactly `call_command("dbrestore", "prod")` on purpose. The only environment guard in the file is the `self.env == "prod"` force check at line 55, which protects prod as a *target*, not prod data as a *source*. Worth correcting the wording so a future reader does not treat this as an enforced control.

<sub>Found by: spk-reviewer-security · CR-006</sub>

src/tools/management/commands/dbrestore.py
```diff
@@ -75,17 +82,17 @@ def handle(self, *args, **options):
...
+                    for obj in s3.list_objects(settings.AWS_BACKUP_BUCKET, prefix=prefix)
+                    if not obj["Key"].endswith("/") and obj["Key"].endswith(RESTORABLE_EXTENSIONS)
                 ]
 
                 if not backup_objects:
                     raise ValueError(
-                        f"{settings.AWS_BACKUP_BUCKET} does not exist or is not listable"
+                        f"No backups found at s3://{settings.AWS_BACKUP_BUCKET}/{prefix} "
+                        f"(expected keys ending in {' or '.join(RESTORABLE_EXTENSIONS)})"
                     )
```
gridcell, line 94: **Info** · `readability` · confidence: high

**Improved error messages are a good, low-risk change**

The error messages raised when no local file or S3 object is found were made much more actionable (they now include the search directory, bucket, prefix, and expected extensions), which will speed up debugging failed restores.

<sub>Found by: spk-reviewer-general-purpose · CR-007</sub>

gridcell (commented):
## AI Code Review

**Assessment:** The PR extends dbrestore's file selection to accept legacy .sql backups alongside .dump (with a nice env-anchoring regex fix), but the restore execution path was never branched for plain-text SQL: _psql_restore_db unconditionally invokes pg_restore -F c, which cannot read a plain-text .sql dump, and the destructive DROP DATABASE step runs before any content validation. Together these mean the PR's stated goal is not achieved end-to-end and can leave the target database dropped and empty on failure; both issues were previously flagged and marked resolved by the author but remain present in the head commit, so they are re-raised here per the severe-finding exception. One shlex-split argv-construction nitpick that was also marked resolved has been suppressed since it is not severe.

| 2 | 2 | 1 | 2 | 3 |

**Files reviewed:** 1 / 1
**Blocking:** yes — 2 severe findings. (Posted as a comment rather than Request Changes because GitHub does not allow requesting changes on your own PR.)
**Previously addressed:** 1 finding(s) suppressed (resolved or acknowledged in prior review)

### Cross-Cutting Concerns
- The PR's core goal (supporting legacy .sql restores) is undermined end-to-end by the combination of CR-001 and CR-002: file selection now accepts .sql files, but the destructive DROP DATABASE step runs before any validation that the selected file can actually be restored by the hardcoded pg_restore -F c invocation.
- All four specialist agents independently converged on the same root cause (CR-003: the unused PGDMP_MAGIC constant), which is a strong signal that content-format sniffing was planned but not finished before merge.
- No unit tests were added for the new file-selection regex, S3 key filtering, or extension handling (CR-004), despite this command being destructive (DROP DATABASE) and the sole regression safety net for these changes.
- A previous AI review round already flagged the pg_restore/.sql mismatch (CR-001) and the DROP-before-validation ordering (CR-002) as severe issues; both threads were marked resolved by the author but the underlying code is unchanged in this commit, so they are re-raised per policy for severe findings.

### Additional Findings

<details>
<summary><strong>CR-004</strong> &middot; Warning &middot; <code>test_gap</code> — No test coverage added for the new local-file/S3 selection logic or the .sql restore path</summary>

This PR changes the regex used to pick a local backup file (line 66), the S3 key filter (line 89), and the accepted-extension validation (line 126), plus claims to add support for restoring `.sql` files — none of which have accompanying unit tests. A search of `src/api/tests` and the repo for dbrestore/dbbackup test files found nothing. Given the destructive nature of this command (`DROP DATABASE`) and the subtlety of the new regex/extension matching (env-anchoring, optional timestamp prefix, tuple-based `str.endswith`), this is exactly the kind of logic that benefits from focused unit tests with mocked `os.listdir`/`s3.list_objects`.

> **Suggestion:** Add unit tests that mock the filesystem/S3 layer to verify: (1) the regex matches `dbbackup.py`-style local filenames and S3-download timestamp-prefixed filenames for a given `self.restore` env, (2) it rejects files belonging to a different environment, (3) both `.dump` and `.sql` files are picked up, and (4) newest-by-mtime selection works when multiple candidates exist.

> **Why it matters:** Without tests, regressions in file-selection (e.g. accidentally restoring a different environment's dump, or silently ignoring valid backups) would only be discovered against real infrastructure, where the blast radius is a dropped database.

Applies to: `src/tools/management/commands/dbrestore.py`

<sub>Found by: spk-reviewer-general-purpose</sub>
</details>

<details>
<summary><strong>CR-008</strong> &middot; Praise &middot; <code>security</code> — User-supplied restore name is re.escape'd before being interpolated into the pattern</summary>

`re.escape(self.restore)` correctly neutralizes the caller-controlled positional argument before it is embedded in the compiled pattern, so a value like `.*` or `(a+)+` cannot turn the filename filter into a match-anything pattern or a catastrophic-backtracking payload. Easy detail to miss when moving from `startswith`/`endswith` to a regex.

Applies to: `src/tools/management/commands/dbrestore.py:66-66`

<sub>Found by: spk-reviewer-security</sub>
</details>

<details>
<summary><strong>CR-009</strong> &middot; Praise &middot; <code>correctness</code> — Env-anchored regex closes a cross-environment restore hazard</summary>

Replacing the plain prefix/suffix `startswith`/`endswith` check with a regex anchored on `self.restore` (with `re.escape`) and an explicit comment explaining the intent is a solid improvement — it prevents a same-directory backup from a different environment (e.g. a stray prod dump) from being picked up when restoring `local`, while still tolerating the S3-download timestamp prefix. Note: CR-006 flags that the accompanying comment overstates this as an unconditional env guard, since `self.restore` can be deliberately overridden via the CLI argument (an intentional call-site pattern elsewhere in the codebase) — the fix itself is sound, but the wording should be tightened.

Applies to: `src/tools/management/commands/dbrestore.py:63-66`

<sub>Found by: spk-reviewer-backend-python</sub>
</details>

<details>
<summary><strong>CR-010</strong> &middot; Praise &middot; <code>correctness</code> — Tuple-based endswith and extension constants are a clean, DRY refactor</summary>

Consolidating the two accepted extensions into `RESTORABLE_EXTENSIONS` and using `str.endswith(tuple)` (lines 89 and 126) instead of repeating string concatenation is a nice simplification over the old single-extension approach.

Applies to: `src/tools/management/commands/dbrestore.py:10-12`

<sub>Found by: spk-reviewer-general-purpose</sub>
</details>


### Agent Coverage
| spk-reviewer-security | security | 1 | 4 | 0 |
| spk-reviewer-backend-python | backend | 1 | 3 | 1 |
| spk-reviewer-python-quality | code_quality | 1 | 1 | 0 |
| spk-reviewer-general-purpose | general | 1 | 7 | 2 |

JBurkinshaw (approved):

PR #759, opened 2026-08-17
[M568] Add table_backup management command to archive old records to S3

gridcell:
Adds a `table_backup` management command that streams a table's older records to line-delimited JSON, uploads the file to S3, and can optionally delete the records it backed up. This gives us a repeatable way to trim growing tables like `api_archivedrecord` without losing the data.

- New `table_backup` command: filters records by a datetime column (default `created_on`, default cutoff one year old), with optional `--start-date`/`--end-date` bounds
- Streams rows through a server-side cursor and gzips as it writes, so memory stays flat regardless of table size; Postgres `to_jsonb` handles serialization including PostGIS and `jsonb` columns
- Safety guards: table/column names validated against the catalog before interpolation, `--delete` requires a verified S3 upload and a single-column primary key, and existing S3 keys get an incrementing suffix rather than being overwritten
- Flags for local-only runs, uncompressed output, keeping the local file, and a custom output directory
- Adds test coverage for filtering, compression, key collision handling, and delete behaviour

## Summary by CodeRabbit

* **New Features**
  * Added a `table_backup` management command for exporting database records within configurable date ranges.
  * Supports plain or gzip-compressed newline-delimited JSON files.
  * Backups can be saved locally, uploaded to S3, or both, with upload verification and unique filenames.
  * Optional deletion removes only successfully backed-up records in batches.
  * Added validation for tables, date fields, primary keys, output options, and backup conditions.


gridcell (commented):
## AI Code Review

**Assessment:** This is a well-engineered command with strong fundamentals: identifier interpolation is properly hardened against SQL injection (deny-by-default regex plus catalog verification), the upload-before-delete ordering is sound, the half-open date interval avoids gaps/overlap across scheduled runs, and to_jsonb gives correct PostGIS/jsonb fidelity. Code quality is clean with no issues raised by the quality reviewer. The recurring theme across the backend, database, and security reviewers, independently corroborated, is that the command is not yet safe to run unattended against large or actively-written tables: the batched --delete runs inside one long transaction, the full PK list is buffered in memory, deletion is not re-validated against the original filter (so a row updated -- or whose datetime field moves past the cutoff -- between export and delete is silently lost with only a stale snapshot on S3), and the --delete integrity gate is an S3 existence check rather than a content/size comparison. None of these are rated severe today (the intended use case is append-only archival tables), but they should be addressed or explicitly documented as constraints before this tool is used more broadly. Test coverage is good in spirit (asserts on real bytes, gzip magic numbers, and a real PostGIS model) but has real gaps at exactly the multi-batch, PK-rejection, and race-safety boundaries the code's own docstrings claim to guarantee.

| 0 | 12 | 0 | 8 | 6 |

**Files reviewed:** 2 / 2

<sub>Note: two earlier CodeRabbit threads on this PR (alternate-datetime-field test coverage; local-only runs overwriting a previous backup) are already resolved in the current head and were not re-raised.</sub>

### Cross-Cutting Concerns

- Two independent reviewers (backend and database) converged on the same three issues from different angles -- the single all-encompassing transaction.atomic() around batched deletes, the unbounded in-memory PK list, and the export/delete race -- which increases confidence these are real, not agent-specific noise.
- The command's own docstrings make explicit consistency claims (memory stays flat; records added between backup and delete are never caught up in it) that the test suite does not verify at the boundary conditions that would actually exercise them (multi-batch deletes, concurrent inserts during the write). Closing the test gaps in CR-009/CR-010/CR-011/CR-012 would substantially increase confidence in the safety claims the code makes for itself.
- None of the destructive-path concerns (CR-001, CR-002, CR-003) are mutually exclusive fixes -- re-applying where_sql at delete time (CR-001), per-batch commits (CR-003), and a real integrity check before delete (CR-002) are complementary and could reasonably be addressed together in one follow-up rather than three separate PRs.
- The python-quality reviewer found the code itself (structure, naming, docstrings, style) to be clean with zero issues; all findings here are about operational safety margins for large/actively-written tables rather than code quality.

### General Findings

<details>
<summary><strong>CR-011</strong> &middot; Warning &middot; <code>test_gap</code> — The backup/delete race-safety guarantee is asserted in a docstring but never proven by a test</summary>

`_delete_records`'s docstring states records are deleted exactly by the primary keys collected during the backup scan, 'so records added between the backup and the delete are never caught up in it.' No test in the suite inserts a new matching record after the backup write completes but before delete runs. A regression that reimplemented delete by re-running the same date-range WHERE clause (instead of using the captured `pks` list) would pass every existing test, since `test_delete_removes_only_backed_up_records` and `test_unverified_upload_aborts_delete` only check that pre-existing 'recent' records survive, not that records created mid-run are protected. Note this is the flip side of CR-001: the current design is untested for the case it claims to guard (post-export inserts) and separately vulnerable to a case it does not address (post-export updates).

> **Suggestion:** Add a test that wraps or monkeypatches `_write_backup` to create one additional matching-aged record partway through the write, then asserts that record still exists after `--delete` completes, proving deletion is scoped to the pks captured at backup time and not to a live re-query.

> **Why it matters:** This is presented as a deliberate consistency guarantee in the code but is currently only documented, not verified; without a test it can silently regress during refactors of `_delete_records` or `_write_backup`.

Applies to: `src/tools/management/commands/table_backup.py`, `src/api/tests/test_table_backup_command.py`

</details>

### What's Done Well

- **Identifier interpolation is properly hardened — no injection path found** — All three reviewers that looked at the SQL independently cleared this. `IDENTIFIER_PATTERN` rejects anything containing a quote, dot, space, semicolon or hyphen before it is ever concatenated, so the surrounding double quotes can never be broken out of; the same function is applied to the table, the `--datetime-field`, and the catalog-derived pk column, so even a hostile column name in the database fails closed rather than injecting. All value-level inputs — dates and the pk batches in `_delete_records` — are bound as parameters, and the S3 key and local filename are built solely from the validated identifier plus strftime output, which closes off path traversal too. `test_invalid_identifier_is_rejected` locks the behaviour in with a real injection payload. Deleting strictly by the pk list captured during export is also the right call for rows inserted after the export began — though it does not protect rows updated during that window (see CR-001), so it is a partial rather than complete consistency guarantee. <sub>(CR-021, `src/tools/management/commands/table_backup.py`)</sub>
- **Upload is verified before any deletion is allowed to proceed** — Checking `s3.file_exists` after `s3.upload_file` and raising a `CommandError` rather than proceeding to delete is the right safeguard for a command whose whole point is 'delete only once the backup is safely off the host'. See CR-002 for deepening this check beyond mere existence. <sub>(CR-022, `src/tools/management/commands/table_backup.py`)</sub>
- **Half-open [start, end) interval avoids gaps and overlap across successive scheduled runs** — Using `< end_date` with an inclusive `>= start_date` means a follow-up run using the previous run's end_date as its start_date covers exactly the adjoining range with no duplicate or skipped rows at the boundary — a detail that's easy to get wrong and is handled correctly here. <sub>(CR-023, `src/tools/management/commands/table_backup.py`)</sub>
- **S3 mock models the upload/exists relationship faithfully, not just call counts** — The `s3_mock` fixture backs both `upload_file` and `file_exists` with the same mutable `keys` set, so the pre-upload collision check and the post-upload verification see consistent state, and tests can pre-seed collisions realistically. This is a much more faithful stand-in than a bare `Mock()` and lets the suite verify actual behavior — which key gets chosen — rather than only that `upload_file` was called. <sub>(CR-024, `src/api/tests/test_table_backup_command.py`)</sub>
- **Gzip content is verified by decoding real bytes, not just by filename** — `test_backup_is_gzipped_by_default` checks the actual gzip magic-number bytes on disk and decodes them, and `test_no_compress_writes_plain_ndjson` confirms the inverse. This avoids the common pitfall of asserting on the `.gz` extension alone while the file itself is unreadable garbage or plain text. <sub>(CR-025, `src/api/tests/test_table_backup_command.py`)</sub>
- **Geometry serialization test uses a real PostGIS-backed model instead of a mock** — `test_geometry_column_serializes` exercises the `to_jsonb` path against the real `Site` table with an actual PostGIS geometry column — exactly the scenario the command's docstring calls out as the reason for delegating JSON encoding to Postgres. Real coverage of a path that would be easy to break silently with a Python-side encoder regression. <sub>(CR-026, `src/api/tests/test_table_backup_command.py`)</sub>

### Agent Coverage

| spk-reviewer-security | security | 2 | 5 | 0 |
| spk-reviewer-backend-python | backend | 1 | 7 | 0 |
| spk-reviewer-database | database | 1 | 10 | 0 |
| spk-reviewer-tests | testing | 2 | 9 | 0 |
| spk-reviewer-python-quality | code_quality | 2 | 0 | 0 |

src/tools/management/commands/table_backup.py
```diff
@@ -0,0 +1,418 @@
...
+            with connection.cursor() as cursor:
+                for offset in range(0, len(pks), DELETE_BATCH_SIZE):
+                    batch = pks[offset : offset + DELETE_BATCH_SIZE]
+                    placeholders = ", ".join(["%s"] * len(batch))
+                    cursor.execute(
+                        f"DELETE FROM {quoted_table} WHERE {pk_column} IN ({placeholders})",
+                        batch,
+                    )
+                    deleted += cursor.rowcount
+        return deleted
```
gridcell: **Warning** · `concurrency` · confidence: high

**Delete-by-captured-pk doesn't re-check the filter, so a row updated (or moved past the cutoff) between export and delete is deleted using a stale snapshot**

`_write_backup` captures the pk list and the `to_jsonb(t)` snapshot in one SELECT; `_delete_records` later deletes strictly `WHERE pk IN (...)` with no re-application of `where_sql`, and the gap between the two (S3 upload + verification) can be minutes on a large table. Two distinct data-loss modes fall out of this on a mutable table: (1) a row's non-key columns are updated after being read but before delete — the backup holds the stale version, yet the current (updated) row is unconditionally deleted, permanently discarding the newer state with no record of it anywhere; (2) the row's own datetime field is bumped past the cutoff in that window — it is no longer 'old' by the filter's own definition, but it still gets deleted because deletion is keyed purely on the previously-captured pk list. This isn't purely theoretical: the command's own help text demonstrates `--datetime-field=updated_on --delete` as a supported combination, which is precisely the scenario where a legitimate concurrent update creates this race.

> **Suggestion:** Re-apply the same `where_sql` (parameterized on the same start/end bounds) to the DELETE, e.g. `DELETE FROM t WHERE pk IN (...) AND <where_sql>`, so a row that no longer qualifies is left alone. For full protection against losing an in-flight update to a still-qualifying row, consider `SELECT ... FOR UPDATE` during export or running export+delete inside a single REPEATABLE READ/SERIALIZABLE transaction. If neither is feasible, explicitly document that this command is only safe for tables whose old records are never mutated after creation.

> **Trade-offs:** Re-checking `where_sql` at delete time is cheap and safe. Adding FOR UPDATE or a single long-running transaction spanning the full export+delete would hold locks (and block vacuum) for the entire export duration, a real cost on a large table that needs weighing against the data-loss risk for the specific tables this command targets.

> **Why it matters:** Silent, permanent loss of the most recent state of a row is the worst failure mode for a data archival tool, and it is not caught by the upload-verification safeguard because the backup file itself is simply stale for that row.

<sub>Found by: spk-reviewer-backend-python, spk-reviewer-database · CR-001</sub>

src/tools/management/commands/table_backup.py
```diff
@@ -0,0 +1,418 @@
...
+            if not s3.file_exists(settings.AWS_BACKUP_BUCKET, key):
+                raise CommandError(
+                    f"Upload of {key} could not be verified; no records have been deleted"
+                )
+            uploaded = True
+            self.stdout.write("Upload complete")
+
+        if delete:
+            deleted = self._delete_records(table, pks)
+            self.stdout.write(f"Deleted {deleted} record(s) from {table}")
```
gridcell, line 250: **Warning** · `security` · confidence: high

**Irreversible bulk DELETE is gated only by an S3 existence check, with no size/integrity comparison and no dry-run or confirmation**

`--delete` permanently removes every backed-up row, and the only safety net is `s3.file_exists()` (a head_object call). That proves an object exists at the key, not that it contains the rows just written. The check also passes if a concurrent run uploaded a different object to that key in between, or if a transfer produced a short/garbled object. `head_object` already returns `ContentLength`, so a real integrity comparison is essentially free. There is also no `--dry-run` and no confirmation for the destructive path, so a mistyped `--end-date` (a future date is accepted — only `start_date >= end_date` is rejected) silently deletes the whole table after backing it up.

> **Suggestion:** Compare the uploaded object against the local file before deleting: expose `head_object` in `api/utils/s3.py` and check `ContentLength` against `os.path.getsize(file_path)`, raising `CommandError` on mismatch. Add `--dry-run` (report the count/date range and exit before writing or deleting) and require an explicit `--noinput` for `--delete`, otherwise prompt with the table name, record count and date range, mirroring how `flush`/`squashmigrations` guard destructive actions.

> **Trade-offs:** Requiring confirmation breaks non-interactive/cron invocations unless `--noinput` is passed, so any scheduled job or runbook using `--delete` must be updated in the same change. Size comparison assumes `upload_file` does not transform the body (it does not — no ContentEncoding is set), so it is safe for both the gzip and plain paths.

> **Why it matters:** Deleted rows are unrecoverable from the application side. If verification passes on an object that is not the backup just written, the command reports success while permanently destroying the only copy of the data.

Refs: CWE-354, CWE-357, OWASP ASVS v4.0 8.3.6

src/tools/management/commands/table_backup.py
gridcell: **Warning** · `concurrency` · confidence: high

**All delete batches share one long-lived transaction, negating most of the lock-duration/vacuum benefit of batching**

`_delete_records` wraps the entire `for offset in range(0, len(pks), DELETE_BATCH_SIZE)` loop in a single `transaction.atomic()` block. Batching the DELETE into chunks of 1000 only reduces the size of each SQL statement — it does not reduce lock hold time or transaction age, since nothing commits until the last batch finishes. For a large backfill (exactly the intended use case), this becomes one long-running transaction that holds row locks on every deleted row for the full duration, prevents autovacuum from advancing the database-wide xmin horizon (causing bloat on other actively-written tables), generates a burst of WAL that must be replicated together, and risks hitting `statement_timeout` or `idle_in_transaction_session_timeout` — at which point the whole delete rolls back and all the batching work is discarded.

> **Suggestion:** Commit each batch independently (drop the outer `transaction.atomic()` and rely on autocommit per statement, or wrap only each individual batch). Because the delete is keyed by an already-captured, already-backed-up pk list, this is safe to re-run: a `DELETE ... WHERE pk IN (...)` for pks already deleted in an earlier batch is a no-op.

> **Trade-offs:** Committing per-batch means a mid-run failure leaves some backed-up records deleted and others not, rather than all-or-nothing. Acceptable here since deletes are idempotent by pk and the backup was verified before any deletes started, but worth calling out if any downstream tooling assumes all-or-nothing deletion.

> **Why it matters:** On the large, old tables this command targets, a single multi-minute transaction increases lock contention, cluster-wide bloat and replication pressure well beyond the target table itself.

<sub>Found by: spk-reviewer-backend-python, spk-reviewer-database · CR-003</sub>

src/tools/management/commands/table_backup.py
```diff
@@ -0,0 +1,418 @@
...
+            cursor.execute(sql, where_params)
+            with self._open_backup_file(file_path, compress) as f:
+                for row in cursor:
+                    if pk_column:
+                        pks.append(row[0])
+                        f.write(row[1])
+                    else:
+                        f.write(row[0])
+                    f.write("\n")
+        return pks
```
gridcell, line 519: **Warning** · `performance` · confidence: high

**Primary keys for every backed-up row are buffered in a single Python list when --delete is set, undermining the streaming design**

`_write_backup` streams row data to the file with flat memory via the server-side cursor (a good pattern, called out in the docstring), but when `collect_pks` is set it still appends every PK to a plain Python list (`pks.append(row[0])`) that is fully materialized before `_delete_records` runs. For a backup covering millions of old rows — the scenario this tool is built for — that list can grow into the hundreds of MB, working directly against the 'memory stays flat no matter how large the backup is' goal stated one method up, and risks OOM on sufficiently large tables.

> **Suggestion:** If very large `--delete` runs are expected, delete PKs in batches as they are read from the cursor (interleaved with the write) instead of collecting the full list first, or document/enforce a practical upper bound on record count when `--delete` is used.

> **Why it matters:** A command designed for archiving old, potentially very large tables should not have a step whose memory footprint scales linearly and unboundedly with the number of records being deleted.

<sub>Found by: spk-reviewer-backend-python, spk-reviewer-database · CR-005</sub>

gridcell, line 519: **Warning** · `error_handling` · confidence: high

**No cleanup of the local backup file if streaming fails partway through**

`_write_backup` has no try/except, and `handle()` doesn't wrap the call either. If the cursor loop raises partway (DB connection drop, disk full, permission error, gzip write failure), a truncated/corrupt file is left behind in `options['output_dir']`. On retry, `_unique_filename`/`_unique_key` will see that name as taken and pick a `_1` suffix instead of cleaning up or overwriting the failed attempt, so failed runs silently accumulate partial files in the output directory over time.

> **Suggestion:** Wrap the write in a try/except that removes `file_path` (if it exists) before re-raising, so a failed run doesn't leave a corrupt file behind.

> **Why it matters:** Leftover corrupt files consume disk space on the (often ephemeral/shared) output directory and can be mistaken for a valid backup by anyone inspecting the directory later.

<sub>Found by: spk-reviewer-backend-python · CR-006</sub>

src/tools/management/commands/table_backup.py
gridcell: **Warning** · `test_gap` · confidence: high

**Delete batching across DELETE_BATCH_SIZE is never exercised**

`_delete_records` loops over `pks` in chunks of `DELETE_BATCH_SIZE` (1000) and sums `cursor.rowcount` across iterations. Every test that exercises `--delete` (e.g. `test_delete_removes_only_backed_up_records`) uses only 2-3 records, so the loop body only ever runs once. A regression in the chunking (off-by-one slicing, wrong accumulation of `deleted`, or a batch that silently drops rows) would not be caught by any current test.

> **Suggestion:** Add a test that monkeypatches `tools.management.commands.table_backup.DELETE_BATCH_SIZE` down to a small number (e.g. 2) and creates 3-5 aged records, then asserts all of them are deleted and the reported count matches, forcing the loop to run multiple iterations.

> **Why it matters:** A bug that only manifests when pks span more than one batch (losing the tail batch, or double-counting) would pass CI and only surface in production once the table accumulated over 1000 eligible rows.

<sub>Found by: spk-reviewer-tests · CR-009</sub>

src/tools/management/commands/table_backup.py
```diff
@@ -0,0 +1,418 @@
...
+
+        if not columns:
+            raise CommandError(f"Table '{table}' has no primary key, so --delete cannot be used")
+        if len(columns) > 1:
+            raise CommandError(
+                f"Table '{table}' has a composite primary key ({', '.join(columns)}), "
+                "so --delete cannot be used"
+            )
+        self._pk_column = columns[0]
+        return self._pk_column
```
gridcell, line 386: **Warning** · `test_gap` · confidence: high

**Composite-primary-key and no-primary-key rejection paths are untested**

`_get_pk_column` raises `CommandError` when a table has zero or more than one primary-key columns, but every `--delete` test in the suite uses `ArchivedRecord`, which has a single-column PK. Neither the 'no primary key' branch nor the 'composite primary key' branch is exercised anywhere in the file.

> **Suggestion:** Create a raw test table inside the test (e.g. `connection.cursor().execute('CREATE TABLE ...')` in a fixture with teardown) and assert `CommandError` with the 'no primary key' / 'composite primary key' messages when `--delete` is passed for each case.

> **Why it matters:** These are the guardrails preventing silent misbehavior of `_delete_records` (which assumes a single pk column) against unsupported schemas; a regression here — e.g. picking the first of multiple pk columns instead of rejecting — could delete the wrong rows in production with no test to catch it.

<sub>Found by: spk-reviewer-tests · CR-010</sub>

src/tools/management/commands/table_backup.py
```diff
@@ -0,0 +1,418 @@
...
+        )
+
+    def _open_backup_file(self, file_path, compress):
+        """Compression happens a line at a time as the records stream past, so memory stays flat
+        no matter how large the backup is."""
+        if compress:
+            return gzip.open(
+                file_path, mode="wt", encoding="utf-8", compresslevel=GZIP_COMPRESS_LEVEL
+            )
+        return open(file_path, "w", encoding="utf-8")
```
gridcell, line 472: **Warning** · `security` · confidence: medium

**Full table dump is written world-readable into a predictable path under world-writable /tmp**

The backup contains every column of every selected row (`to_jsonb(t)`), which for MERMAID tables can include profile emails, project data and other PII. It is written with `gzip.open`/`open`, so the file gets mode 0666 & ~umask (typically 0644 — readable by every local user/process), into `/tmp/mermaid` created by `os.makedirs(..., exist_ok=True)` at default mode 0777 & ~umask. The filename is fully predictable, and `_unique_filename()` uses `os.path.exists()`, which returns False for a dangling symlink — so a pre-planted symlink at that path would be followed by the subsequent `open()`. There is also no try/finally around the write: if upload verification fails or `_delete_records()` raises, the plaintext dump is left on disk indefinitely. In the normal Docker deployment (single non-root app user) this is low-impact, but the command is also usable on shared hosts where other processes share /tmp.

> **Suggestion:** Create the directory and file with restrictive modes and refuse to follow pre-existing links: `os.makedirs(output_dir, mode=0o700, exist_ok=True)`, and in `_open_backup_file` use `os.open(file_path, os.O_WRONLY | os.O_CREAT | os.O_EXCL | os.O_NOFOLLOW, 0o600)` and hand the fd to `gzip.open`/`open`. Also consider defaulting `--output-dir` to a per-run `tempfile.mkdtemp()` (mode 0700) rather than a fixed `/tmp/mermaid`.

> **Trade-offs:** O_EXCL makes an existing-file collision fatal rather than silently suffixed; `_unique_filename()` already resolves collisions, so the only behaviour change is that a genuine race errors instead of overwriting — the safer outcome. `mode=0o700` is not applied to an already-existing `/tmp/mermaid`, so an explicit `os.chmod` may be needed.

> **Why it matters:** A full-table export containing PII sitting world-readable in a shared temp directory is a local information-disclosure vector, and the predictable name plus non-exclusive create makes symlink redirection possible on a multi-tenant host.

Refs: CWE-732, CWE-377, CWE-59, CWE-459

src/tools/management/commands/table_backup.py
```diff
@@ -0,0 +1,418 @@
...
+    def _summarize(self, table, datetime_field, where_sql, where_params):
+        quoted_table = quote_identifier(table)
+        quoted_field = quote_identifier(datetime_field)
+        with connection.cursor() as cursor:
+            cursor.execute(
+                f"SELECT COUNT(*), MIN(t.{quoted_field}), MAX(t.{quoted_field}) "
+                f"FROM {quoted_table} t WHERE {where_sql}",
+                where_params,
+            )
+            return cursor.fetchone()
```
gridcell, line 408: **Warning** · `performance` · confidence: medium

**Two full passes over the table (COUNT/MIN/MAX, then the export) plus an ORDER BY, with no index requirement on the datetime column**

`_summarize` does `SELECT COUNT(*), MIN(...), MAX(...)` and `_write_backup` separately does `SELECT ... ORDER BY t.<datetime_field>` — both filtered on the same `where_sql`. If the datetime column has no index, both are full sequential scans, and the second additionally requires a full sort. Because this runs outside an explicit transaction, Django's postgres backend declares a WITH HOLD cursor, meaning the entire sort must complete and be materialized server-side before the first row streams to the client. For the large, old tables this tool is meant to archive, this can be expensive with no guardrail or warning if the column is unindexed.

> **Suggestion:** Document that an index on the datetime field is a prerequisite for running this against large tables, or add a check that warns when no such index exists. Also consider whether `ORDER BY` is actually required — delete doesn't depend on ordering — and drop it if it's only for a nicer output file, since that removes the sort cost.

> **Why it matters:** An unindexed sequential scan plus sort on a multi-million-row table run without warning could take an unexpectedly long time or spill large sort temp files, surprising an operator running this as routine maintenance.

<sub>Found by: spk-reviewer-database · CR-007</sub>

src/tools/management/commands/table_backup.py
```diff
@@ -0,0 +1,418 @@
...
+            cursor.execute(
+                """
+                SELECT a.attname
+                FROM pg_index i
+                JOIN pg_attribute a ON a.attrelid = i.indrelid AND a.attnum = ANY(i.indkey)
+                WHERE i.indrelid = to_regclass(%s) AND i.indisprimary
+                """,
+                [quote_identifier(table)],
+            )
+            columns = [r[0] for r in cursor.fetchall()]
```
gridcell, line 376: **Warning** · `correctness` · confidence: medium

**Primary-key discovery query doesn't distinguish key columns from INCLUDE (non-key) columns in pg_index.indkey**

`a.attnum = ANY(i.indkey)` matches every attribute listed in `indkey`, but `indkey` contains both true key columns and any INCLUDE columns (Postgres has allowed INCLUDE on PRIMARY KEY-backing indexes since PG 11). `pg_index.indnkeyatts` tells you how many of the leading entries in `indkey` are actual key columns; this query ignores it. A single-column PK with an INCLUDE column would be reported as a 2-column 'composite' PK here and `--delete` would be refused — fails safe, but incorrectly.

> **Suggestion:** Restrict to true key columns, e.g. filter with `a.attnum = ANY(i.indkey[0:i.indnkeyatts-1])` (indkey is 0-indexed) or join against `unnest(i.indkey[:i.indnkeyatts])`.

> **Why it matters:** Unlikely to affect existing MERMAID tables today, but as a general-purpose command it will misclassify any table whose PK-backing index later gains an INCLUDE column, blocking `--delete` unnecessarily and confusingly.

<sub>Found by: spk-reviewer-database · CR-008</sub>

src/tools/management/commands/table_backup.py
```diff
@@ -0,0 +1,418 @@
...
+        for n in range(1, MAX_FILENAME_SUFFIX + 1):
+            candidate = f"{stem}_{n}{extension}"
+            if not os.path.exists(os.path.join(output_dir, candidate)):
+                self.stdout.write(f"{filename} already exists; backing up to {candidate} instead")
+                return candidate
+
+        raise CommandError(
+            f"{filename} already exists, as do the first {MAX_FILENAME_SUFFIX} suffixed variants "
+            "of it"
+        )
```
gridcell, line 463: **Warning** · `test_gap` · confidence: medium

**MAX_FILENAME_SUFFIX exhaustion (both S3 and local-file variants) is untested**

`_unique_key` and `_unique_filename` both raise `CommandError` once `MAX_FILENAME_SUFFIX` (1000) suffixed names are all taken. `test_suffix_only_applied_when_key_is_taken` and the incrementing-suffix tests only ever check suffixes `_1`/`_2`, never the exhaustion path.

> **Suggestion:** Patch `tools.management.commands.table_backup.MAX_FILENAME_SUFFIX` to a small value (e.g. 2) in a dedicated test, pre-seed the `s3_mock` key set (or pre-create local files for the `--no-upload` variant) with the base name plus all suffixed variants up to that limit, and assert `CommandError` is raised for both `_unique_key` and `_unique_filename`.

> **Why it matters:** This is the only safety valve preventing the loop from silently returning a wrong/overwritten filename; without a test forcing exhaustion, an off-by-one in `range(1, MAX_FILENAME_SUFFIX + 1)` or a missing raise would go unnoticed.

<sub>Found by: spk-reviewer-tests · CR-012</sub>

src/tools/management/commands/table_backup.py
```diff
@@ -0,0 +1,418 @@
...
+        if value is None or timezone.is_aware(value):
+            return value
+        return timezone.make_aware(value)
+
+    def _validate_table(self, table):
+        """Confirm the table exists in a schema on the search path."""
+        with connection.cursor() as cursor:
+            cursor.execute("SELECT to_regclass(%s)", [quote_identifier(table)])
+            if cursor.fetchone()[0] is None:
+                raise CommandError(f"Table '{table}' does not exist")
```
gridcell, line 335: **Info** · `security` · confidence: high

**to_regclass() accepts any relation kind, not just tables**

`to_regclass()` resolves views, materialized views, sequences, indexes, foreign tables and composite types as well as ordinary tables, so `_validate_table()` confirms 'some relation with this name is on the search_path', not 'this is a table'. Downstream damage is limited — `_get_pk_column()` finds no primary key for a view, so `--delete` is refused with a clear error — but a simple auto-updatable view is worth ruling out explicitly given what `--delete` does. On the safe side: resolution goes through `search_path` consistently in both the validation query and the data queries (no TOCTOU between check and use), and Postgres' 63-byte identifier truncation applies identically to both, so a long name cannot validate one relation and read another.

> **Suggestion:** Add relkind filtering, e.g. `SELECT c.oid FROM pg_class c WHERE c.oid = to_regclass(%s) AND c.relkind IN ('r', 'p')`, and mention in the help text that the table is resolved via search_path.

Refs: CWE-20

<sub>Found by: spk-reviewer-security · CR-014</sub>

src/tools/management/commands/table_backup.py
```diff
@@ -0,0 +1,418 @@
...
+        delete = options["delete"]
+        no_upload = options["no_upload"]
+        compress = not options["no_compress"]
+
+        # Deleting records whose only backup is a local file on an ephemeral container would be
+        # data loss, so the two flags are mutually exclusive.
+        if delete and no_upload:
+            raise CommandError("--delete cannot be used with --no-upload")
+
+        self._validate_table(table)
```
gridcell, line 174: **Info** · `maintainability` · confidence: high

**PK-column validation for --delete happens late, after the count pass, directory creation, and S3 key probing**

`_get_pk_column` (which raises `CommandError` for a missing or composite PK) is only invoked inside `_write_backup`, well after `_summarize` has already scanned the table, `os.makedirs` has run, and `_unique_key`/`_unique_filename` have already made S3 `file_exists` calls. If `--delete` is requested against a PK-less or composite-PK table, all of that work happens before the command fails.

> **Suggestion:** Call `_get_pk_column(table)` upfront in `handle()` — alongside the existing `--delete`/`--no-upload` mutual-exclusion check — when delete is set, so the command fails fast before any table scanning or S3 probing.

<sub>Found by: spk-reviewer-database · CR-016</sub>

src/tools/management/commands/table_backup.py
```diff
@@ -0,0 +1,418 @@
...
+            return key
+
+        for n in range(1, MAX_FILENAME_SUFFIX + 1):
+            candidate = self._s3_key(table, f"{stem}_{n}{extension}")
+            if not s3.file_exists(settings.AWS_BACKUP_BUCKET, candidate):
+                self.stdout.write(f"{key} already exists; backing up to {candidate} instead")
+                return candidate
+
+        raise CommandError(
+            f"{key} already exists, as do the first {MAX_FILENAME_SUFFIX} suffixed variants of it"
```
gridcell, line 441: **Info** · `security` · confidence: medium

**Key-uniqueness check is TOCTOU-racy; rely on bucket versioning as the real anti-overwrite control**

`_unique_key()` probes with `file_exists()` and then uploads later, so two concurrent runs (or a retried cron job overlapping itself) can both select the same key and the second put silently replaces the first — and with `--delete` both runs then delete their rows, so one backup's contents are gone. The `_N` suffix loop is a good defence against the ordinary sequential case, but it cannot be atomic against S3 on its own. Worth confirming for `AWS_BACKUP_BUCKET`, since it now holds the only copy of deleted production records: object versioning enabled, Block Public Access on, default SSE, and a lifecycle/retention policy consistent with the data-retention obligations for the PII in these dumps. `s3.upload_file` sets no ServerSideEncryption or ACL, so bucket defaults govern.

> **Suggestion:** Enable S3 object versioning on `AWS_BACKUP_BUCKET` (and, if available, `IfNoneMatch: "*"` on the put for a truly atomic create-if-absent) so a lost race is recoverable rather than silent.

Refs: CWE-367

<sub>Found by: spk-reviewer-security · CR-013</sub>

src/tools/management/commands/table_backup.py
```diff
@@ -0,0 +1,418 @@
...
+        sql = (
+            f"SELECT {select_columns} FROM {quoted_table} t "
+            f"WHERE {where_sql} ORDER BY t.{quoted_field}"
+        )
+
+        pks = []
+        # A server-side cursor keeps memory flat regardless of how many records match.
+        with connection.chunked_cursor() as cursor:
+            cursor.execute(sql, where_params)
+            with self._open_backup_file(file_path, compress) as f:
```
gridcell: **Info** · `docs` · confidence: medium

**'Memory stays flat' claim understates what a WITH HOLD server-side cursor does on the Postgres side**

Because `_write_backup` is not wrapped in `transaction.atomic()`, the connection is in autocommit mode, so Django's postgres backend creates the named cursor with `withhold=True`. A WITH HOLD cursor's result set is fully materialized into server-side temp storage when the declaring transaction commits — effectively right when `cursor.execute(sql)` returns. So while client-side Python memory does stay flat as the docstring says, Postgres itself must fully execute the query (including the ORDER BY sort) and hold the entire result set before any row streams back, which can mean substantial temp file usage on the DB server for a very large export.

> **Suggestion:** Either wrap the export explicitly in `transaction.atomic()` (which yields a true non-holdable server-side cursor that streams incrementally without full materialization, at the cost of holding the transaction open for the whole export), or update the comment to say this guarantees flat client memory, not flat server-side resource usage.

<sub>Found by: spk-reviewer-database · CR-015</sub>

src/tools/management/commands/table_backup.py
```diff
@@ -0,0 +1,418 @@
...
+
+        Postgres does the JSON encoding via to_jsonb, so every column type the database can
+        output -- including PostGIS geometry and jsonb -- is serialized without needing a
+        Python-side encoder.
+        """
+        quoted_table = quote_identifier(table)
+        quoted_field = quote_identifier(datetime_field)
+        pk_column = self._get_pk_column(table) if collect_pks else None
+
+        select_columns = "to_jsonb(t)::text"
```
gridcell: **Info** · `docs` · confidence: medium

**to_jsonb output is faithful in Postgres but a naive JSON restore path can lose numeric precision or mishandle bytea**

Using `to_jsonb(t)` is a good choice for cross-type fidelity — jsonb/array columns nest correctly, PostGIS geometry serializes as EWKB hex text that is directly restorable via INSERT, and UUIDs and timestamps serialize as parseable strings. Two caveats worth documenting for whoever writes the restore side: (1) NUMERIC/DECIMAL columns become JSON numbers, and a generic parser (Python's `json.loads` without `parse_float=Decimal`) will coerce them to float and can lose precision on restore; (2) bytea columns serialize as a `\x`-prefixed hex string, which round-trips correctly only if the restore code knows to hex-decode that convention.

> **Suggestion:** Note these conventions in the command's docstring or help text (or in the eventual restore tool) so a future restore script parses NUMERIC columns with Decimal-aware JSON parsing and hex-decodes bytea fields explicitly.

<sub>Found by: spk-reviewer-database · CR-018</sub>

src/tools/management/commands/table_backup.py
```diff
@@ -0,0 +1,418 @@
...
+
+        file_path = os.path.join(options["output_dir"], filename)
+
+        pks = self._write_backup(
+            table, file_path, where_sql, where_params, datetime_field, delete, compress
+        )
+        written = len(pks) if delete else record_count
+        self.stdout.write(
+            f"Wrote {written} record(s) to {file_path} "
+            f"({os.path.getsize(file_path) / 1024:,.1f} KiB)"
```
gridcell, line 236: **Info** · `correctness` · confidence: low

**Logged record count for non-delete runs comes from a separate, earlier query than the actual write**

`_summarize` (a COUNT/MIN/MAX query) and `_write_backup`'s SELECT both apply the same `where_sql`, but they run as two separate statements with no shared snapshot. When delete is False, the 'Wrote N record(s)' message uses `record_count` from `_summarize` rather than counting what `_write_backup` actually wrote, so if rows matching the filter are inserted or removed between the two queries, the logged count can diverge from the number of lines actually in the file. The file contents themselves are still correct.

> **Suggestion:** Have `_write_backup` return the number of lines it wrote (it already iterates every row) and report that, so the log line always describes the file that exists.

> **Why it matters:** Low impact — this affects a log line, not the data written — but could be confusing when troubleshooting a discrepancy between the reported count and the file's actual line count.

<sub>Found by: spk-reviewer-backend-python · CR-017</sub>

src/api/tests/test_table_backup_command.py
```diff
@@ -0,0 +1,317 @@
...
+    call_command("table_backup", TABLE, "--output-dir", str(tmp_path), "--keep-local")
+
+    files = _files(tmp_path)
+    assert len(files) == 1
+    rows = _read_ndjson(files[0])
+    assert {r["model"] for r in rows} == {"old_a", "old_b"}
+    assert str(recent[0].pk) not in {r["id"] for r in rows}
+    # Nothing is deleted without --delete
+    assert ArchivedRecord.objects.filter(pk__in=[r.pk for r in old]).count() == 2
+    assert upload.call_count == 1
```
gridcell, line 116: **Info** · `test_gap` · confidence: low

**Round-trip content check stops at top-level fields; nested JSONField content is never verified**

`ArchivedRecord.record` is itself a JSONField. Tests only assert on `r["model"]` and `r["id"]` from the written NDJSON; the nested `record` value produced by Postgres's `to_jsonb(t)` is never checked against the original dict.

> **Suggestion:** In `test_backs_up_only_records_older_than_default_cutoff` (or a new small test), assert `rows[i]["record"] == {"name": "x"}` to confirm nested jsonb columns survive the to_jsonb -> file -> json.loads round trip without being double-encoded as a string or otherwise mangled.

> **Why it matters:** Nested JSON/PostGIS columns are exactly the case the command's docstring calls out as the reason for using to_jsonb instead of a Python encoder; verifying only scalar columns leaves that rationale unverified.

<sub>Found by: spk-reviewer-tests · CR-019</sub>

src/tools/management/commands/table_backup.py
```diff
@@ -0,0 +1,418 @@
...
+    def _build_where(self, datetime_field, start_date, end_date):
+        quoted_field = quote_identifier(datetime_field)
+        clauses = [f"t.{quoted_field} < %s"]
+        params = [end_date]
+        if start_date:
+            clauses.append(f"t.{quoted_field} >= %s")
+            params.append(start_date)
+        # Rows with a NULL datetime field have no age to judge, so they are never backed up.
+        clauses.append(f"t.{quoted_field} IS NOT NULL")
+        return " AND ".join(clauses), params
```
gridcell, line 397: **Info** · `test_gap` · confidence: low

**NULL datetime-field exclusion (IS NOT NULL clause) is not directly tested**

`_build_where` always appends `t.<field> IS NOT NULL`, so rows with a null datetime value are never backed up regardless of the date range. All records in the `aged_records` fixture are created via `auto_now_add`, so this clause is never exercised with an actual NULL value.

> **Suggestion:** Add a case (e.g. via the `--datetime-field=updated_on` path already exercised in `test_alternate_datetime_field`, where `updated_on` can be left NULL for one record) asserting that a record with a NULL value in the filtered column is excluded even though it would otherwise match an open-ended range.

> **Why it matters:** This is a deliberate exclusion rule; without a test it could regress into either backing up NULL-dated rows or crashing on the comparison.

<sub>Found by: spk-reviewer-tests · CR-020</sub>

gridcell:
@coderabbitai review

PR #760, opened 2026-08-21
[M2082] Add cache method when querying data for validators

gridcell:
## Summary by CodeRabbit

* **Performance**
  * Reduced repeated database queries during validation by reusing annotation and sample-unit data within each validation run.
* **Tests**
  * Added coverage for validation caching, cache isolation, query efficiency, and annotation row reuse.

### DIFF
--- a/src/api/submission/validations/validators/base.py
+++ b/src/api/submission/validations/validators/base.py
@@ -95,6 +95,7 @@ def validate_list(func):
 
 class BaseValidator:
     result = None
+    RUN_CACHE_KEY = "_validation_run_cache"
 
     def __init__(self, **kwargs):
         self.context = kwargs or {}
@@ -115,6 +116,16 @@ class BaseValidator:
     @validator_result
     def skip(self, context=None):
         return OK, None, context
+
+    def get_run_cache(self, record, namespace):
+        """Namespaced scratch space shared by every validator in one validation run.
+
+        ValidationRunner hands the same collect record dict to each validator in a
+        run, so the dict doubles as a cache that goes away when the run ends. Only
+        available to validators that receive the serialized record, not the ones
+        declaring requires_instance.
+        """
+        return record.setdefault(self.RUN_CACHE_KEY, {}).setdefault(namespace, {})
 
     def get_value(self, record, key):
         return get_value(record, key, delimiter=".")
--- a/src/api/tests/validators/test_sample_unit_consistency.py
+++ b/src/api/tests/validators/test_sample_unit_consistency.py
@@ -286,3 +286,37 @@ def test_different_quadrat_size_different_value(
     assert result.code == DifferentQuadratSizeValidator.DIFFERENT_QUADRAT_SIZE
     assert result.context["quadrat_size"] == 2.0
     assert result.context["other_quadrat_size"] == 1.0
+
+
+# ==================== per-validation-run caching tests ====================
+
+
+def test_bpqt_consistency_validators_share_one_sibling_fetch(
+    benthic_photo_quadrat_transect1,
+    valid_benthic_pq_transect_collect_record,
+    django_assert_num_queries,
+):
+    """BPQT stacks three of these checks, all against the same quadrat_transect row."""
+    quadrat_transect = valid_benthic_pq_transect_collect_record.data["quadrat_transect"]
+    quadrat_transect["num_quadrats"] = 2
+    quadrat_transect["num_points_per_quadrat"] = 100
+    quadrat_transect["len_surveyed"] = 50
+    valid_benthic_pq_transect_collect_record.save()
+
+    validators = [
+        _get_num_quadrats_validator(),
+        _get_num_points_per_quadrat_validator(),
+        DifferentTransectLengthValidator(
+            protocol_path="data.protocol",
+            site_path="data.sample_event.site",
+            management_path="data.sample_event.management",
+            sample_date_path="data.sample_event.sample_date",
+            len_surveyed_path="data.quadrat_transect.len_surveyed",
+        ),
+    ]
+    record = CollectRecordSerializer(instance=valid_benthic_pq_transect_collect_record).data
+
+    # One sample event lookup and one sibling sample unit fetch, shared by all three.
+    with django_assert_num_queries(2):
+        for validator in validators:
+            assert validator(record).status == OK
--- a/src/api/submission/validations/validators/classification_benthic_photo_quadrat.py
+++ b/src/api/submission/validations/validators/classification_benthic_photo_quadrat.py
@@ -113,6 +113,15 @@ class DuplicateImageValidator(BaseValidator):
 
 class BaseAnnotationValidator(BaseValidator):
     def get_rows(self, collect_record, **kwargs):
+        # The confirmed, unclassified and region validators all need the same rows,
+        # so build them once per validation run.
+        cache = self.get_run_cache(collect_record, "annotation_rows")
+        if "rows" not in cache:
+            cache["rows"] = self._build_rows(collect_record)
+
+        return cache["rows"]
+
+    def _build_rows(self, collect_record):
         cr_id = collect_record.get("id")
 
         num_points_per_quadrat = (collect_record["data"].get("quadrat_transect") or {}).get(
@@ -122,7 +131,9 @@ class BaseAnnotationValidator(BaseValidator):
             num_points_per_quadrat = Classifier.latest().num_points
 
         annos = (
-            Annotation.objects.select_related("point", "point__image")
+            Annotation.objects.select_related(
+                "point", "point__image", "benthic_attribute", "growth_form"
+            )
             .filter(point__image__collect_record_id=cr_id)
             .order_by(
                 "point__image__created_on",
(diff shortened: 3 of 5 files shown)


src/api/submission/validations/validators/base.py
```diff
@@ -116,6 +117,16 @@ def name(self):
...
 
+    def get_run_cache(self, record, namespace):
+        """Namespaced scratch space shared by every validator in one validation run.
+
+        ValidationRunner hands the same collect record dict to each validator in a
+        run, so the dict doubles as a cache that goes away when the run ends. Only
+        available to validators that receive the serialized record, not the ones
+        declaring requires_instance.
+        """
+        return record.setdefault(self.RUN_CACHE_KEY, {}).setdefault(namespace, {})
```
gridcell, line 128: **Info** · `maintainability` · confidence: high

**Run cache stores Django model instances and model classes inside the API-shaped record dict**

Verified the cache cannot currently leak. ValidationRunner.validate() builds `collect_record_dict = self.serializer(instance=collect_record).data` as a fresh local (src/api/submission/validations/base.py:176), only `runner.to_dict()` (version/status/results) is persisted into CollectRecord.validations (src/api/submission/utils.py:258, 301-305), and the record returned to the client is re-serialized from a fresh serializer after the DB update (src/api/submission/utils.py:309). The ingest path (src/api/ingest/utils.py:202) and the validate/submit viewset actions (src/api/resources/collect_record.py:149, 163) both funnel through that same function, so nothing writes the mutated dict back. RUN_CACHE_KEY is an immutable string class attribute, so there is no shared cross-request state. The residual concern is structural rather than exploitable: _validation_run_cache now hides non-JSON-serializable values (SampleEvent instances, model classes as tuple-key members, sibling sample-unit rows) inside a dict whose other keys are all API response fields. Any future code that returns, json.dumps()es, deep-copies, or writes back the validator-facing record dict would either 500 on serialization or emit unintended data. Note also that Sentry is initialized with default include_local_variables=True (src/app/settings.py:332-341), so an exception inside a validator will ship the frame locals, including this dict; the reprs added are low-value and the record payload itself was already in those locals, so this is not a new leak class.

> **Suggestion:** Optional hardening: hold the scratch space off to the side rather than inside the serialized record, e.g. have ValidationRunner create an explicit RunContext and hand it to validators. Failing that, a short comment on RUN_CACHE_KEY stating the dict must never be serialized, persisted, or returned would preserve the invariant for future readers.

> **Why it matters:** Mixing internal ORM objects into a dict shaped like an API response is a latent leakage/serialization footgun if a future change starts returning or storing that dict.

src/api/tests/validators/test_sample_unit_consistency.py
```diff
@@ -286,3 +286,37 @@ def test_different_quadrat_size_different_value(
...
+            sample_date_path="data.sample_event.sample_date",
+            len_surveyed_path="data.quadrat_transect.len_surveyed",
+        ),
+    ]
+    record = CollectRecordSerializer(instance=valid_benthic_pq_transect_collect_record).data
+
+    # One sample event lookup and one sibling sample unit fetch, shared by all three.
+    with django_assert_num_queries(2):
+        for validator in validators:
+            assert validator(record).status == OK
```
gridcell, line 322: **Info** · `maintainability` · confidence: low

**Exact query-count assertion is a reasonable but implementation-coupled regression guard**

django_assert_num_queries(2) pins the test to the exact number of SQL statements the three validators issue together (one sample-event lookup, one sibling fetch). That's appropriate here since the test's whole purpose is to lock in the caching optimization, but it will need updating any time the query shape changes for unrelated reasons -- adding select_related fields that don't add a query is fine, but any future added lookup will break this test even if caching still works correctly per-namespace.

<sub>Found by: spk-reviewer-tests · CR-003</sub>

gridcell (commented):
## AI Code Review

**Assessment:** This is a well-tested, correctly-scoped performance refactor that adds a per-validation-run cache (BaseValidator.get_run_cache) to eliminate repeated DB queries for sample-event and annotation lookups during collect-record validation, with no confirmed correctness, security, or code-quality defects. All four specialist agents traced the cache lifecycle end-to-end and confirmed it never leaks into persisted data or API responses, and that cache keys are fully scoped to the ORM filter identity (no cross-tenant/cross-record risk). The only notable gaps are a missing query-count regression test for the cached-None branch of _get_sample_event, and a PR-description/diff mismatch (the description references image-storage-bucket and upload-logging fixes that are not present in this diff). Nothing here warrants blocking the PR.

| 0 | 1 | 0 | 3 | 2 |

**Files reviewed:** 5 / 5

### Cross-Cutting Concerns
- get_run_cache mutates the collect-record dict in place by adding a `_validation_run_cache` key; verified safe rather than an open risk -- ValidationRunner.validate() builds a fresh serializer dict per call, only `runner.to_dict()` (version/status/results) is ever persisted, and every caller (src/api/submission/utils.py, src/api/ingest/utils.py, src/api/resources/collect_record.py) funnels through that same path, so the cache key never leaks into CollectRecord.validations or any API response.
- select_related tuples are used as members of dict cache keys throughout this diff; confirmed hashable at every call site, so there is no `TypeError: unhashable type` risk.
- This PR is a pure in-memory query-optimization refactor -- no migration or schema change was needed or included, consistent with the diff.

<details>
<summary><strong>CR-001</strong> &middot; Warning &middot; <code>test_gap</code> — No query-count test covers the cached-None branch of _get_sample_event across multiple validators</summary>

test_bpqt_consistency_validators_share_one_sibling_fetch proves caching works when a sample event and sibling rows exist. There is no equivalent test for the case where _get_sample_event finds no matching SampleEvent (.first() returns None). The production code caches None correctly via `if key not in cache`, but nothing in the test suite verifies that a second validator sharing the same record and the same (site, management, sample_date) key avoids re-querying when the lookup misses. test_different_num_quadrats_no_existing only checks the functional OK result with a single validator, not the query count across a stacked run.

> **Suggestion:** Add a test analogous to test_bpqt_consistency_validators_share_one_sibling_fetch that runs two or more SampleEventConsistencyValidator subclasses against a record whose site/management/sample_date do not match any existing SampleEvent, asserting django_assert_num_queries(1) (one miss lookup, cached and reused, with no sibling fetch since sample_event is None).

> **Why it matters:** A future change that swaps `if key not in cache` for a truthiness check (e.g. `if not cache.get(key)`) would silently break caching for the not-found case and re-issue a query per validator; no current test would catch that regression.

Applies to: `src/api/tests/validators/test_sample_unit_consistency.py`, `src/api/submission/validations/validators/sample_unit_consistency.py`

</details>

<details>
<summary><strong>CR-004</strong> &middot; Info &middot; <code>docs</code> — PR description references changes that are not present in this diff</summary>

The PR description (CodeRabbit-generated summary) claims bug fixes for image-storage bucket configuration and image-upload diagnostic logging, in addition to the validation-caching performance work. The actual diff for this PR touches only src/api/submission/validations/validators/base.py, classification_benthic_photo_quadrat.py, sample_unit_consistency.py, and their two test files -- there are no image-bucket or upload-logging changes anywhere in the diff.

> **Suggestion:** Trim the PR description to match the actual diff (validation caching + tests) so reviewers and future readers of the merge commit aren't misled about what shipped in this PR.

<sub>Found by: spk-reviewer-aggregator</sub>
</details>

<details>
<summary><strong>CR-005</strong> &middot; Praise &middot; <code>security</code> — Cache keys include every field used in the underlying ORM filter, so no rows can cross a tenant boundary</summary>

Both caches are keyed on the full filter identity rather than a partial key, which is exactly what prevents a cached-row authorization bypass. _get_sample_event keys on (str(site_id), str(management_id), sample_date) -- the complete set of filter kwargs -- and the ids come from valid_id(), which rejects anything that is not a UUID. _get_sibling_sample_units keys on (model, sample_event_path, str(sample_event.pk), select_related), and the sample_event it keys on was itself resolved through the site+management+date filter, so a validator cannot be handed sample units belonging to a different site, management, sample event, or project. model and sample_event_path come from the hardcoded PROTOCOL_CONFIG dict behind a `protocol not in self.PROTOCOL_CONFIG` guard, so an attacker-supplied protocol value cannot reach an arbitrary model or ORM lookup path. None results are cached via `if key not in cache` rather than a falsy check, avoiding a repeated-query/cache-bypass inconsistency. Growth is bounded by the number of distinct protocol/model pairs a single record's validation set touches, so there is no unbounded-growth DoS vector.

Applies to: `src/api/submission/validations/validators/sample_unit_consistency.py`

<sub>Found by: spk-reviewer-security</sub>
</details>

<details>
<summary><strong>CR-006</strong> &middot; Praise &middot; <code>test_gap</code> — Caching tests pair query-count with identity assertions, avoiding a false-positive-prone pattern</summary>

test_get_rows_cached_per_validation_run combines django_assert_num_queries(0) with an `is` identity check, and test_get_rows_not_cached_across_validation_runs combines `is not` with `==`. This double-barreled approach is exactly right for testing a cache: query count alone could pass for the wrong reason (e.g. an empty result set), and identity alone doesn't prove no query ran. Verified these tests pass against the current implementation and genuinely fail if either the run-scoped cache or the cross-run isolation is broken.

Applies to: `src/api/tests/validators/test_classification_benthic_photo_quadrat.py`

</details>

### Agent Coverage
| spk-reviewer-security | security | 5 | 2 | 0 |
| spk-reviewer-backend-python | backend | 3 | 0 | 0 |
| spk-reviewer-python-quality | code_quality | 5 | 0 | 0 |
| spk-reviewer-tests | testing | 2 | 3 | 0 |

saanobhaai (approved):

PR #761, opened 2026-09-02
[M714] API Key implementation

gridcell:


src/api/admin/base.py
```diff
@@ -1,396 +1,621 @@
...
+        # Revoking retires a key and keeps the row, which is what answers
+        # "what did this credential do, and when did it stop working". Delete
+        # throws that away, so it stays with the superuser.
+        return request.user.is_superuser and super().has_delete_permission(request, obj)
+
+    def get_actions(self, request):
+        actions = super().get_actions(request)
+        if not request.user.is_superuser:
+            actions.pop("generate_replacement_keys", None)
+        return actions
```
gridcell, line 231: **Warning** · `security` · confidence: high

**APIKeyAdmin restricts add/delete to superusers but not change, letting non-superuser staff extend or reactivate a credential**

`has_add_permission` and `has_delete_permission` are both gated on `request.user.is_superuser` (with the comment "Minting a credential for any profile is a superuser action (C5)"), but there is no `has_change_permission` override, and the editable form fields are ("profile", "name", "expires_at", "never_expires", "is_active") with only "profile" becoming read-only on change (via `get_readonly_fields`, which appends it only when `obj is not None`). Any staff user granted the standard Django `api.change_apikey` permission (e.g. via a group, without being a superuser) can therefore open an existing key in the change form and: (1) extend `expires_at` on someone else's key or tick `never_expires`, turning a bounded credential into a permanent one, defeating the stated mitigation that `expires_at` defaults to a year rather than to never; or (2) set `is_active=True` with a future `expires_at` on a key the daily maintenance sweep deactivated at expiry, restoring a credential whose secret may be in the hands of a former integrator. Since `is_usable` is `is_active and revoked_at is None and not is_expired`, this is materially the same power as issuing a new credential for that profile. Revocation is correctly one-way (`revoked_at` is read-only and `authenticate` rejects on `revoked_at is not None` independently of `is_active`), so a truly revoked key cannot be resurrected this way; the gap is specifically expiry and active-state, the other two enforcement points.

> **Suggestion:** Add a `has_change_permission` override consistent with `has_add_permission`/`has_delete_permission` (superuser-only): return `request.user.is_superuser and super().has_change_permission(request, obj)`. If non-superuser staff need to browse and revoke keys, keep change open instead and make `expires_at`, `never_expires`, and `is_active` read-only for non-superusers via `get_readonly_fields`, leaving the `revoke_keys` action as the one thing they can do.

> **Trade-offs:** Fully superuser-gating change removes the ability for non-superuser staff to run the `revoke_keys` action (Django admin actions require change permission), pushing incident response onto superusers; the read-only-fields variant avoids that but requires verifying no existing workflow relies on non-superuser staff editing these fields.

> **Why it matters:** Issuance is deliberately restricted to superusers because a key is the whole of its profile's access; leaving expiry and activation editable by any staff user with change permission lets a lower-privileged account prolong or reinstate that access without minting anything, undermining the model's stated security story.

<sub>Found by: spk-reviewer-security, spk-reviewer-backend-python · CR-001</sub>

src/api/auth_backends.py
```diff
@@ -175,13 +179,160 @@ def _validate_app(self, client_id):
...
+
+        cache_key = f"apikey:last_used:{api_key.key_id}"
+        if cache.get(cache_key):
+            return
+        cache.set(cache_key, True, self.last_used_throttle)
+        now = timezone.now()
+        # update() to avoid touching updated_on/updated_by on every call
+        APIKey.objects.filter(pk=api_key.pk).update(last_used_at=now, last_used_ip=ip)
+        api_key.last_used_at = now
+        api_key.last_used_ip = ip
```
gridcell, line 343: **Warning** · `error_handling` · confidence: high

**_touch writes an unvalidated client-controlled string into an inet column**

`_touch` writes `ip` (which comes straight from the client-controlled `X-Forwarded-For`) into `APIKey.last_used_ip`, a `GenericIPAddressField` (Postgres `inet`), via `.update()`. `.update()` performs no model validation, so `X-Forwarded-For: unknown` (which some proxies and CDNs legitimately emit) or any junk value reaches Postgres as an `inet` literal and raises `invalid input syntax for type inet`; a junk value containing `:` raises `ValidationError` from `clean_ipv6_address`. Either way an otherwise successfully authenticated request becomes a 500. `_get_client_ip`'s own fallback string "unknown" has the same problem when neither header nor `REMOTE_ADDR` is present. The ordering also masks the failure: `cache.set(cache_key, ...)` happens before the write, so the exception occurs once per key per throttle window and then silently stops for 60 seconds, which makes it awkward to diagnose.

> **Suggestion:** Validate before writing, and never let audit bookkeeping break the request: normalize the value with Python's `ipaddress.ip_address(value)`, catching `ValueError` and falling back to `None`, before it reaches `.update()`, and wrap the `.update()` call in a try/except around `DatabaseError` that logs a warning rather than propagating. Consider moving `cache.set` after the write so a transient failure retries rather than being masked for the rest of the throttle window.

> **Why it matters:** A client-supplied header can turn a valid, authenticated API request into a 500, which is both an availability bug for the key holder and a trivially reachable error path for an attacker holding any valid key.

src/api/tests/test_api_key_auth.py
```diff
@@ -0,0 +1,208 @@
...
+        APIKeyAuthentication().authenticate(_request(path=f"/v1/projects/?access_token={raw}"))
+        is None
+    )
+
+
+def test_secret_comparison_is_constant_time(key_pair):
+    _, raw = key_pair
+    with patch("api.utils.apikeys.hmac.compare_digest", return_value=True) as mock_compare:
+        _authenticate(raw)
+    mock_compare.assert_called_once()
```
gridcell, line 172: **Warning** · `test_gap` · confidence: high

**test_secret_comparison_is_constant_time does not verify constant-time behavior**

This test patches hmac.compare_digest to always return True and calls _authenticate(raw) with the correct raw key once, then asserts mock_compare.assert_called_once(). That only proves secret_matches invokes hmac.compare_digest at all; it says nothing about timing, and it never exercises the comparison with a wrong secret to show the call happens uniformly regardless of match/mismatch (e.g. no short-circuit on a bad secret). The test name promises a security property that the assertions do not check.

> **Suggestion:** Either rename the test to reflect what it actually verifies (e.g. test_secret_match_uses_hmac_compare_digest), or strengthen it to also call with a mismatched secret and assert compare_digest is still invoked with two full-length arguments in both cases, which is the actual guarantee against timing side-channels.

> **Why it matters:** A test that claims to cover a security property but doesn't creates false confidence and could mask a future regression (e.g. someone adding an early-return short-circuit before the compare_digest call) that this test would still pass.

<sub>Found by: spk-reviewer-tests · CR-008</sub>

src/api/auth_backends.py
```diff
@@ -175,13 +179,160 @@ def _validate_app(self, client_id):
...
+
+        raw = smart_str(auth[1])
+        try:
+            env, key_id, secret = parse_api_key(raw)
+        except ValueError:
+            raise self._fail(request, "malformed_key", None, ip=ip)
+
+        # Counted separately from the IP: one bad key rotating through hosts is
+        # the same incident, and one bad host should not lock out a good key.
+        self._check_rate_limit(request, "key_id", key_id)
```
gridcell, line 260: **Warning** · `security` · confidence: medium

**Per-key_id failure counter is checked before secret verification, so it can lock out a valid key**

The `key_id` scope is checked before the secret is verified, and the check applies to every request bearing that `key_id`, including the legitimate holder's request with the correct secret. `key_id` is not itself a secret: it is logged in `[apikey.failed_auth]`/`[apikey.rate_limited]` lines, displayed in the admin list and CSV export, and is the third segment of any key that has ever been shared or pasted. Anyone who learns a live `key_id` can send 10 bad-secret requests for that `key_id` and have the real client answered 429 for the remainder of the minute, repeating each minute for a sustained outage. `test_throttled_ip_is_throttled_even_with_a_good_key` and `test_key_id_is_throttled_across_ips` encode this as intended behaviour. The documented rationale ("one bad key rotating through hosts is the same incident, and one bad host should not lock out a good key") argues for counting per key_id, which is right, but does not require blocking a request whose secret verifies.

> **Suggestion:** Keep recording failures per `key_id`, but do not gate on the key_id counter before proof of possession. Either move the `key_id` gate to run only after `secret_matches` has been evaluated (block only when the secret is wrong), or compute `key_over_limit = self.failure_limiter.retry_after("key_id", key_id)` up front and only raise `exceptions.Throttled(wait=key_over_limit)` inside the branch where `secret_matches(secret, api_key.secret_hash)` returns False, rather than before checking the secret at all.

> **Trade-offs:** Checking the key_id limit after the database lookup means an attacker who is over the key_id limit still costs one indexed SELECT and one SHA-256 per request. That is a small, bounded cost, and the per-IP gate (once the X-Forwarded-For trust issue is addressed) still caps it.

> **Why it matters:** As written, knowledge of a non-secret identifier is enough to deny service to a legitimate machine client indefinitely, which inverts the intent of a limiter whose docstring states "a client presenting a good key is never throttled by this".

src/api/admin/base.py
```diff
@@ -1,396 +1,621 @@
...
+            request,
+            format_html(
+                "API key <strong>{}</strong> issued for {}. Copy it now: it is not "
+                "stored and cannot be shown again.<br><code>{}</code>",
+                key.name,
+                key.profile,
+                raw,
+            ),
+            messages.SUCCESS,
+        )
```
gridcell, line 293: **Warning** · `security` · confidence: medium

**Raw key is delivered through the messages framework, which by default persists it in a browser cookie**

`_show_raw_key` claims the raw key "reaches the browser through the messages framework (cookie, falling back to the session) for exactly one response, which is the same exposure as rendering it in the page, and nothing stores it." `MESSAGE_STORAGE` is not set in src/app/settings.py, so Django's default `FallbackStorage` tries `CookieStorage` first: the message, containing the full raw key, is base64-encoded and signed (not encrypted) into a messages cookie, written to the browser's cookie jar (on disk for a non-session cookie), and sent back to the server in the Cookie header of the next request before being cleared. That has none of the properties of rendering it once in the response body. Practical consequences: the secret lands in a file on the operator's disk, is replayed to the server at least once where it can be captured by request-header logging, WAF sampling, or a Sentry event that includes cookies, and survives in the cookie jar if the operator never loads another admin page.

> **Suggestion:** Force session-backed storage for this one message so the secret never enters a cookie, e.g. write it via `django.contrib.messages.storage.session.SessionStorage` explicitly, or set `MESSAGE_STORAGE = "django.contrib.messages.storage.session.SessionStorage"` in settings. A stronger option is to stop using messages for the secret entirely and render it from an intermediate response_add/custom admin template so it exists only in the one response body.

> **Trade-offs:** Switching MESSAGE_STORAGE globally moves all admin messages into the session (requires a session, and changes behaviour for any flow relying on cookie storage); scoping the change to this one message avoids that.

> **Why it matters:** The raw key is the credential in full; the cookie round-trip puts it at rest on a client disk and in flight in a request header where infrastructure logging is much more likely to capture it than a response body.

src/api/tests/test_api_key_permissions.py
```diff
@@ -0,0 +1,250 @@
...
+        user=SimpleNamespace(profile=profile1),
+        GET=QueryDict(),
+        POST=QueryDict(),
+        data={},
+        query_params=QueryDict(),
+        META={},
+        auth=key,
+    )
+
+    assert MockRequest.load_request(stub).auth is key
```
gridcell, line 250: **Warning** · `test_gap` · confidence: medium

**Mock/ViewRequest tests only check attribute copying, not the isinstance-based routing that consumes it**

test_view_request_carries_the_api_key and test_mock_request_keeps_the_api_key both assert only that `.auth is key` survives construction of a ViewRequest / MockRequest. Neither test calls permissions.get_request_api_key(...) (or resources.base.get_request_profile(...)) against the resulting object to confirm the isinstance(api_key, APIKey) check those functions perform actually succeeds on these mock objects. A future change that wraps or copies the auth object (e.g. serializes/deserializes it, or substitutes a proxy) would keep `.auth is key` passing here while breaking the real consumer silently.

> **Suggestion:** Extend these two tests to also assert get_request_api_key(view_request) is key and/or get_request_profile(view_request) == profile1, so the test exercises the actual production function these mocks exist to support rather than just the attribute copy.

> **Why it matters:** get_request_api_key's isinstance check is the single point that separates a key-authenticated request from a JWT one for write attribution; testing only the surrounding plumbing gives false confidence that the real check is covered.

<sub>Found by: spk-reviewer-tests · CR-007</sub>

gridcell (commented):
## AI Code Review

**Assessment:** PR #761 introduces a new APIKey machine-credential system (model, migration, an Auth0-parallel authentication backend, Django admin, per-identifier rate limiting, a daily maintenance command, and CloudWatch alarms) into what was previously a JWT-only API. The credential design itself (SHA-256 hash of a CSPRNG secret, constant-time comparison, fail-closed 401s, no serializer ever exposing the hash) is sound. The two findings most worth the author's attention are a live privilege gap in `APIKeyAdmin` (`has_change_permission` is not superuser-gated like add/delete, so any staff user with the standard `change_apikey` permission can un-expire or reactivate a key) and a `key_id`-scoped failure counter that is checked before secret verification, so anyone who learns a non-secret `key_id` can lock out a legitimate integration indefinitely. Both the per-IP rate limiter and the `last_used_ip` audit column trust a client-controlled `X-Forwarded-For` value, which is also spoofable and can turn a valid authenticated request into a 500 when written unvalidated into an `inet` column. Test coverage of the new authentication paths is broad but has specific gaps around the sync/push write-attribution path, mock-based tests that don't exercise the real consumer function, and no CI safeguard against model/migration drift. No documentation was added for the new `Authorization: ApiKey` scheme.

| Severe | Warning | Question | Info |
|--------|---------|----------|------|
| 0 | 11 | 4 | 9 |

**Files reviewed:** 29 / 29

### Cross-Cutting Concerns

- No documentation anywhere in the PR describes the new Authorization: ApiKey scheme for API consumers; no .md file was changed at all, despite this being a new externally-facing authentication mechanism.
- A log-marker string contract couples application code to infrastructure: auth_backends.py, models/base.py, utils/apikeys.py, and api_key_maintenance.py emit bracketed markers (e.g. [apikey.failed_auth], [apikey.rate_limited], [apikey.created], [apikey.revoked]) that iac/stacks/constructs/alerts.py matches by literal string. Renaming a marker in application code would silently break the CloudWatch alarms with no compile-time or test-time signal.
- src/api/admin/base.py is a roughly 620-line diff that is mostly an unrelated whole-file reformat with the new admin classes interleaved, which makes the actual change hard to review in isolation and will complicate future git blame on this file.
- The rate limiter and the last_used_at write throttle both depend on a shared cache; correctness under multiple gunicorn workers rests on CACHES being the shared DatabaseCache rather than a per-process backend, which is currently the case but is an implicit dependency worth calling out explicitly in code comments.
### Additional Findings

<details>
<summary><strong>CR-003</strong> &middot; Warning &middot; <code>security</code> — Rate limiting and audit data are keyed off a spoofable X-Forwarded-For value</summary>

`_get_client_ip` takes the leftmost `X-Forwarded-For` entry. `config/webapp.nginxconf` sets `proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;`, which appends the peer address to whatever the client sent, and the ALB in front of it does the same, so the leftmost element is entirely attacker-controlled; no trusted-proxy-count handling (e.g. ProxyFix) exists in settings.py. This was pre-existing code used only for log lines previously; this PR promotes it to (a) the identifier the failure limiter enforces on for the "ip" scope and (b) a value written to the `APIKey.last_used_ip` column. Consequences: (1) limiter evasion -- an attacker can send a fresh `X-Forwarded-For` value on each request and never accumulate an IP-scoped count, so the "ip" scope only constrains well-behaved clients, and can similarly rotate a fake `key_id`-adjacent header value per attempt to blunt correlation; (2) targeted lockout -- conversely, an attacker who knows a legitimate client's egress IP can send bad-key requests with `X-Forwarded-For: <victim-ip>` and have the victim answered with 429 for the rest of the minute, repeatable indefinitely, since `_check_rate_limit(request, "ip", ip)` runs before any credential work and this blocks requests bearing a valid key; (3) forged audit trail -- `last_used_ip` (and the `ip=` field in `[apikey.failed_auth]`, which feeds the CloudWatch metric filter) is whatever the caller says it is, so the one field that would answer "where was this leaked key used from" cannot be relied on during an incident.

> **Suggestion:** Derive the client IP from the right-hand side of the header rather than the left: with a known number of trusted proxies (nginx + ALB), take the Nth-from-last entry behind a `settings.TRUSTED_PROXY_COUNT`, or adopt an explicit trusted-proxy helper (e.g. django-ipware with `IPWARE_META_PRECEDENCE_ORDER` and proxy count set) and reject/ignore the header when the count does not match. If the exact hop count is not stable, prefer `REMOTE_ADDR` for enforcement and treat XFF as untrusted metadata only.

> **Trade-offs:** Counting proxy hops is deployment-coupled: if the hop count is wrong, every caller collapses onto one internal address and the per-IP limiter becomes a global limiter (or stops matching real clients). Validate against a real ALB request before rolling out, and keep the per-key_id scope as the primary defence in the interim.

> **Why it matters:** An enforcement decision and an audit field both derive from a header any caller can set, so the per-IP limit is bypassable by the attacker it targets and usable as a denial-of-service against a legitimate integration, while the recorded origin of key usage is untrustworthy during incident response.

Applies to: `src/api/auth_backends.py:21-23`

<sub>Not diff-anchorable: `src/api/auth_backends.py` lines 21-23 are not part of this PR's diff.</sub>

<sub>Found by: spk-reviewer-security, spk-reviewer-backend-python</sub>

</details>

<details>
<summary><strong>CR-006</strong> &middot; Warning &middot; <code>test_gap</code> — Sync/push updated_by attribution for API-key writes is never tested end-to-end</summary>

`resources/sync/utils.create_view_request` now copies `request.auth` onto the ViewRequest so that `resources/base.get_request_profile` can tell an API-key request from a JWT one during a push. The only test that touches this (test_api_key_permissions.py::test_view_request_carries_the_api_key) builds a hand-rolled `_Req` stub class directly and asserts `create_view_request(_Req()).auth is key`; it never goes through an actual /v1/push/ (or protocol-specific push) HTTP call made with an ApiKey-authenticated APIClient, and never asserts that a record written via that path actually gets `updated_by` set to the key's profile. Every other write-attribution test in the suite goes through the direct REST create endpoint, not the sync push pipeline that create_view_request feeds into.

> **Suggestion:** Add an integration test that POSTs to the sync push endpoint (e.g. /v1/push/) using an ApiKey-authenticated APIClient with a payload for a writable record type, then asserts the persisted record's updated_by equals the key's profile. This exercises the real request object DRF builds, not a stand-in.

> **Why it matters:** The push path is a distinct code path from direct REST writes (it builds its own ViewRequest per record) and is exactly where a request.auth plumbing regression would silently misattribute writes made via API key over sync, without any test catching it.

Applies to: `src/api/tests/test_api_key_permissions.py`, `src/api/resources/sync/utils.py`, `src/api/resources/base.py`

</details>

<details>
<summary><strong>CR-009</strong> &middot; Warning &middot; <code>maintainability</code> — Cache-clearing and key-creation fixtures are duplicated across five test files instead of shared via fixtures/apikeys.py</summary>

An identical 4-line clear_cache autouse fixture (cache.clear() before and after each test) is copy-pasted verbatim in test_api_key_auth.py, test_api_key_public.py, test_api_key_permissions.py, test_api_key_ratelimit.py, and test_api_key_lifecycle.py. Likewise, a near-identical _make_key/key_pair APIKey-creation helper is independently reimplemented in test_api_key_auth.py, test_api_key_public.py, test_api_key_permissions.py, test_api_key_lifecycle.py, and test_api_key_admin.py, each with slightly different default expiries (365 days, 30 days, or default_expires_at()). fixtures/apikeys.py already exists as the shared module for this feature (it currently only holds the api_key_audit_logs fixture and api_key_audit_lines helper), but it is not used to centralize either of these.

> **Suggestion:** Move clear_cache into fixtures/apikeys.py as a shared autouse fixture (or session-level conftest fixture) and add a single parametrizable api_key_pair/make_api_key fixture there for all five files to import, rather than each file hand-rolling its own.

> **Why it matters:** Duplicated setup logic drifts over time (as already seen in the differing default expiries) and increases the chance that a fix to the cache-isolation or key-creation logic in one file is missed in the others.

Applies to: `src/api/tests/fixtures/apikeys.py`, `src/api/tests/test_api_key_auth.py`, `src/api/tests/test_api_key_public.py`, `src/api/tests/test_api_key_permissions.py`, `src/api/tests/test_api_key_ratelimit.py`, `src/api/tests/test_api_key_lifecycle.py`, `src/api/tests/test_api_key_admin.py`

</details>

<details>
<summary><strong>CR-010</strong> &middot; Warning &middot; <code>test_gap</code> — No CI safety net catches model/migration drift (tests run --no-migrations, no makemigrations --check)</summary>

src/pytest.ini / make test / .github/workflows/_test.yml all run pytest with --no-migrations, which builds the test schema directly from the current model state rather than by applying src/api/migrations/0134_apikey.py. .github/workflows/pr.yml runs isort/ruff/ruff-format and a cdk-diff job, but there is no python manage.py makemigrations --check --dry-run (or equivalent) step anywhere in CI. For this PR, the migration and the APIKey model in src/api/models/base.py were hand-verified (and confirmed with sqlmigrate) to agree field-by-field, but nothing in CI would have caught it if they had not, and nothing will catch drift the next time either file is edited without the other; none of the new API key tests would catch it either, since they all run against the model-derived schema rather than the migration.

> **Suggestion:** Add a python manage.py makemigrations --check --dry-run step to the lint or test CI job so any future model change without a matching migration (or vice versa) fails the build before merge.

> **Trade-offs:** This will also flag any pre-existing drift elsewhere in the codebase the first time it runs, which may need a one-off cleanup migration before the check can be turned on as a hard gate.

> **Why it matters:** With --no-migrations tests, an out-of-sync migration is invisible until it is applied to a real database (staging/production), where a missing column, wrong max_length, or missing constraint surfaces as a runtime error or silent data-shape mismatch instead of a CI failure.

Applies to: `.github/workflows/_test.yml`, `.github/workflows/pr.yml`, `src/pytest.ini`, `src/api/migrations/0134_apikey.py`, `src/api/models/base.py`

<sub>Found by: spk-reviewer-tests, spk-reviewer-database</sub>

</details>

<details>
<summary><strong>CR-012</strong> &middot; Question &middot; <code>maintainability</code> — No metric filter/alarm for the apikey.created/revoked/expired/stale audit markers</summary>

This PR adds an audit trail via audit_logger (api.apikeys) that emits [apikey.created], [apikey.revoked], and [apikey.expired]/[apikey.stale]. Those go to console_info at INFO with propagate=False, but land on the same stdout stream as the 'api' logger's console handler, so they should reach the same api_log_group as the failed_auth/rate_limited lines. Yet alerts.py only adds filters for [apikey.failed_auth] and [apikey.rate_limited]. There is no metric for a burst of revocations (bulk incident response, or a bug revoking keys unexpectedly), nor for the daily maintenance command silently failing to run: a cron that never fires produces no log lines and so no alarm.

> **Suggestion:** If revocation/expiry/stale volume is intentionally audit-only, say so in the section comment so a future reader does not read it as an oversight. Otherwise add an alarm on [apikey.revoked] for unexpected revocation bursts, plus a scheduled-job heartbeat (CloudWatch Events rule + alarm) confirming api_key_maintenance actually ran.

> **Why it matters:** A mass-revocation incident, or a bug revoking the wrong keys, currently produces no alert; operators would only find it by searching logs. The PR itself calls revocation the counterpart of apikey.created for audit purposes, so its absence from monitoring may not be deliberate.

Applies to: `iac/stacks/constructs/alerts.py`, `src/api/models/base.py`, `src/api/admin/base.py`, `src/tools/management/commands/api_key_maintenance.py`

<sub>Found by: spk-reviewer-devops</sub>

</details>

<details>
<summary><strong>CR-014</strong> &middot; Question &middot; <code>performance</code> — is_active/expires_at/revoked_at/last_used_at have no indexes backing the maintenance-command and admin filters</summary>

api_key_maintenance.py filters on is_active, expires_at, revoked_at, last_used_at, and created_on (none indexed beyond the automatic FK index on profile_id and the unique index on key_id), and APIKeyAdmin.list_filter filters on expires_at/is_active/revoked_at as well. Given this is a one-row-per-credential table expected to stay small (per-profile, superuser-issued keys), full-table scans are likely negligible for the foreseeable future, and the maintenance command is a daily batch job, not a per-request hot path.

> **Suggestion:** No action needed now; if the table grows into the tens/hundreds of thousands of rows, consider a partial index such as an index on expires_at where is_active to keep the daily sweep and admin filters fast.

> **Why it matters:** Left unaddressed at scale, the daily maintenance command's two queries and the admin list filters would degrade to sequential scans, though this is not a near-term concern at the expected row counts for this table.

Applies to: `src/api/migrations/0134_apikey.py`, `src/tools/management/commands/api_key_maintenance.py`

<sub>Found by: spk-reviewer-database</sub>

</details>

<details>
<summary><strong>CR-023</strong> &middot; Info &middot; <code>test_gap</code> — ImproperlyConfigured propagation from a misconfigured ENVIRONMENT is untested at the authenticate() call site</summary>

APIKeyAuthentication.authenticate calls get_environment_label() and lets ImproperlyConfigured propagate uncaught when settings.ENVIRONMENT is not one of the known values, per the inline comment "let it surface as a 500". test_api_key_utils.py tests get_environment_label and the api.checks startup check raising/catching ImproperlyConfigured, but no test in test_api_key_auth.py exercises APIKeyAuthentication.authenticate itself with a misconfigured settings.ENVIRONMENT to confirm the exception actually surfaces (uncaught) through the authentication flow rather than being swallowed somewhere in DRF's exception handling.

> **Suggestion:** Add a small test that overrides settings.ENVIRONMENT to an invalid value and asserts APIKeyAuthentication().authenticate(...) raises ImproperlyConfigured (not AuthenticationFailed), confirming it is not accidentally caught as a 401.

Applies to: `src/api/auth_backends.py`, `src/api/tests/test_api_key_auth.py`

</details>

### Agent Coverage

| spk-reviewer-security | security | 29 | 8 | 0 |
| spk-reviewer-backend-python | backend | 28 | 5 | 0 |
| spk-reviewer-tests | testing | 11 | 7 | 0 |
| spk-reviewer-database | database | 2 | 4 | 0 |
| spk-reviewer-devops | infrastructure | 1 | 4 | 0 |
| spk-reviewer-python-quality | code_quality | 29 | 0 | 0 |

iac/stacks/constructs/alerts.py
```diff
@@ -363,6 +363,76 @@ def __init__(
...
+                metric=apikey_rate_limit_metric.metric(
+                    statistic="Sum",
+                    period=Duration.minutes(5),
+                ),
+                threshold=1,
+                evaluation_periods=1,
+                comparison_operator=cw.ComparisonOperator.GREATER_THAN_OR_EQUAL_TO_THRESHOLD,
+                treat_missing_data=cw.TreatMissingData.NOT_BREACHING,
+            )
+        )
```
gridcell, line 434: **Warning** · `maintainability` · confidence: medium

**Rate-limit alarm fires on the first 429 with no aggregation window**

threshold=1 with GREATER_THAN_OR_EQUAL_TO_THRESHOLD and evaluation_periods=1 over a 5-minute Sum means this alarm, and its paired OK notification, fires and resolves on the first throttled API-key request. The pattern is copied verbatim from Auth0RateLimitAlarm, where the trigger is calls to the Auth0 Management API: a rare, server-initiated event. For API keys the limiter (FailureRateLimiter('apikey', limit=10, window=60)) is tripped by ordinary client mistakes (a rotated key with a client not yet updated, a stale integration retrying on 401), and each qualifying request keeps re-emitting [apikey.rate_limited] for as long as the client retries, so one misconfigured integration can flap this alarm repeatedly.

> **Suggestion:** Raise the threshold (>=3-5 in 5 minutes) or widen the period, or at minimum document that this alarm is expected to fire on routine client misconfiguration so on-call does not treat every occurrence as a security incident.

> **Why it matters:** An alarm that pages on routine, expected client behavior trains responders to ignore it, defeating its purpose as a leaked-or-guessed-key signal.

<sub>Found by: spk-reviewer-devops · CR-011</sub>

src/api/permissions.py
```diff
@@ -74,6 +74,25 @@ def _load():
...
+    `request.auth` is the JWT string for an Auth0 caller and None for an
+    anonymous one, so the isinstance check is what separates a key-authenticated
+    request from every other kind.
+
+    A key grants exactly its profile's access, so nothing in the permission
+    classes needs this to decide yes or no. It is here for the callers that
+    care *how* a request was authenticated rather than as whom: attributing a
+    write (resources/base.get_request_profile) and refusing to let a key mint
+    another key.
+    """
```
gridcell, line 89: **Question** · `docs` · confidence: medium

**Docstring references a 'refuse to let a key mint another key' guard that doesn't exist**

get_request_api_key's docstring says it exists partly for "refusing to let a key mint another key". Searching the diff and the rest of the codebase, get_request_api_key is only consumed by resources/base.py (for updated_by attribution); there is no API endpoint or permission check anywhere that mints an APIKey via DRF at all -- issuance only happens through the Django admin (session auth, superuser-gated), which never goes through this function. The referenced guard appears to be aspirational/not-yet-implemented.

> **Suggestion:** Either implement the described guard if a DRF-facing key-issuance endpoint is planned, or trim the docstring to describe only the current, real caller (write attribution) so it doesn't describe a protection that isn't actually in place.

> **Why it matters:** A docstring describing a security guard that doesn't exist could give a future reader false confidence that key-minting-by-key is already prevented at this layer.

<sub>Found by: spk-reviewer-backend-python · CR-015</sub>

src/api/models/base.py
```diff
@@ -214,3 +220,77 @@ class Meta:
...
+    on the project being touched, so a key reaches exactly the data its owner
+    reaches and no more - and loses access the moment a membership goes away,
+    with no scope list to keep in step.
+
+    The trade is blast radius: a leaked key is the whole of its profile's
+    access. That is why issuing one is a superuser action, why `expires_at`
+    defaults to a year rather than to never, and why revocation is one call.
+    """
+
+    profile = models.ForeignKey("Profile", related_name="api_keys", on_delete=models.CASCADE)
```
gridcell, line 239: **Question** · `correctness` · confidence: low

**profile CASCADE deletes the audit-bearing APIKey row along with the profile**

profile = models.ForeignKey("Profile", related_name="api_keys", on_delete=models.CASCADE) means deleting a Profile silently deletes every APIKey row that named it, including revoked_at/revoked_reason/last_used_at history. The class docstring and src/api/signals/apikeys.py explicitly call this out as intentional ("a deleted profile takes its keys with it through the FK cascade"), and the parallel audit_logger.info(...) calls at creation/revocation mean the audit trail survives independently in logs even if the DB row is gone. Given that reasoning is already documented and a log-based fallback exists, this is a question rather than a defect: is log retention alone considered sufficient for "which keys existed, for whom, for how long" if a Profile row is ever hard-deleted (as opposed to the more common deactivation path), or would on_delete=models.SET_NULL plus a tombstone be worth reconsidering for that specific case?

<sub>Found by: spk-reviewer-database · CR-013</sub>

src/api/auth_backends.py
```diff
@@ -175,13 +179,160 @@ def _validate_app(self, client_id):
...
+        # ImproperlyConfigured here is a deployment error, not a client error;
+        # let it surface as a 500. api.checks catches it at startup.
+        if env != get_environment_label():
+            raise self._fail(
+                request,
+                "wrong_environment",
+                key_id,
+                detail="key issued for a different environment",
+                ip=ip,
+            )
```
gridcell, line 271: **Info** · `security` · confidence: high

**wrong_environment is the one failure with a distinguishable response detail**

Every other rejection path relies on _fail's opaque default ("Invalid API key"), with the comment "which check failed is not the caller's business, and saying so would narrow a guessing attack". The environment mismatch path passes detail="key issued for a different environment", so it is the single reason a caller can distinguish from the outside. The information disclosed is minimal (the caller already knows the env label it sent, and there is no cross-environment guessing gain since each environment has its own database), but it is inconsistent with the stated design, and it does confirm to a scanner that a well-formed mmd_<env>_... shape was parsed and reached the env comparison. If the distinct message is intentional operator ergonomics, it is worth saying so in the comment; otherwise drop the detail and rely on the [apikey.failed_auth] reason=wrong_environment log line, which already carries the diagnosis to the side that should have it.

<sub>Found by: spk-reviewer-security · CR-017</sub>

src/api/utils/ratelimit.py
```diff
@@ -0,0 +1,85 @@
...
+        self.namespace = namespace
+        self.limit = limit
+        self.window = window
+
+    def _window_start(self, now):
+        return int(now // self.window) * self.window
+
+    def _cache_key(self, scope, identifier, window_start):
+        return f"{self.namespace}:fail:{scope}:{identifier}:{window_start}"
+
```
gridcell, line 40: **Info** · `security` · confidence: high

**Fixed window allows a 2x burst at the boundary and is resettable by cache culling**

Two known properties of the fixed-window design, both consistent with the module's "counting is deliberately approximate" note: (1) boundary burst -- because the window is wall-clock aligned and part of the cache key, a caller can place limit failures just before window_start + window and another limit immediately after, i.e. 20 failures in a fraction of a second; given the counter exists to cap wasted work rather than to stop guessing a 256-bit secret, 2x is not materially different from 1x; (2) counter reset via cache eviction -- the default cache is DatabaseCache, whose culling (MAX_ENTRIES/CULL_FREQUENCY) deletes a random fraction of rows when the table exceeds the limit, and which is also cleared wholesale by anything calling cache.clear(); a busy site can therefore drop live failure counters. The consequence worth tracking is on the observability side: the mermaid-<env>-apikey-rate-limited alarm fires on [apikey.rate_limited] lines, so eviction-induced resets mean the alarm undercounts sustained abuse. If the limiter is ever repurposed as a real security control, it should move to a sliding window on a cache with atomic ops and no culling.

<sub>Found by: spk-reviewer-security · CR-018</sub>

src/api/models/base.py
```diff
@@ -214,3 +220,77 @@ class Meta:
...
+    name = models.CharField(max_length=100)
+    key_id = models.CharField(max_length=12, unique=True, db_index=True)
```
gridcell, line 241: **Info** · `style` · confidence: high

**key_id declares both unique=True and db_index=True, which is redundant but confirmed harmless**

key_id = models.CharField(max_length=12, unique=True, db_index=True) sets db_index=True alongside unique=True. Running sqlmigrate api 0134_apikey against the live model confirms this does not produce a second, redundant index: Postgres generates exactly one UNIQUE constraint plus the standard varchar_pattern_ops index, identical in shape to AuthUser.user_id and Application.client_id elsewhere in this same file, which declare unique=True alone with no db_index. So there is no functional, storage, or write-time cost here, only a superfluous option that is inconsistent with how the sibling models in this file express the same constraint.

> **Suggestion:** If consistency with sibling models is desired, drop the redundant db_index=True in a future migration; treat this as a style/consistency cleanup rather than a performance fix, since the current declaration has no measurable cost.

<sub>Found by: spk-reviewer-backend-python, spk-reviewer-database · CR-019</sub>

src/api/tests/test_api_key_lifecycle.py
```diff
@@ -0,0 +1,221 @@
...
+    assert "actor=someone@example.com" in logged[0]
+    assert "reason=manual" in logged[0]
+
+
+def test_revoke_from_a_signal_names_the_system_as_actor(profile1, project1, api_key_audit_logs):
+    key, _ = _make_key(profile1)
+    key.revoke(PROFILE_DEACTIVATED)
+
+    logged = api_key_audit_lines(api_key_audit_logs, "revoked")
+    assert "actor=system" in logged[0]
```
gridcell, line 94: **Info** · `readability` · confidence: high

**Test name claims to exercise the deactivation signal but calls revoke() directly instead**

test_revoke_from_a_signal_names_the_system_as_actor calls key.revoke(PROFILE_DEACTIVATED) directly rather than going through the revoke_api_keys_for_deactivated_profile signal receiver (which is exercised separately, and correctly, by test_deactivating_a_profile_revokes_its_keys just above it in the same file). It only confirms that revoke()'s default actor parameter is "system" regardless of who calls it, true for any caller, not specific to the signal path. The name suggests signal coverage that this test does not provide.

> **Suggestion:** Rename to something like test_revoke_defaults_actor_to_system, or fold this assertion into test_deactivating_a_profile_revokes_its_keys where the signal is actually triggered.

<sub>Found by: spk-reviewer-tests · CR-022</sub>

src/app/settings.py
```diff
@@ -174,6 +174,7 @@
 REST_FRAMEWORK = {
     "DEFAULT_AUTHENTICATION_CLASSES": (
         "api.auth_backends.JWTAuthentication",
+        "api.auth_backends.APIKeyAuthentication",
         "rest_framework.authentication.SessionAuthentication",
     ),
     "DEFAULT_PERMISSION_CLASSES": ("rest_framework.permissions.IsAuthenticatedOrReadOnly",),
```
gridcell, line 180: **Info** · `correctness` · confidence: medium

**Multi-space ApiKey headers 401 from the JWT backend, bypassing API key failure counting and audit logging**

With DEFAULT_AUTHENTICATION_CLASSES ordered JWTAuthentication, APIKeyAuthentication, SessionAuthentication, JWTAuthentication.authenticate calls get_jwt_token(request) first, and get_jwt_token raises AuthenticationFailed("Invalid Authorization header. Credentials string should not contain spaces.") whenever the header splits into more than two parts. So Authorization: ApiKey foo bar never reaches APIKeyAuthentication. The outcome is still fail-closed (401), which is what test_header_with_spaces_is_401 asserts, and no work is done. The side effects are that such attempts emit no [apikey.failed_auth] line, are not counted by the failure limiter, and are invisible to the new mermaid-<env>-apikey-failed-auth metric filter; the "malformed_header" branch is only reachable for a bare Authorization: ApiKey (one token) under this class ordering. Putting APIKeyAuthentication before JWTAuthentication in the tuple would let the key backend claim its own scheme first and close both gaps.

<sub>Found by: spk-reviewer-security · CR-016</sub>

src/api/signals/apikeys.py
```diff
@@ -0,0 +1,35 @@
...
+    with it through the FK cascade. This hook is here so that if one is ever
+    added, the keys stop working at the same moment the person does, instead of
+    outliving the account silently.
+    """
+
+    if getattr(instance, "is_active", True):
+        return
+
+    for key in APIKey.objects.filter(profile_id=instance.pk, revoked_at__isnull=True):
+        key.revoke(PROFILE_DEACTIVATED)
```
gridcell, line 35: **Info** · `maintainability` · confidence: medium

**Deactivation-revocation signal is currently a no-op and would miss bulk updates**

revoke_api_keys_for_deactivated_profile early-returns via getattr(instance, 'is_active', True). Profile has no is_active field today, so this getattr always resolves to the default True and the handler never reaches the APIKey.objects.filter(...) loop; the docstring explicitly acknowledges this is groundwork for a future flag. One thing the docstring does not call out: even once such a flag exists, this is a post_save receiver, so a bulk Profile.objects.filter(...).update(is_active=False) would not fire it at all (Django does not send per-instance signals for QuerySet.update()), silently leaving those profiles' keys live.

> **Suggestion:** When Profile.is_active (or equivalent) is introduced, document/enforce that deactivation must go through .save() (or add a corresponding hook on whatever bulk deactivation path is chosen) so this signal cannot be silently bypassed.

> **Why it matters:** Not a live bug today since the field does not exist, but leaving the bulk-update gap undocumented means a future implementer could reasonably assume this signal fully covers deactivation and use .update() for a batch job, silently leaving stale keys valid.

<sub>Found by: spk-reviewer-backend-python · CR-020</sub>

iac/stacks/constructs/alerts.py
```diff
@@ -363,6 +363,76 @@ def __init__(
...
+                metric=apikey_failed_auth_metric.metric(
+                    statistic="Sum",
+                    period=Duration.minutes(5),
+                ),
+                threshold=20,
+                evaluation_periods=1,
+                comparison_operator=cw.ComparisonOperator.GREATER_THAN_THRESHOLD,
+                treat_missing_data=cw.TreatMissingData.NOT_BREACHING,
+            )
+        )
```
gridcell, line 401: **Info** · `correctness` · confidence: low

**Failed-auth alarm threshold inherited from Auth0 without accounting for the per-identifier throttle**

The threshold of 20-in-5-minutes and the "possible credential stuffing" description are copied from the Auth0 alarm, but for API keys _fail() stops being reached for a given ip/key_id as soon as the limiter throttles it (10 failures/60s per identifier): _check_rate_limit short-circuits before _fail() runs, as the PR's own comment notes. Reaching 20 failed_auth events in 5 minutes from a single IP or key_id is therefore capped well below what the threshold implies, and effectively requires multiple distinct identifiers. Not necessarily wrong, since distributed guessing across many IPs is still the real threat, but worth confirming the number was chosen with this interaction in mind rather than inherited unchanged.

<sub>Found by: spk-reviewer-devops · CR-021</sub>

iac/stacks/constructs/alerts.py
```diff
@@ -363,6 +363,76 @@ def __init__(
...
+        # client) and wants its own line in Slack. Markers come from
+        # APIKeyAuthentication in auth_backends.py.
+
+        apikey_failed_auth_metric = logs.MetricFilter(
+            self,
+            "ApiKeyFailedAuthMetricFilter",
+            log_group=api_log_group,
+            filter_pattern=logs.FilterPattern.literal('"[apikey.failed_auth]"'),
+            metric_namespace=f"MERMAID/{env_id}/APIKey",
+            metric_name="FailedAuthentications",
```
gridcell, line 379: **Info** · `style` · confidence: low

**Mixed ApiKey/APIKey casing between construct IDs and metric namespace**

Construct IDs use 'ApiKey' (ApiKeyFailedAuthMetricFilter, ApiKeyFailedAuthAlarm) while the metric namespace uses 'APIKey' (MERMAID/{env_id}/APIKey) and the Django model is APIKey. Minor, but a mixed convention in a file that otherwise keeps naming consistent (Auth0*, Sentry*, Rds* each pick one casing).

<sub>Found by: spk-reviewer-devops · CR-024</sub>

PR #762, opened 2026-09-04
M714

gridcell:
## Summary by CodeRabbit

* **New Features**
  * Added self-service API key creation and management, including filtering, renaming, expiration, and revocation.
  * Added API key authentication for machine clients using Bearer credentials.
  * API keys inherit the owner’s access permissions and expose secrets only once at creation.
* **Security**
  * Added rate limiting and monitoring for failed API key authentication.
  * Added safeguards preventing secrets from appearing in API responses, exports, or logs.
* **Documentation**
  * Documented API key usage, lifecycle, management endpoints, and restrictions.
* **Maintenance**
  * Added automated handling and reporting for expired or stale keys.


gridcell:
@coderabbitai review

gridcell (commented):
## AI Code Review

**Assessment:** This PR's core credential design is sound (256-bit secrets, SHA-256 hashing with constant-time comparison, fail-closed defaults, correct FK/on_delete choices, a migration that matches the model exactly), but the review surfaced twelve warning-level gaps concentrated in three areas: the API key security perimeter (secret_hash reachable via the generic `?fields=` parameter, a spoofable X-Forwarded-For value used both as rate-limit identity and as an unvalidated write to an `inet` column, an admin permission that lets `api.add_apikey` holders mint a key for an arbitrary profile, a rate-limit design that lets a third party lock out a known key_id, several GET-only endpoints where a revoked/forged key is silently ignored rather than rejected, a non-secure cookie carrying the one-time raw key, a stored-XSS vector in `ProfileAdmin.linked_email`, and no ceiling on key lifetime or count); a shared performance gap where neither the admin list filters nor the daily maintenance command have supporting indexes on APIKey; and two test-coverage gaps (the rate limiter's concurrent-write race path, and admin CSV-export leak checks that assert on configuration rather than executed output). None of the previously addressed review threads recur in the current findings set - the items raised here are all newly identified. No finding rises to severe/blocking; all are contained by the fact that most require an already-elevated permission (admin `api.add_apikey`) or a client/proxy misconfiguration.

| Severe | Warning | Question | Info |
|--------|---------|----------|------|
| 0 | 12 | 2 | 9 |

**Files reviewed:** 31 / 31

### Cross-Cutting Concerns
- Several independent findings (secret_hash reachable via `?fields=`, a spoofable X-Forwarded-For rate-limit identity, key_id-based lockout of a legitimate credential, and authenticators being cleared entirely on certain GET-only viewsets) all trace back to the same root cause: the stated 'a present-but-bad key always fails closed with a 401' invariant was only enforced in the primary authentication code path, not in the adjacent shared frameworks (the generic field-selection mechanism, the rate limiter's identity model, and per-method authenticator overrides) that this feature now depends on. A follow-up audit of every consumer of these shared frameworks, rather than patching each surfaced instance individually, would likely catch further instances of the same gap.
- The admin surface (src/api/admin/base.py) concentrates three independent security-relevant findings in one file (unrestricted profile impersonation via `api.add_apikey`, a raw key delivered through a non-Secure cookie, and a stored-XSS vector in `ProfileAdmin.linked_email` that an `api.add_apikey`-holding staff session could exploit to self-escalate). Since the whole file is newly added/reformatted in this diff, a dedicated second security pass focused only on this admin module would likely be more efficient than resolving these as unrelated line items.
- The database and backend reviewers independently converged on the same missing-index gap from two different entry points (the daily maintenance command's queries, and the admin list view's filters), confirming that both consumers read the same unindexed columns and should be covered by a single follow-up migration rather than two separate fixes.
- The tests agent's admin CSV-export gap (asserting on configuration rather than executed output) pairs with the security agent's `?fields=secret_hash` and raw-key-cookie findings: all three point to the same policy gap of testing that secret_hash-adjacent surfaces are configured not to leak, rather than testing that their actual rendered output never contains the digest or raw secret. A shared 'assert-on-rendered-output' test utility for anything touching `secret_hash` would close all three gaps at once.

### General Findings

<details>
<summary><strong>CR-002</strong> &middot; Warning &middot; <code>security</code> — Rate-limit identity comes from the client-controlled leftmost X-Forwarded-For entry, so the IP limiter is trivially bypassed</summary>

`_get_client_ip` takes `HTTP_X_FORWARDED_FOR.split(",")[0]`. Behind an ALB/nginx, the leftmost entry is whatever the client sent; the proxy appends its own observation to the right rather than replacing the header. An attacker hammering bad credentials can therefore send a fresh `X-Forwarded-For: 1.2.3.<n>` per request and never accumulate failures in the `ip` scope. The `key_id` scope does not compensate: a malformed key records only the IP scope (`_fail` guards on `if key_id`), and an attacker who varies the key_id gets a fresh counter each time. The result is that the failure limiter, whose stated purpose is to make bad-credential hammering loud and cheap, imposes no bound on an attacker who bothers to set one header, while still fully applying to well-behaved clients.

> **Suggestion:** Derive the client IP from a trusted position in the chain rather than the leftmost entry: take the rightmost entry, or the Nth-from-right based on the known number of trusted proxies, or have nginx overwrite `X-Forwarded-For` with `$remote_addr` and read only that. For example:
> 
> ```python
> TRUSTED_PROXY_DEPTH = 1  # settings
> 
> def _get_client_ip(request):
>     xff = request.META.get("HTTP_X_FORWARDED_FOR", "")
>     if xff:
>         parts = [p.strip() for p in xff.split(",") if p.strip()]
>         idx = -min(len(parts), settings.TRUSTED_PROXY_DEPTH)
>         return parts[idx]
>     return request.META.get("REMOTE_ADDR") or ""
> ```
> 
> Also record the `key_id` scope for well-formed-but-unknown key ids so id rotation is at least visible.

> **Trade-offs:** Rightmost-entry logic depends on the exact proxy depth in each environment; if the depth is wrong the recorded IP becomes the load balancer's and all callers pool into one bucket, which would over-throttle. Pin the depth per environment and verify against a real request in dev before prod. Note `_get_client_ip` is also used by the existing `[auth0.failed_auth]` log line, so this changes that field too.

> **Why it matters:** A rate limiter that a one-line header change defeats provides no protection against the exact scenario it was built for, and the CloudWatch `apikey-rate-limited` alarm will never fire for a competent attacker while the operator believes the control is in place.

Evidence:
- src/api/auth_backends.py:26-28 - `return xff.split(",")[0].strip() if xff else ...`
- src/api/auth_backends.py:322-324 - `record_failure("ip", ip)`; `key_id` scope only recorded `if key_id`

References: CWE-290, CWE-807, CWE-770

</details>

<details>
<summary><strong>CR-010</strong> &middot; Warning &middot; <code>performance</code> — No indexes support the daily maintenance queries or the admin list filters on a table whose rows are never purged</summary>

APIKey rows are permanent by design: revoke() and the expiry sweep in api_key_maintenance.py never delete a row, only flip flags, so the table only grows over the life of the product. The two daily maintenance queries filter on (is_active, expires_at) and (is_active, expires_at, revoked_at, last_used_at/created_on), and neither combination is backed by any index; key_id's unique index is the only index besides the automatic FK indexes on profile/created_by/updated_by. The same missing coverage also affects the admin surface: `APIKeyExpiryFilter` and the `revoked` `EmptyFieldListFilter` in the Django admin list view filter on the same `is_active`, `expires_at`, `revoked_at`, and `last_used_at` columns. Today's row count is small enough that this is invisible, but nothing in the schema keeps it that way, and a daily cron doing two sequential scans over an ever-growing, never-shrinking table - plus an admin list view exercising the same columns - is exactly the pattern that degrades silently as the key count grows.

> **Suggestion:** Add partial indexes tailored to the two maintenance queries, e.g. `Index(fields=['expires_at'], condition=Q(is_active=True, expires_at__isnull=False), name='apikey_active_expiring_idx')` for `_deactivate_expired`, and a second partial index covering `is_active=True, expires_at__isnull=True, revoked_at__isnull=True` for `_report_stale`. These can be added in a follow-up migration; a CONCURRENTLY-created partial index is cheap even on a table that has already grown. Alternatively, a plain composite index on `(is_active, expires_at)` via `Meta.indexes` would cover both the admin filters and the maintenance command in a single, simpler migration if partial-index syntax is not desired.

> **Trade-offs:** Partial indexes add a small amount of write overhead on every APIKey insert/update and need a separate migration (with CREATE INDEX CONCURRENTLY if applied to a populated production table) rather than being folded into 0134.

> **Why it matters:** Without these indexes, the daily maintenance command's cost scales linearly with total historical key count (active, expired, and revoked), not with the number of keys that actually need attention, since the table keeps every key ever issued. The same full-scan cost applies to the admin list view's filters as the table grows. Not urgent while the table is small, but would need a follow-up migration once the key count is significant.

Applies to: `src/api/models/base.py`, `src/tools/management/commands/api_key_maintenance.py`

<sub>Found by: spk-reviewer-database, spk-reviewer-backend-python · CR-010</sub>

</details>

<details>
<summary><strong>CR-011</strong> &middot; Warning &middot; <code>test_gap</code> — No test exercises the concurrent-creation race branch of FailureRateLimiter.record_failure</summary>

`record_failure` in src/api/utils/ratelimit.py has an explicit fallback path for the case where two workers race between `cache.incr()` raising `ValueError` and `cache.add()`: if `cache.add()` returns False (another worker created the counter first), it falls back to a second `cache.incr()`. None of the tests in test_api_key_ratelimit.py exercise this branch; `test_limiter_blocks_at_the_limit` and friends only hit the single-writer path (incr succeeds, or incr fails then add succeeds). The code's own docstring calls this race out as an intentional design decision, but the path that handles it is untested.

> **Suggestion:** Add a test that mocks `api.utils.ratelimit.cache.add` to return False on the first call (simulating a second worker winning the race) and asserts `record_failure` still returns the correct incremented count via the fallback `cache.incr()`, e.g. a dedicated `test_limiter_handles_concurrent_first_failure`.

> **Why it matters:** This is the one branch in the failure counter explicitly designed for concurrent access; if a future change breaks the fallback (swallows the second ValueError incorrectly, or loses the count), nothing in the suite would catch it, and the limiter could silently undercount failures under real production concurrency.

Applies to: `src/api/tests/test_api_key_ratelimit.py`, `src/api/utils/ratelimit.py`

<sub>Found by: spk-reviewer-tests · CR-011</sub>

</details>

<details>
<summary><strong>CR-007</strong> &middot; Warning &middot; <code>security</code> — API keys are silently ignored (not rejected) on every endpoint using `method_authentication_classes = {"GET": []}`</summary>

`APIKeyAuthentication`'s docstring and `AnonymousJWTAuthentication` were both written so that a key that is present but bad always fails closed with a 401, and `AnonymousJWTAuthentication.authenticate` was patched specifically to validate keys on public endpoints. That guarantee does not hold for the viewsets that clear the authenticator list per method via `MethodAuthenticationMixin`: `BaseAttributeApiViewSet` (src/api/resources/base.py:622), `BaseChoiceApiViewSet` (:705), `fish_size.py:22`, `project_tag.py:22`, `classification/classifier.py:35`, `classification/label_mapping.py:49`. On a GET, `request.authenticators` is emptied, so a revoked, expired or forged key is neither validated nor logged nor rate-limited; the request just becomes anonymous and returns 200. Two effects: a client whose key was revoked keeps getting successful responses from these endpoints and never learns the credential is dead (the `[apikey.failed_auth]` metric and its CloudWatch alarm see nothing), and a key-authenticated caller silently loses access to their own `PROPOSED` attributes because `BaseAttributeApiViewSet.get_queryset` sees `AnonymousUser`.

> **Suggestion:** Replace `{"GET": []}` with `{"GET": [AnonymousJWTAuthentication]}` on those viewsets (it already returns `AnonymousUser` when no credential is present, and now validates API keys), or add `APIKeyAuthentication` to those lists so a presented key is always adjudicated. Add a test asserting a revoked key gets 401 from one of these public GET endpoints.

> **Trade-offs:** Switching from `[]` to `AnonymousJWTAuthentication` means a malformed JWT on these endpoints, which is currently ignored, would start returning 401. That is the intended fail-closed behaviour but it is a behaviour change for existing clients sending stale tokens to public reference-data endpoints, so it is worth checking the mobile/collect clients before shipping.

> **Why it matters:** The stated invariant that a present-but-bad key always 401s is what makes revocation observable to clients and to the new CloudWatch alarms. A set of endpoints where revocation is invisible undermines both.

Evidence:
- src/api/resources/mixins.py:196-199 - `request.authenticators = [auth() for auth in self.method_authentication_classes[method]]`
- src/api/resources/base.py:622 - `method_authentication_classes = {"GET": []}`
- src/api/auth_backends.py:225-227 - "A key that is present but bad always fails closed with a 401."

References: CWE-613, CWE-306

<sub>Found by: spk-reviewer-security · CR-007</sub>

</details>

<details>
<summary><strong>CR-018</strong> &middot; Info &middot; <code>maintainability</code> — Key-rotation lineage ('replaces') is not persisted in the schema, only in log lines</summary>

`APIKey.issue()` and `log_key_created()` accept a `replaces` argument used purely to write `replaces=<key_id>` into the `[apikey.created]` audit log line; there is no `replaces`/`previous_key` FK column on the APIKey model or in the 0134 migration. The only way to reconstruct which key rotated which is by grepping CloudWatch logs; the database has no queryable relationship between a revoked key and the key that replaced it.

> **Suggestion:** If key-rotation history is expected to be inspectable via the ORM/admin rather than only logs, add a nullable self-referential FK (e.g. `replaces = models.ForeignKey('self', null=True, blank=True, on_delete=models.SET_NULL, related_name='replaced_by')`) and set it in `issue()`.

> **Trade-offs:** Requires a new migration adding the column/FK; nullable and self-referential to the same currently-empty table, so a low-risk additive change with no backfill needed.

> **Why it matters:** Without a persisted link, auditing or building any report around key rotation depends entirely on log retention and log parsing rather than a query against the APIKey table.

Applies to: `src/api/models/base.py`, `src/api/utils/apikeys.py`

<sub>Found by: spk-reviewer-database · CR-018</sub>

</details>

<details>
<summary><strong>CR-020</strong> &middot; Info &middot; <code>maintainability</code> — The APIKey-minting helper is redefined independently in seven test files instead of living in fixtures/apikeys.py</summary>

src/api/tests/fixtures/apikeys.py only exposes `api_key_audit_logs` and `api_key_audit_lines`. The near-identical `_make_key`/`key_pair` helper (create an APIKey via `generate_api_key()` + `APIKey.objects.create(...)`) is copy-pasted with small variations across test_api_key_auth.py, test_api_key_ratelimit.py, test_api_key_public.py, test_api_key_permissions.py, test_api_key_resource.py, test_api_key_lifecycle.py, and test_api_key_admin.py.

> **Suggestion:** Move a single `make_api_key(profile, **kwargs)` helper (and/or a `key_pair` fixture) into fixtures/apikeys.py so all seven modules share one implementation of how a key is minted for tests.

> **Why it matters:** Seven independent copies of the same construction logic are seven places that can drift; one already differs in default expiry (30 vs 365 days), and a change to APIKey's required fields or defaults means finding and updating every copy rather than one shared fixture.

Applies to: `src/api/tests/fixtures/apikeys.py`, `src/api/tests/test_api_key_auth.py`, `src/api/tests/test_api_key_ratelimit.py`, `src/api/tests/test_api_key_public.py`, `src/api/tests/test_api_key_permissions.py`, `src/api/tests/test_api_key_resource.py`, `src/api/tests/test_api_key_lifecycle.py`, `src/api/tests/test_api_key_admin.py`

<sub>Found by: spk-reviewer-tests · CR-020</sub>

</details>

<details>
<summary><strong>CR-021</strong> &middot; Info &middot; <code>readability</code> — Docstrings and comments reference unresolvable internal ticket codes (C4/C5/C6/C8) and "the API key plan"</summary>

Ruff/isort compliance, type annotations, and dead-code checks were clean across the API key changes; the only recurring code-quality issue is that several docstrings and comments cite an internal planning document and ticket codes that are not resolvable by anyone outside the original design discussion: src/api/utils/apikeys.py:14 references 'the "Hash, do not encrypt" decision in the API key plan'; src/api/utils/apikeys.py:32 references ticket code '(C8)'; src/api/utils/ratelimit.py:1 references 'C8 of the API key plan' in its module docstring; src/api/admin/base.py:247 references 'C5' ('Timed rotation with an automatic tail is C5'); src/tools/management/commands/api_key_maintenance.py:1 begins its docstring with 'C4:'; src/api/tests/fixtures/apikeys.py:10 references '(C8)'; and src/api/tests/test_api_key_admin.py:1 begins its module docstring with 'C6:'. None of these codes are resolvable from the codebase itself, so a future reader without access to the original planning document cannot verify the reasoning behind these decisions.

> **Suggestion:** Replace each ticket/plan reference with an inline explanation of the reasoning (as already done well elsewhere in this PR's docstrings), or, if an external tracker item exists, cite it by a resolvable link or issue number rather than a bare internal code.

> **Why it matters:** Comments and docstrings are the primary way future maintainers understand why a security-sensitive design choice (hashing over encryption, rate-limiting strategy, rotation behavior) was made; a reference to an internal document that will not survive outside this PR's original context degrades to noise once that document is inaccessible.

Evidence:
- src/api/utils/apikeys.py:14 - 'the "Hash, do not encrypt" decision in the API key plan'
- src/api/utils/apikeys.py:32 - '(C8)'
- src/api/utils/ratelimit.py:1 - 'C8 of the API key plan'
- src/api/admin/base.py:247 - 'Timed rotation with an automatic tail is C5'
- src/tools/management/commands/api_key_maintenance.py:1 - docstring begins with 'C4:'
- src/api/tests/fixtures/apikeys.py:10 - '(C8)'
- src/api/tests/test_api_key_admin.py:1 - module docstring begins with 'C6:'

Applies to: `src/api/utils/apikeys.py`, `src/api/utils/ratelimit.py`, `src/api/admin/base.py`, `src/tools/management/commands/api_key_maintenance.py`, `src/api/tests/fixtures/apikeys.py`, `src/api/tests/test_api_key_admin.py`

<sub>Found by: spk-reviewer-python-quality · CR-021</sub>

</details>


### Agent Coverage

| spk-reviewer-security | security | 31 | 10 | 0 |
| spk-reviewer-backend-python | backend | 20 | 3 | 0 |
| spk-reviewer-database | database | 5 | 4 | 0 |
| spk-reviewer-tests | testing | 11 | 5 | 0 |
| spk-reviewer-devops | infrastructure | 1 | 1 | 0 |
| spk-reviewer-docs | documentation | 2 | 0 | 0 |
| spk-reviewer-python-quality | code_quality | 20 | 7 | 0 |

src/api/resources/apikey.py
```diff
@@ -0,0 +1,172 @@
...
+    def get_queryset(self):
+        profile = getattr(self.request.user, "profile", None)
+        if profile is None:
+            return APIKey.objects.none()
+        return APIKey.objects.filter(profile=profile).order_by("-created_on")
+
+    def get_serializer_class(self):
+        if self.action == "create":
+            return APIKeyCreateSerializer
+        return super().get_serializer_class()
```
gridcell, line 167: **Warning** · `security` · confidence: high

**`?fields=secret_hash` exposes the stored key digest through the generic field-limiting feature**

`APIKeyViewSet.get_serializer_class()` delegates to `BaseApiViewSet.get_serializer_class()`, which honours the `fields` query parameter by building a subclass of `APIKeySerializer` whose `Meta.fields` is whatever the caller asked for. The validation in `get_serializer_class_for_fields` (src/api/resources/base.py:573-578) accepts any name in `{f.name for f in model._meta.get_fields()}`, and `secret_hash` is a concrete model field. Because the generated class subclasses `APIKeySerializer`, DRF's "declared field must be in fields" assertion is skipped, so `GET /v1/apikeys/?fields=key_id,secret_hash` renders the digest. The `exclude` on the admin form and the deliberate omission from `APIKeySerializer.Meta.fields` are both bypassed, contradicting the docstring claim that a digest never leaves the database. Exploitability is limited (a caller only sees hashes of their own keys, and the secret has 256 bits of entropy), but this pushes credential-verification material into client code, client logs and any intermediary that records response bodies.

> **Suggestion:** Deny the field explicitly on this viewset, e.g. override `get_serializer_class_for_fields` (or add a `sensitive_fields` hook on `BaseApiViewSet`) so `secret_hash` raises a `ValidationError`:
> 
> ```python
> SENSITIVE_FIELDS = {"secret_hash"}
> 
> def get_serializer_class_for_fields(self, serializer_class, fields):
>     requested = {f.strip() for f in fields.split(",")}
>     if requested & self.SENSITIVE_FIELDS:
>         raise ValidationError({"fields": "Invalid field names: secret_hash"})
>     return super().get_serializer_class_for_fields(serializer_class, fields)
> ```
> 
> Add a regression test asserting `GET /apikeys/?fields=key_id,secret_hash` does not return the digest.

> **Trade-offs:** A generic deny-list on `BaseApiViewSet` would apply to every viewset, so confirm no existing client relies on requesting a field that lands on the list.

> **Why it matters:** The digest is credential material. The code and the new documentation both state it can never leave the database; a one-parameter bypass of that control on a brand-new credential endpoint is the kind of gap that gets copied into client integrations and logs.

Evidence:
- src/api/resources/apikey.py:141-144 - `get_serializer_class` falls through to `super().get_serializer_class()`
- src/api/resources/base.py:573-578 - `valid_fields |= {f.name for f in model._meta.get_fields()}`
- src/api/resources/apikey.py:33 - "`secret_hash` is left out on purpose"

References: CWE-200, CWE-522, OWASP API3:2023 Broken Object Property Level Authorization

<sub>Found by: spk-reviewer-security · CR-001</sub>

src/api/auth_backends.py
```diff
@@ -175,13 +189,172 @@ def _validate_app(self, client_id):
...
+        A busy client would otherwise write to the row on every request.
+        """
+
+        cache_key = f"apikey:last_used:{api_key.key_id}"
+        if cache.get(cache_key):
+            return
+        cache.set(cache_key, True, self.last_used_throttle)
+        now = timezone.now()
+        # update() to avoid touching updated_on/updated_by on every call
+        APIKey.objects.filter(pk=api_key.pk).update(last_used_at=now, last_used_ip=ip)
```
gridcell: **Warning** · `security` · confidence: high

**Unvalidated X-Forwarded-For value written to the `inet` column in `_touch`, causing a 500 and poisoning the audit field**

`_touch` writes `_get_client_ip(request)` straight into `APIKey.last_used_ip`, a `GenericIPAddressField` (Postgres `inet`). `QuerySet.update()` performs no model validation, and `GenericIPAddressField.get_prep_value` only normalises IPv6; any other string is passed through verbatim. Two consequences: (1) a request carrying a valid key and `X-Forwarded-For: notanip` raises `django.db.utils.DataError: invalid input syntax for type inet` from inside `authenticate()`, which DRF does not handle, producing a 500 on an otherwise valid request - the same happens when both XFF and `REMOTE_ADDR` are absent, because `_get_client_ip` falls back to the literal string `"unknown"`; (2) where the value is a syntactically valid but forged address, `last_used_ip`, the field an operator would use to answer "where is this leaked key being used from", is attacker-controlled and silently wrong.

> **Suggestion:** Validate before persisting and store `None` when the value is not an address, while keeping the raw value only in the log line:
> 
> ```python
> import ipaddress
> 
> def _client_ip_for_db(ip):
>     try:
>         return str(ipaddress.ip_address(ip))
>     except ValueError:
>         return None
> ```
> 
> and use it in `_touch`. Separately, drop the `"unknown"` sentinel from `_get_client_ip` (return `""`/`None`) so it can never reach a typed column.

> **Why it matters:** An attacker-supplied header turns every request from a valid key into a 500, which is a cheap denial of service against any client whose proxy chain emits a non-address value, and it makes the one forensic field on the credential untrustworthy.

Evidence:
- src/api/auth_backends.py:341 - `APIKey.objects.filter(pk=api_key.pk).update(last_used_at=now, last_used_ip=ip)`
- src/api/auth_backends.py:28 - fallback value is the string `"unknown"`
- src/api/models/base.py:249 - `last_used_ip = models.GenericIPAddressField(null=True, blank=True)`

References: CWE-20, CWE-117, CWE-248

<sub>Found by: spk-reviewer-security · CR-003</sub>

src/api/admin/base.py
```diff
@@ -1,396 +1,602 @@
...
+    # override here. Issuing is not a superuser-only action: /apikeys/ lets any
+    # signed-in person mint keys for themselves, and this page is the same act
+    # for someone else's profile, gated by the permission rather than by a flag.
+
+    def has_delete_permission(self, request, obj=None):
+        # Deletion is the one narrower privilege, and it is not about issuing.
+        # Revoking retires a key and keeps the row, which is what answers "what
+        # did this credential do, and when did it stop working". Delete throws
+        # that away, so it stays with the superuser.
+        return request.user.is_superuser and super().has_delete_permission(request, obj)
```
gridcell, line 229: **Warning** · `security` · confidence: high

**`api.add_apikey` in the Django admin is effectively unrestricted profile impersonation**

`APIKeyAdmin` deliberately relies on the plain `api.add_apikey` model permission with no `has_add_permission` override, and `profile` is an editable autocomplete field on the add form (it is only made read-only for an existing object, in `get_readonly_fields`). Any staff user granted `api.add_apikey` can therefore mint a key against an arbitrary profile, including a profile that is a `ProjectProfile` ADMIN on projects the staff user has no membership in, and immediately use `Authorization: Bearer mmd_...` to read and write that project's data as that person. `save_model` shows them the raw secret. The comment at lines 217-221 justifies this by analogy to `/apikeys/`, but `/apikeys/` is strictly self-service (`APIKeyViewSet.get_queryset` filters to `request.user.profile` and `create` uses `request.user.profile`), whereas this page issues credentials for anyone. That makes a permission whose name reads like "can add a row" equivalent to "can impersonate any MERMAID user". This is a distinct finding from a previously addressed thread about restricting edits to an existing key's `expires_at`/`never_expires`; this one concerns the add path and the editable `profile` field.

> **Suggestion:** Either gate issuing on somebody else's behalf to superusers, or restrict the selectable profile. The narrowest version keeps self-service working for ordinary staff:
> 
> ```python
> def formfield_for_foreignkey(self, db_field, request, **kwargs):
>     if db_field.name == "profile" and not request.user.is_superuser:
>         kwargs["queryset"] = Profile.objects.filter(email=request.user.email)
>     return super().formfield_for_foreignkey(db_field, request, **kwargs)
> ```
> 
> Alternatively add `has_add_permission` returning `request.user.is_superuser and super().has_add_permission(request)`, matching the reasoning already applied to `has_delete_permission`. Whichever is chosen, state the consequence of granting `api.add_apikey` in the admin help text so it is not handed out casually.

> **Trade-offs:** Superuser-only issuance means support staff can no longer provision a key for a user who cannot complete the self-service flow; if that workflow matters, prefer the queryset restriction plus an explicit note, or add a separate custom permission (e.g. `issue_apikey_for_others`) so the capability is grantable but named for what it does.

> **Why it matters:** A leaked or over-granted `api.add_apikey` permission yields a long-lived credential for any user in the system, with no project membership required and no notification to the impersonated user. That is a full authorization bypass for project data, and the resulting key looks legitimate in every audit trail except the `[apikey.created] actor=` line.

Evidence:
- src/api/admin/base.py:145 - form fields include `profile`
- src/api/admin/base.py:208-215 - `profile` is only made read-only when `obj is not None`
- src/api/admin/base.py:217-221 - "Issuing is not a superuser-only action"
- src/api/admin/base.py:291-301 - `save_model` generates the key and reveals the raw secret

References: CWE-269, CWE-863, OWASP API1:2023 Broken Object Level Authorization

src/api/auth_backends.py
```diff
@@ -175,13 +189,172 @@ def _validate_app(self, client_id):
...
+        self._check_rate_limit(request, "ip", ip)
+
+        try:
+            env, key_id, secret = parse_api_key(raw)
+        except ValueError:
+            raise self._fail(request, "malformed_key", None, ip=ip)
+
+        # Counted separately from the IP: one bad key rotating through hosts is
+        # the same incident, and one bad host should not lock out a good key.
+        self._check_rate_limit(request, "key_id", key_id)
```
gridcell: **Warning** · `security` · confidence: high

**A known key_id can be locked out by a third party: the key_id failure limit is checked before the secret is verified**

`authenticate()` calls `_check_rate_limit(request, "key_id", key_id)` immediately after parsing and before `secret_matches`, and `_fail` records a failure in the `key_id` scope for every rejection. An attacker who learns a `key_id`, which the code treats as non-secret and which appears in log lines, the admin list, `search_fields`, the CSV export and the `/apikeys/` API response, can send 10 requests per minute with `mmd_<env>_<that_key_id>_garbage` and the legitimate holder of that key is answered `429` for the remainder of every minute, indefinitely, at a cost of 10 requests/minute to the attacker. Combined with the X-Forwarded-For issue, the attacker cannot be throttled out of doing this by the IP scope.

> **Suggestion:** Separate "is this key under attack" from "is this caller under attack". Options, roughly in order of preference: (1) only count a `key_id`-scope failure when the presented secret was correct but the key was unusable (revoked/expired/inactive), which is the real signal of a stale deployed credential, and rely on the fixed IP scope for wrong-secret attempts; (2) key the counter on `(key_id, ip)` rather than `key_id` alone, so one bad host cannot lock out a good key - the docstring at lines 258-259 states this as the goal, but the current scope does not achieve it; (3) keep the `key_id` counter for alerting only (log/metric) and do not raise `Throttled` on it.

> **Trade-offs:** Option 1 narrows the blast radius but weakens the guard against an attacker guessing secrets for a known key_id; given a 256-bit secret that trade is safe. Option 2 increases the number of cache rows in DatabaseCache proportional to distinct (key_id, ip) pairs.

> **Why it matters:** The rate limiter currently converts a deliberately non-secret identifier into a denial-of-service handle against a specific production integration, and the pipeline it protects (sync push, ingest, reporting) is exactly where downtime is expensive.

Evidence:
- src/api/auth_backends.py:258-260 - key_id limit checked before `secret_matches`
- src/api/auth_backends.py:323-324 - `if key_id: self.failure_limiter.record_failure("key_id", key_id)`
- src/api/utils/apikeys.py:10 - "`key_id` identifies the `APIKey` row and is safe to log"

References: CWE-770, CWE-645

src/api/resources/apikey.py
```diff
@@ -0,0 +1,172 @@
...
+        never_expires = attrs.get("never_expires", False)
+        if never_expires and expires_at is not None:
+            raise serializers.ValidationError(
+                {"expires_at": "Set expires_at or never_expires, not both."}
+            )
+        if never_expires:
+            attrs["expires_at"] = None
+        elif expires_at is None:
+            attrs["expires_at"] = default_expires_at()
+        return attrs
```
gridcell, line 134: **Warning** · `security` · confidence: high

**No maximum lifetime or per-profile cap on self-service key creation, so the `never_expires` control is decorative**

`APIKeyCreateSerializer.validate_expires_at` only requires `expires_at > now`. A caller can POST `{"name": "x", "expires_at": "9999-12-31T00:00:00Z"}` and get a key that outlives any plausible deployment, so the design intent stated in the docstring, that a key which never expires has to be asked for with `never_expires` so no-expiry is never the silent default, is bypassed without ever setting `never_expires`. There is also no cap on how many keys one profile may hold, and no DELETE (by design), so a script in a loop can create unbounded permanent rows, each of which is a live credential for that profile's full access, and each of which the maintenance sweep will never report because it only inspects `expires_at__isnull=True`.

> **Suggestion:** Enforce a ceiling in `validate_expires_at` and cap the number of live keys per profile in `create`:
> 
> ```python
> MAX_LIFETIME_DAYS = 365 * 2
> 
> def validate_expires_at(self, value):
>     if value is not None:
>         if value <= timezone.now():
>             raise serializers.ValidationError("expires_at must be in the future.")
>         if value > timezone.now() + timedelta(days=MAX_LIFETIME_DAYS):
>             raise serializers.ValidationError(
>                 f"expires_at cannot be more than {MAX_LIFETIME_DAYS} days from now."
>             )
>     return value
> ```
> 
> and in `APIKeyViewSet.create`, reject with 400 when the profile already has N usable keys. Also widen `_report_stale` in `api_key_maintenance.py` to cover far-future expiries, not just `expires_at__isnull=True`.

> **Trade-offs:** A hard cap on key count can break a legitimate user who runs several independent integrations; pick the limit generously (e.g. 20) and make it a setting rather than a constant so it can be raised without a deploy.

> **Why it matters:** Expiry is one of the three stated compensating controls for the fact that a key carries the whole of its profile's access. If the ceiling is unbounded and the count is unbounded, the only remaining control is manual revocation of credentials nobody is tracking.

Evidence:
- src/api/resources/apikey.py:95-98 - only `value <= timezone.now()` is rejected
- src/api/resources/apikey.py:86-88 - "so that no-expiry is never the silent default"
- src/tools/management/commands/api_key_maintenance.py:80 - stale report filters `expires_at__isnull=True` only

References: CWE-613, CWE-770

<sub>Found by: spk-reviewer-security · CR-006</sub>

src/api/admin/base.py
```diff
@@ -1,396 +1,602 @@
...
+            request,
+            format_html(
+                "API key <strong>{}</strong> issued for {}. Copy it now: it is not "
+                "stored and cannot be shown again.<br><code>{}</code>",
+                key.name,
+                key.profile,
+                raw,
+            ),
+            messages.SUCCESS,
+        )
```
gridcell: **Warning** · `security` · confidence: medium

**Raw API key delivered through the messages framework lands in a client-side cookie that is not marked Secure**

`_show_raw_key` puts the raw credential into `message_user`. With the default `MESSAGE_STORAGE` (`FallbackStorage`), that serialises the message into the `messages` cookie. Django sets that cookie `httponly=True` but with `secure=settings.SESSION_COOKIE_SECURE`, and this project never sets `SESSION_COOKIE_SECURE` (settings.py configures only `SECURE_PROXY_SSL_HEADER`), so it defaults to `False`. The raw key is therefore written to the browser's cookie store in a non-Secure cookie and will be re-sent on the next request to the admin, including over plain HTTP if any path to the origin permits it. The comment claims the exposure is the same as rendering it in the page, but rendering it in a response body does not persist it to disk or replay it on a subsequent request. `generate_replacement_keys` compounds this by queuing one such message per selected row, which also risks silently spilling to the session backend past the 4096-byte cookie limit.

> **Suggestion:** Render the raw key in a one-shot template response instead of the messages framework, e.g. have `save_model` stash it on `self` and override `response_add`/`response_post_save_add` to render an intermediate "copy your key" page. If the messages framework is kept, at minimum set `SESSION_COOKIE_SECURE = True` and `CSRF_COOKIE_SECURE = True` for the dev/prod branches of settings.py so the cookie cannot traverse plain HTTP.

> **Trade-offs:** Setting `SESSION_COOKIE_SECURE = True` will break local development over HTTP unless it is scoped to `ENVIRONMENT in ("dev", "prod")`; scope it the same way `DEBUG` already is.

> **Why it matters:** The whole design rests on the raw secret existing for exactly one response. A non-Secure cookie extends its lifetime to the browser's cookie store and its exposure to at least one further request, which is the kind of leak that later shows up in a proxy log or a shared workstation.

Evidence:
- src/api/admin/base.py:274-289 - `_show_raw_key` via `self.message_user(...)`
- src/api/admin/base.py:277-278 - "cookie, falling back to the session"
- src/app/settings.py - no `SESSION_COOKIE_SECURE` / `CSRF_COOKIE_SECURE` set

References: CWE-315, CWE-614, CWE-522

<sub>Found by: spk-reviewer-security · CR-008</sub>

src/api/admin/base.py
```diff
@@ -1,396 +1,602 @@
...
+        "linked_email",
+        "project_count",
+        "has_collect_state",
+        "has_explore_state",
+    )
+    search_fields = ["first_name", "last_name", "email"]
+
+    @admin.display(description="Email", ordering="email")
+    def linked_email(self, obj):
+        return format_html(f'<a href="mailto:{obj.email}">{obj.email}</a>')
```
gridcell: **Warning** · `security` · confidence: medium

**`format_html` with a pre-interpolated f-string in `ProfileAdmin.linked_email` bypasses escaping**

`format_html(f'<a href="mailto:{obj.email}">{obj.email}</a>')` interpolates before `format_html` runs, so the result is marked safe without any of its arguments being escaped. `Profile.email` is written from Auth0 `user_info` in `JWTAuthentication._validate_profile` via `profile.save()` (no `full_clean`), and is editable in the admin, so a value containing markup is stored verbatim and rendered as HTML in the profile changelist and in every autocomplete result. This line is unchanged behaviour, but the whole file was re-added in this diff (a CRLF to LF renormalisation), so it is inside the reviewed hunks, and the new `APIKeyAdmin` adds `profile__email` to `search_fields` plus `autocomplete_fields = ("profile",)`, widening where the value is rendered.

> **Suggestion:** Pass the values as arguments so they are escaped:
> 
> ```python
> return format_html('<a href="mailto:{}">{}</a>', obj.email, obj.email)
> ```

> **Trade-offs:** None expected; `format_html` with positional arguments is a drop-in replacement that preserves the rendered link markup while escaping the interpolated values.

> **Why it matters:** Stored XSS in the Django admin runs with a staff session, which on this deployment includes users holding `api.add_apikey`, so an injected script can mint an API key for any profile without further interaction.

Evidence:
- src/api/admin/base.py:327-328 - `return format_html(f'<a href="mailto:{obj.email}">{obj.email}</a>')`
- src/api/auth_backends.py:131-135 - profile email populated from Auth0 `user_info` without validation
- src/api/admin/base.py:182-189 - `profile__email` search field and `autocomplete_fields`

References: CWE-79, OWASP A03:2021 Injection

<sub>Found by: spk-reviewer-security · CR-009</sub>

src/api/tests/test_api_key_admin.py
```diff
@@ -0,0 +1,397 @@
...
+    assert "secret_hash" not in APIKeyAdminForm(instance=key).fields
+    assert "secret_hash" not in key_admin.get_readonly_fields(request, obj=key)
+    assert key.secret_hash not in str(key)
+
+
+def test_admin_offers_no_all_fields_export(key_admin):
+    # BaseAdmin's all-fields export walks every concrete field, secret_hash
+    # included, so this page must not inherit it.
+    assert isinstance(key_admin, BaseAdmin) is False
+    assert "export_model_all_as_csv" not in key_admin.get_actions(_request())
```
gridcell, line 389: **Warning** · `test_gap` · confidence: medium

**Admin hash-leak checks assert on ModelAdmin config, not on the actually rendered CSV export output**

`test_admin_never_exposes_the_hash` and `test_admin_offers_no_all_fields_export` verify that `secret_hash` is absent from `list_display`, `exportable_fields`, `get_fields()`, and the form's fields, and that `export_model_all_as_csv` is not offered as an action. None of this invokes `export_model_display_as_csv(key_admin, request, queryset)` (the real CSV-writing path in src/api/admin/base.py) and inspects the response body for the absence of `secret_hash`. Since these are the tests standing between a leaked hash and a shipped CSV export, asserting on static configuration is a weaker guarantee than asserting on generated output: a future refactor that changes how `export_model_as_csv` resolves fields (e.g. following a relation) could reintroduce a leak without failing any test here. This is distinct from a previously resolved thread that asked only for an assertion on the action name returned by `get_actions`; this finding is about the export output itself never being executed and inspected.

> **Suggestion:** Add a test that calls `export_model_display_as_csv(key_admin, request, APIKey.objects.filter(pk=key.pk))`, reads the CSV bytes off the returned `HttpResponse`, and asserts `key.secret_hash` does not appear anywhere in the content, mirroring the approach already used for the sync-pull leak check in test_sync_pull_project_profiles_has_no_api_keys.

> **Why it matters:** secret_hash is a SHA-256 digest of the credential; while not directly usable, it is exactly the artifact an admin-export leak is designed to prevent, and the current tests only check that it is not configured to leak rather than that it does not.

<sub>Found by: spk-reviewer-tests · CR-012</sub>

src/api/tests/test_api_key_auth.py
```diff
@@ -0,0 +1,240 @@
...
+@pytest.mark.parametrize(
+    "mangle",
+    [
+        pytest.param(lambda raw: raw + "x", id="wrong_secret"),
+        pytest.param(lambda raw: "mmd_local_aaaaaaaaaaaa_nope", id="unknown_key_id"),
+        pytest.param(lambda raw: "mmd_local_short_secret", id="bad_key_id_length"),
+        pytest.param(lambda raw: "mmd_nonsense", id="prefix_only"),
+        pytest.param(lambda raw: raw.replace("_local_", "_prod_", 1), id="wrong_environment"),
+    ],
+)
```
gridcell, line 99: **Question** · `test_gap` · confidence: low

**wrong_environment mangle assumes settings.ENVIRONMENT is exactly "local" without pinning it**

This parametrize case mangles the raw key by replacing the literal substring `"_local_"` with `"_prod_"`. It relies on `settings.ENVIRONMENT` being `"local"` at test-run time (true today via .env's ENV=local), but nothing in the test pins that assumption. If the suite is ever run with ENV=dev or ENV=prod, `raw` would not contain `"_local_"`, `.replace()` becomes a silent no-op, and the mangled key becomes identical to the valid one; `pytest.raises(AuthenticationFailed)` would then fail for reasons unrelated to the wrong-environment check, producing a confusing failure rather than testing the intended behavior.

> **Suggestion:** Use the `settings` fixture to pin `settings.ENVIRONMENT = "local"` for this test, or derive the substring to replace from `settings.ENVIRONMENT` rather than hardcoding `_local_`, so the environment dependency is explicit rather than implicit in ambient config.

<sub>Found by: spk-reviewer-tests · CR-013</sub>

iac/stacks/constructs/alerts.py
```diff
@@ -363,6 +363,76 @@ def __init__(
...
+                metric=apikey_rate_limit_metric.metric(
+                    statistic="Sum",
+                    period=Duration.minutes(5),
+                ),
+                threshold=1,
+                evaluation_periods=1,
+                comparison_operator=cw.ComparisonOperator.GREATER_THAN_OR_EQUAL_TO_THRESHOLD,
+                treat_missing_data=cw.TreatMissingData.NOT_BREACHING,
+            )
+        )
```
gridcell, line 434: **Question** · `maintainability` · confidence: low

**ApiKeyRateLimitAlarm fires on a single occurrence per 5-minute window**

threshold=1 with evaluation_periods=1 and GREATER_THAN_OR_EQUAL_TO_THRESHOLD means any single 429 (one IP or key_id hitting 10 failures in 60s) pages the shared Slack channel. This mirrors the existing Auth0RateLimitAlarm, but the trigger condition for an API key rate limit (10 bad requests/minute from one IP or key_id) is more plausibly reached by benign client misconfiguration (an app that has not rotated an expired key yet, or a batch script retrying) than the Auth0 Management API quota this pattern was originally designed for. Worth confirming the expected on-call cadence before merge, since a chatty client could produce recurring pages rather than a rare-but-meaningful signal.

<sub>Found by: spk-reviewer-devops · CR-014</sub>

src/api/migrations/0134_apikey.py
```diff
@@ -0,0 +1,69 @@
...
+                ("created_on", models.DateTimeField(auto_now_add=True)),
+                ("updated_on", models.DateTimeField(auto_now=True)),
+                ("name", models.CharField(max_length=100)),
+                ("key_id", models.CharField(db_index=True, max_length=12, unique=True)),
+                ("secret_hash", models.CharField(max_length=64)),
+                ("is_active", models.BooleanField(default=True)),
+                ("expires_at", models.DateTimeField(blank=True, null=True)),
+                ("last_used_at", models.DateTimeField(blank=True, null=True)),
+                ("last_used_ip", models.GenericIPAddressField(blank=True, null=True)),
+                ("revoked_at", models.DateTimeField(blank=True, null=True)),
```
gridcell, line 33: **Info** · `maintainability` · confidence: high

**db_index=True on key_id is redundant alongside unique=True**

`key_id` is declared with both `unique=True` and `db_index=True`. Django's schema editor only creates a separate btree index when `field.db_index and not field.unique` is true (`_field_should_be_indexed` in `django/db/backends/base/schema.py`), so `db_index=True` here is a no-op: the unique constraint's implicit index is the only index actually created. Not a duplicate-index performance problem, just a misleading field declaration.

> **Suggestion:** Drop `db_index=True` from the `key_id` field since `unique=True` already guarantees an index for lookups.

> **Why it matters:** Purely a readability/consistency nit; leaving it in does not create a redundant index in Postgres, but it does suggest to a future reader that an extra index exists when it does not.

Evidence:
- key_id = models.CharField(max_length=12, unique=True, db_index=True)
- migration: ("key_id", models.CharField(db_index=True, max_length=12, unique=True)),

<sub>Found by: spk-reviewer-database · CR-015</sub>

src/api/checks.py
```diff
@@ -0,0 +1,15 @@
...
+
+@register()
+def check_api_key_environment(app_configs, **kwargs):
+    """API keys embed an environment label, so ENVIRONMENT must be one we map."""
+
+    try:
+        get_environment_label()
+    except ImproperlyConfigured as err:
+        return [Error(str(err), id="api.E001")]
+    return []
```
gridcell, line 15: **Info** · `error_handling` · confidence: medium

**The environment check does not run under gunicorn, so a bad ENV surfaces as request-time 500s rather than at startup**

`check_api_key_environment` is registered as a Django system check, and `APIKeyAuthentication.authenticate` relies on it: "ImproperlyConfigured here is a deployment error... api.checks catches it at startup" (src/api/auth_backends.py:262-263). System checks run for `manage.py` commands (including `runserver` and `migrate`), but not when the app is served through `app.wsgi`/gunicorn, which is the production path (`make runserverplus`). With an unrecognised `ENV`, plausible given `ENVIRONMENT = os.environ.get("ENV") or "local"` is free text and only local/dev/prod are accepted, the container would boot healthy and then 500 on every API-key request and every key creation. Behaviour is fail-closed, which is right; the gap is only in when it is detected.

> **Suggestion:** Run the same validation eagerly at WSGI/ASGI application startup (e.g. call `check_api_key_environment` from `app/wsgi.py` before creating the application object) so a bad ENV fails container health checks instead of surfacing as request-time 500s.

> **Why it matters:** A misconfigured ENV variable would otherwise pass a healthy startup and container health check, only to fail every API-key request once traffic arrives, delaying detection until customer impact rather than at deploy time.

Evidence:
- src/api/checks.py:7-15 - `@register()` system check
- src/api/auth_backends.py:262-264 - comment relies on startup detection
- src/app/settings.py:10 - `ENVIRONMENT = os.environ.get("ENV") or "local"`

References: CWE-1188

<sub>Found by: spk-reviewer-security · CR-016</sub>

src/app/settings.py
```diff
@@ -385,6 +386,13 @@ def _sentry_before_send(event, hint):
             "filters": ["require_debug_false"],
             "class": "django.utils.log.AdminEmailHandler",
         },
+        # Fixed at INFO, unlike "console": the API key audit trail has to reach
+        # CloudWatch in production, where DEBUG_LEVEL is WARNING.
+        "console_info": {
+            "level": "INFO",
+            "class": "logging.StreamHandler",
+            "stream": sys.stdout,
+        },
```
gridcell, line 413: **Info** · `maintainability` · confidence: medium

**console_info handler for the api.apikeys audit logger has no formatter**

The new `console_info` handler (used exclusively by the `api.apikeys` logger for [apikey.created]/[apikey.revoked]/[apikey.expired]/[apikey.stale] audit lines) does not reference the `file` formatter that other structured log lines use, so it falls back to logging's bare `%(message)s` default. Every other configured handler in this LOGGING dict either uses `file` or Django's default formatter with more context.

> **Suggestion:** Attach the existing `file` formatter (or an equivalent) to `console_info` so audit lines carry a timestamp/level/logger name consistent with the rest of the log stream, which also helps correlate them in CloudWatch Insights.

> **Why it matters:** Not a functional bug (the CloudWatch MetricFilters match on the literal `[apikey.*]` tag substring regardless of formatting), but inconsistent formatting makes the audit trail harder to correlate manually against other log lines during an incident.

<sub>Found by: spk-reviewer-backend-python · CR-017</sub>

src/api/tests/test_api_key_auth.py
```diff
@@ -0,0 +1,240 @@
...
+        APIKeyAuthentication().authenticate(_request(path=f"/v1/projects/?access_token={raw}"))
+        is None
+    )
+
+
+def test_secret_comparison_is_constant_time(key_pair):
+    _, raw = key_pair
+    with patch("api.utils.apikeys.hmac.compare_digest", return_value=True) as mock_compare:
+        _authenticate(raw)
+    mock_compare.assert_called_once()
```
gridcell: **Info** · `test_gap` · confidence: medium

**test_secret_comparison_is_constant_time does not test constant-time behavior**

The test patches `hmac.compare_digest` to always return `True` and then authenticates with the correct raw key, where the real comparison would also return True. All this verifies is that `secret_matches` calls `hmac.compare_digest` exactly once; it says nothing about timing safety, and it does not exercise the mismatch path (a wrong secret also going through `compare_digest`).

> **Suggestion:** Rename the test to reflect what it checks (e.g. `test_secret_comparison_uses_hmac_compare_digest`), or extend it to also assert `compare_digest` is invoked on the wrong-secret path so the design intent is pinned on both branches.

<sub>Found by: spk-reviewer-tests · CR-019</sub>

src/api/auth_backends.py
```diff
@@ -175,13 +189,172 @@ def _validate_app(self, client_id):
...
+                "wrong_environment",
+                key_id,
+                detail="key issued for a different environment",
+                ip=ip,
+            )
+
+        try:
+            api_key = APIKey.objects.select_related("profile").get(key_id=key_id)
+        except APIKey.DoesNotExist:
+            raise self._fail(request, "unknown_key_id", key_id, ip=ip)
```
gridcell, line 310: **Info** · `security` · confidence: low

**key_id rate-limit scope is keyed on an unverified, attacker-chosen value**

In APIKeyAuthentication.authenticate, `key_id` is extracted from the raw bearer token via parse_api_key and used as the cache-key identifier for the 'key_id' rate-limit scope before any database lookup confirms the key exists. Since key_id only has to match a 12-char alphanumeric regex, a caller can mint an unbounded number of distinct, never-issued key_id values (from different IPs, so the per-IP limiter does not catch it) and each gets its own FailureRateLimiter cache row for the current window in DatabaseCache. This is a related but distinct concern from the higher-confidence finding on this same code path about a known, real key_id being used to lock out its legitimate holder (CR-005): this one is about cache-row inflation from fabricated key_ids rather than denial of service against an existing integration.

> **Suggestion:** Consider rate-limiting on the raw credential/IP only until the key_id is confirmed to exist in the database, or cap the number of distinct key_id cache entries created per IP within a window.

> **Why it matters:** A distributed prober could inflate the DatabaseCache-backed cache with many short-lived rows per minute; low impact today, but worth a second look if this cache table is otherwise lightly used.

Evidence:
- src/api/auth_backends.py:254-260: `env, key_id, secret = parse_api_key(raw)` then `self._check_rate_limit(request, "key_id", key_id)` before `APIKey.objects.get(key_id=key_id)` is attempted.

<sub>Found by: spk-reviewer-backend-python · CR-022</sub>

src/api/models/base.py
```diff
@@ -214,3 +220,101 @@ class Meta:
...
+        key_id, secret_hash, raw = generate_api_key()
+        key = cls.objects.create(
+            profile=profile,
+            name=name,
+            key_id=key_id,
+            secret_hash=secret_hash,
+            expires_at=expires_at,
+            created_by=created_by,
+            updated_by=created_by,
+        )
```
gridcell: **Info** · `concurrency` · confidence: low

**key_id generation has no retry/backoff on a unique-constraint collision**

`generate_api_key()` draws a 12-character key_id from a 62-character alphabet (~3.2e21 possibilities), so a collision is astronomically unlikely at any realistic key volume. `issue()` still calls `cls.objects.create(...)` directly with no catch for `IntegrityError` on `key_id`, so on the vanishingly rare occasion of a collision (or an RNG defect), key issuance would raise an unhandled 500 instead of retrying with a fresh key_id.

> **Suggestion:** Wrap the create in a small retry loop (2-3 attempts) that catches `IntegrityError` on the unique key_id constraint and regenerates, purely as defense in depth.

> **Why it matters:** Low practical risk given the keyspace, but an unhandled IntegrityError surfaces as an opaque 500 to whoever is minting the key rather than transparently retrying.

Evidence:
- key_id, secret_hash, raw = generate_api_key()
- key = cls.objects.create(profile=profile, name=name, key_id=key_id, secret_hash=secret_hash, ...)

<sub>Found by: spk-reviewer-database · CR-023</sub>

gridcell:
@coderabbitai review

src/api/models/base.py
```diff
@@ -214,3 +225,141 @@ class Meta:
...
+    on the key rather than on who asks for one: `expires_at` defaults to a year
+    rather than to never, revocation is one call, and a key can never mint
+    another key. The self-service endpoint adds two ceilings of its own,
+    `settings.API_KEY_MAX_LIFETIME_DAYS` and `settings.API_KEY_MAX_PER_PROFILE`,
+    so one person cannot accumulate permanent credentials without limit.
+    """
+
+    profile = models.ForeignKey("Profile", related_name="api_keys", on_delete=models.CASCADE)
+    name = models.CharField(max_length=100)
+    key_id = models.CharField(max_length=12, unique=True)
```
saanobhaai: Is this redundant with the primary key?

PR #767, opened 2026-09-11
Serve the Beta classifier from the inference Lambda

gnieuwenhuis:
Points both environments' inference Lambda at the Beta classifier image (`v1-1`) and gives it the S3 access it needs. Nothing invokes the function — the API-side env vars and the `lambda:InvokeFunction` grant stay with the cutover PR that reads them.

- `InferenceSettings` gains a required `classifier_version` and rejects a pair that disagrees with `image_tag`; `image_bucket` is removed as wrong for prod and duplicative of `DjangoSettings.ic_bucket_name`.
- `InferenceStack` takes `(bucket, prefix)` pairs and grants read + put on every one. There's no staging indirection: with the bucket-policy grant below in place, every image bucket the function touches is read-write, so a per-pair access flag would have nothing left to encode.
- The config bucket stays read-only, and the IAM5 suppressions name the actions and resources they cover.
- A new alarm, `mermaid-{env}-inference-feature-store-errors` — see below.

## Writing to `coral-reef-training`

`coral-reef-training` is an [AWS Open Data Sponsorship Program](https://github.com/awslabs/open-data-registry/blob/main/datasets/coralreef-image-classification-training.yaml) bucket, hosted in a foreign AWS-owned account (`557690602013`). Its bucket policy is not ours to edit from this repo, so an identity-side `grant_put` on our execution role can't do it alone — CDK emits one half of a cross-account grant, and until the bucket side exists, the other half does nothing.

The bucket side now exists. A statement has been applied to `coral-reef-training`'s policy:

```json
{
  "Principal": { "AWS": "arn:aws:iam::554812291621:root" },
  "Action": "s3:PutObject",
  "Resource": "arn:aws:s3:::coral-reef-training/mermaid/*",
  "Condition": { "ArnEquals": { "lambda:SourceFunctionArn":
    "arn:aws:lambda:us-east-1:554812291621:function:prod-mermaid-inference-pyspacer" } }
}
```

Worth calling out:

- The account principal is scoped by the condition, not broad. `lambda:SourceFunctionArn` pins the **function**, so no IAM role had to be pinned and the execution role stays CDK-generated. Verified after the policy was applied: an `AdministratorAccess` role in our own account still gets `AccessDenied` writing there — only that one function can.
- A role ARN couldn't be used as the `Principal` — S3 requires a named principal to exist when the policy is written, and it resolves it to that role's unique ID, so a role CDK later replaced would silently stop matching. Conditioning on the function sidesteps both.
- The function ARN stays unqualified (no version/alias), and the **function name is load-bearing**: renaming it revokes prod's write with no deploy-time error. `inference.py` carries a comment recording this next to `function_name=`.

With that grant in place, the staging prefix, the ECS relocation hop, and the distinction between read-only and read-write buckets all go away — see below.

## What else this removes

- No staging prefix, no lifecycle rule, no relocation hop. `CommonStack` is back to zero diff against `dev`, and `ic_s3_path_staging` is gone from `DjangoSettings`.
- The `BucketAccess` enum is gone. It used to let a caller state read-only vs. read-write per `(bucket, prefix)` pair; now every pair gets both, so there's nothing left for it to distinguish.

## The new alarm

`data-mermaid/mermaid-inference` PR #6 makes a failed feature-vector write non-fatal: the function returns success with `feature_vector_output: null` instead of raising. That's invisible to both the Lambda `Errors` metric and the existing `ProcessingErrors` filter, and the vector is lost with nothing retrying it.

`FeatureStoreErrorsAlarm` is a log metric filter on the marker `[classify.feature_store_error]` in the Lambda's log group, threshold 5 in a 5-minute window, wired to the same alerts topic as the existing alarms.

`image_tag` deliberately stays `v1-1` in this PR. The alarm is inert until the `v1-2` image carrying mermaid-inference#6 is deployed by #768 — safe in the meantime, since `v1-1` still raises on a store failure and the existing `ProcessingErrors` alarm covers that gap.

## Deploying this

- Merging to `dev` redeploys `dev-mermaid-inference` onto `v1-1`. Nothing invokes it, so this is a free proof that the image pulls and the function boots.
- Prod repoints on its next release tag, also uninvoked — so the later cutover tag does not have to repoint the image and flip the API in one deploy.
- **Rollback.** Both functions currently run tag `v2-1`, which has no git tag behind it and is reproducible only from its digest `sha256:106017ea34e6a3b57d306b81cc1d760f6cf283679bda87feec3812555000a00f`. The target `v1-1` is `sha256:967e119f1484723cf5d5a2e71f1f9246feba403677409d97cb8987c6fcd96025`, confirmed present in ECR. The ECR lifecycle rule expires untagged images only.

## Verification

- `cdk synth` exits 0 — no unsuppressed cdk-nag findings.
- The prod role's put statements are exactly `coral-reef-training/mermaid/*` and `mermaid-image-processing/mermaid-production-test/*`; dev's is exactly `mermaid-image-processing/mermaid/*`.
- `quality-reviewer`: PASS, against the synthesized templates.

Not proven here: an end-to-end write through the Lambda's own role. Nothing invokes the function until #768, and `SimulatePrincipalPolicy` can't evaluate a resource policy's condition — the bucket-policy grant above is confirmed applied and confirmed scoped, not confirmed reachable by an actual invocation.

Closes https://github.com/data-mermaid/mermaid-classifier/issues/99

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added support for deploying classifier versions with matching image tags.
  * Inference services can now use multiple image storage locations with read/write access.
  * Added monitoring for repeated feature-store errors.
  * Deployment ordering now ensures inference services are available before dependent APIs.

* **Bug Fixes**
  * Improved validation to prevent mismatched classifier versions and image tags.

* **Documentation**
  * Updated the deployment runbook with classifier version configuration and validation guidance.


gnieuwenhuis:
[Serve the Beta Classifier from the inference Lambda](https://trello.com/c/69LA5VaF/2108-serve-the-beta-classifier-from-the-inference-lambda)

gnieuwenhuis:
[Run classification on the inference Lambda, delete the in-process path](https://trello.com/c/H1XRq5sw/2109-run-classification-on-the-inference-lambda-delete-the-in-process-path)

docs/deploy-classifier-version.md
ms280690, line 1: Is there a better way to do versionaing?

iac/settings/dev.py
ms280690, line 1: Is there a better way to do versionaing

iac/settings/prod.py
ms280690, line 1: Is there a better way to do versionaing

iac/settings/settings.py
```diff
@@ -56,33 +56,74 @@ class DjangoSettings:
...
-    image_bucket: str = "mermaid-image-processing"
     memory_mb: int = 10240
     timeout_minutes: int = 10
     ephemeral_storage_gb: int = 2
-    reserved_concurrency: int = 20
+    # A rolling deployment runs the image-worker ECS service at up to 200% of
+    # its task count with old tasks still draining, so the concurrent-invoke
+    # ceiling is double IMAGE_WORKER_MAX_TASKS, not the steady-state count.
+    # Reserved concurrency is subtracted from the account's unreserved pool.
+    reserved_concurrency: int = 2 * IMAGE_WORKER_MAX_TASKS
```
ms280690, line 97: why 2x?

iac/app.py
```diff
@@ -43,6 +43,9 @@
     "mermaid-api-infra-common",
     env=cdk_env,
     tags=tags,
+    # Unoverridden in both dev.py and prod.py, so either settings module names
+    # the same literal — this bucket and prefix are shared across envs.
+    staging_prefix=DEV_SETTINGS.api.ic_s3_path_staging,
```
ms280690: if they are shared between envs is the name `staging_prefix` the most appropriate

iac/settings/settings.py
```diff
@@ -56,33 +56,74 @@ class DjangoSettings:
     email_host: str = "smtp.gmail.com"
     email_port: str = "587"
     mc_user: str = "Mermaid"
+    # Mirrors IMAGE_S3_PATH in src/app/settings.py: the key prefix under
```
saanobhaai, line 59: It feels less than ideal to have it in two places

docs/deploy-classifier-version.md
gnieuwenhuis: There are a number of things we can do to harden versioning with the model versions, but I think they are out of scope for this PR. New issue created as a followup: https://github.com/data-mermaid/mermaid-api/issues/771

iac/settings/dev.py
gnieuwenhuis: see above comment

iac/settings/prod.py
gnieuwenhuis: see above comment

iac/app.py
gnieuwenhuis: this is changing - Kim let me know that we can add permissions to write directly to the prod training bucket - changes are coming

iac/settings/settings.py
gnieuwenhuis: The Lambda's only caller is the image-worker ECS service, and each worker task processes one job at a time, so concurrent invokes can never exceed the task count. The service caps at 3 tasks — but during a rolling deploy ECS runs the service at up to 200% (MaximumPercent: 200) while old tasks drain, so 6 tasks can briefly coexist. Reserved concurrency is a hard ceiling, so 3 would throttle image jobs on every deploy.

iac/nag_suppressions.py
```diff
@@ -746,32 +752,35 @@ def suppress_inference(stack: Stack) -> None:
...
-                "CDK-generated by grant_read; the Lambda needs read access to "
-                "all coral-reef patch images stored in this bucket.",
-                applies_to=[
-                    "Resource::<MermaidImageProcessingBackupBucket138A6358.Arn>/*",
-                ],
+                reason=f"{ACCEPTED}: the object wildcard on each image-bucket grant is "
+                "CDK-generated and confined to that bucket's own key prefix; every image "
+                "bucket is read-write, including the foreign coral-reef-training bucket, "
+                "whose bucket policy grants this function's exact name a matching put "
+                "statement.",
```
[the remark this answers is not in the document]
gnieuwenhuis: Leaving this as is — `grant_put` never emits an `s3:PutObject*` wildcard, so there is no IAM5 finding to suppress.

- In aws-cdk-lib 2.237.1 the put grant enumerates its actions. The synthesized statement is `s3:PutObject`, `s3:PutObjectLegalHold`, `s3:PutObjectRetention`, `s3:PutObjectTagging`, `s3:PutObjectVersionTagging`, `s3:Abort*` — `s3:Abort*` is the only wildcard, and it is already in `applies_to`.
- `cdk synth` exits 0 on this branch for both `dev-mermaid-inference` and `prod-mermaid-inference`, with no unsuppressed cdk-nag findings.
- Adding `Action::s3:PutObject*` would suppress a finding that is never raised.

iac/stacks/inference.py
```diff
@@ -78,13 +83,28 @@ def __init__(
...
         config_bucket.grant_read(self.function, "classifier/*")
-        image_bucket.grant_read(self.function)
+
+        # (bucket, key prefix) per env: the function reads source images and writes
+        # extracted feature vectors back under the same prefix, in every image
+        # bucket including the foreign coral-reef-training bucket, whose policy
+        # grants this function's exact name a matching put statement.
+        for bucket, prefix in image_buckets:
+            bucket.grant_read(self.function, f"{prefix}*")
+            bucket.grant_put(self.function, f"{prefix}*")
```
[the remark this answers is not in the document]
gnieuwenhuis: Leaving this as is — the in-account staging prefix was removed deliberately, and writing direct is what this PR now does.

- `coral-reef-training`'s bucket policy grants `s3:PutObject` on `mermaid/*` to this function by ARN (`lambda:SourceFunctionArn`), so the write lands without a staging hop or a later relocation step.
- Feature vectors belong beside the images they were extracted from, so read and write share one prefix by design; each grant is scoped to `{prefix}*`, not the whole bucket.

gnieuwenhuis:

CodeRabbit raised two findings on the IAM grants: an unsuppressed `s3:PutObject*` wildcard, and write access reaching the production source prefix. Both were pushed back on, none accepted, and no code changed. The first does not reproduce — `grant_put` enumerates its put actions rather than wildcarding them, so `s3:Abort*` is the only wildcard it emits and it is already suppressed; `cdk synth` exits 0 on this branch for both inference stacks. The second asks to restore the in-account staging prefix this branch removed on purpose, now that `coral-reef-training`'s bucket policy grants put to the function by ARN.

iac/app.py
gnieuwenhuis: changes done - this was removed.

ms280690 (approved):

iac/settings/settings.py
gnieuwenhuis: One place is in the app (django settings), and the other is in the iac code (here). The only reason it needs to be here is to scope permissions for the S3 path. Worst case scenario is the two end up drifting, in which case we'll get a clear AccessDenied at runtime, not a hidden problem that surfaces months later.

PR #768, opened 2026-09-12
Run classification on the inference Lambda and delete the in-process path

gnieuwenhuis:
Coral point-classification moves out of the API container and onto the already-deployed inference Lambda (ECR `v1-2`, serving the Beta model from its legacy pickle), and the in-process pyspacer path is deleted in the same change. There is no toggle and no fallback: merging to `dev` **is** the dev cutover.

- Adds `api.utils.inference`, a boto3 + pure-pydantic client for the function, and rewires `_classify_image` to it
- Deletes the in-process classify path, its S3 model fetch and `/tmp` cache, and drops `pyspacer` — with it torch, torchvision and scikit-learn leave the image
- Points each feature-vector write at the image's own bucket and prefix, so the Lambda writes its final key directly
- Makes `_write_classification_results` idempotent under a row lock, so SQS redelivery can no longer double an image's points and annotations
- Threads `profile_id` (previously accepted and dropped) and a per-job SQS visibility timeout through `classify_image_job`
- Surfaces `is_default` and a `?version=` filter on the classifier endpoint, so callers can tell which version is live
- Wires the container env vars, grants `lambda:InvokeFunction` to the image worker alone, and gives that worker an explicit log group so its failures can be alarmed
- Scopes the image-bucket grants to what each role actually reads, dropping three that the deleted in-process pyspacer path was the only caller of
- Moves both envs to ECR `v1-2`, the build that reports a failed feature-vector write instead of discarding the classification

## Stacked — review last

Built on #767 (the inference IaC), merged in as `a4f9ba34`. Those commits drop out of this diff once it lands. #728 — schema + `register()` + migration `0134` — has already merged.

## Feature vectors go straight to their final key

Prod's image bucket `coral-reef-training` is an AWS Open Data bucket in account `557690602013`. A statement on its bucket policy grants this stack's Lambda `s3:PutObject` on `mermaid/*`, matched by an `ArnEquals` condition on `lambda:SourceFunctionArn`, so the function writes each feature vector to its final key with its own execution role. No staging prefix, no relocation hop, and no contributing-org credentials inside the Lambda — which matters, because pyspacer caches one thread-local S3 resource from process-global config and cannot hold two credential identities at once.

Two things this makes load-bearing:

- **A policy in an account we do not control.** `grant_read` on that bucket was already load-bearing for reading images; the put grant now is too. Its identity half is CDK-generated here, but the resource half lives over there.
- **The function's name.** The condition matches `arn:aws:lambda:us-east-1:554812291621:function:prod-mermaid-inference-pyspacer` exactly, so renaming the function revokes prod's feature-vector writes with no deploy-time error — only a rising `FeatureStoreErrors` metric. `pyspacer_function_name` in `iac/settings/settings.py` is the single place that name is formed, and its docstring says so.

A failed write still does not fail the classification — the point predictions are the product — but it is no longer silent. `v1-2` answers with `feature_vector_output: null` and logs `[classify.feature_store_error]`; `feature_vector_file` is left unset rather than pointing at a key that holds nothing, and #767's `FeatureStoreErrors` filter alarms at 5 or more in 5 minutes. That covers a failure neither the Lambda `Errors` metric nor `ProcessingErrors` can see.

## Worth knowing while reviewing

- **The pinned lookup is a safety fix, not just plumbing.** `_resolve_active_classifier()` uses `.get(version=...)` rather than `Classifier.latest()`. Registering a newer row can no longer silently become the active classifier.
- **Re-classification overwrites instead of orphaning.** `FileField.save()` appends a suffix via `get_available_name()` when a key exists; `put_object` overwrites. First classification is byte-identical.
- **The idempotent writer cascades to confirmed annotations.** Accepted: redelivery happens minutes after upload, and the delete is what repairs a partially-written first attempt.
- **Local classification now needs a real Lambda** and images in S3. A direct consequence of deleting the fallback, not a bug.
- **Three numbers must agree** — the botocore retry ceiling, `INFERENCE_JOB_VISIBILITY_TIMEOUT`, and `image_sqs_message_visibility`. They derive from botocore's `total_max_attempts`, not from the Lambda's 600s timeout. Each site names the others.
- **The image worker's `[classify.processing_error]` alarm stays, renamed.** Its filter and metric were called `FeatureVectorRelocation*`, but the marker is still logged by `_classify_image` on a permanent failure and by `simpleq/workers.py` when extending a job's SQS visibility fails. Both are new in this PR, so renaming them to `ClassifyProcessingError*` loses no metric history.
- `docs/deploy-classifier-version.md` gains a `register_classifier` step. It works only for a version whose S3 prefix carries a `model.json`; `v1` has none, tracked as data-mermaid/mermaid-classifier#102.

## Carried, not fixed

- The image worker receives 10 SQS messages sharing one visibility clock, so a slow batch can redeliver its tail. Pre-existing; the per-image budget improves from ~6s to ~150s here, and the idempotent writer makes a duplicate delivery harmless.
- The image worker's old auto-generated log group is retained with no expiry and should be deleted by hand after rollout; in-flight worker logs can drop for the length of that rollout.

## Verification

- Full `api/tests`: **599 passed**
- Container image: `spacer`, `torch`, `sklearn` absent; `numpy`, `PIL`, `pandas`, `pyarrow` present
- `cdk synth`: exit 0 across all 11 stacks, zero unsuppressed cdk-nag findings
- `pre-commit`: all hooks pass
- `quality-reviewer`: earlier passes on the cutover, plus one on the direct-write change — its finding, that no test failed if the echo comparison were loosened to a null check, is closed by `1025f1fe`

Closes data-mermaid/mermaid-classifier#54

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **New Features**
  - Classification now uses the configured inference service and classifier version.
  - Classifier listings identify the active version and support filtering by version.
  - Classification results avoid duplicating points during retries and preserve reviewed points.
  - Added monitoring for classification processing errors.

- **Bug Fixes**
  - Improved handling of inference failures, timeouts, and retryable errors.
  - Queue workers continue processing remaining jobs when visibility updates fail.
  - Classification jobs now use configurable point counts and visibility timeouts.
  - Missing inference configuration is now reported before processing starts.

- **Documentation**
  - Added deployment guidance for registering and validating classifier versions.


docs/deploy-classifier-version.md
```diff
@@ -82,13 +82,53 @@ It builds the `pyspacer-function` Lambda image (baking in
...
+`INFERENCE_CLASSIFIER_VERSION` on the API service and both workers (`ApiStack`),
+so bumping it redeploys `dev-mermaid-api-django` / `prod-mermaid-api-django`
+(new task definition revisions) alongside `dev-mermaid-inference` /
+`prod-mermaid-inference` — expect both stacks in the same deploy.
+
+> InferenceStack deploys before ApiStack, so a bump serves the new `vN` from
+> the Lambda before the API/worker tasks roll to match. Classifications
+> handled by a task still on `INFERENCE_CLASSIFIER_VERSION=vN-1` during that
+> window fail the classifier-version drift check — nothing is lost, the
+> message redelivers automatically about 25 minutes later. Drain or pause the
```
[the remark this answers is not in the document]
gnieuwenhuis: Fixed in 4bb7481f — the aside names the ceiling now instead of stopping at the first redelivery.

- Four deliveries over about 75 minutes, then the DLQ at about 100: SQS moves the message on the receive that would exceed `max_receive_count`, not on the last delivery. That figure is a floor, since visibility is extended after receive.
- Seven-day retention there and no automatic return, so the step asks for the rollout to finish inside that window or the DLQ to be redriven once it has.

src/api/utils/inference.py
```diff
@@ -0,0 +1,342 @@
...
+                extra={"traceparent": traceparent, "error_code": envelope.error_code.value},
+            )
+            raise InferenceError(
+                envelope.message,
+                retryable=envelope.retryable,
+                error_code=envelope.error_code.value,
+            )
+
+        try:
+            response = parse_classify_response(payload)
```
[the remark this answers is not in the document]
gnieuwenhuis: Fixed in 2c076141 — `classify_via_lambda` rejects `valid_rowcol=False`, and compares the requested and returned coordinates as multisets before it returns anything.

- Both raise a non-retryable `InferenceError`. A producer that answers wrongly answers the same way on redelivery, so this fails once and alarms rather than cycling to the DLQ.
- The requested multiset is built from the request payload actually sent rather than re-derived from `points`, so it cannot diverge from the coordinates the Lambda received, and the log line carries a bounded sample of the missing and unexpected coordinates — the raised message stays short because it lands in a user-visible status field.
- Neither shape is reachable from the image deployed today: the handler builds `point_results` by zipping the requested points with the scores, and pyspacer raises on a `valid_rowcol=False` features object before a response exists. The contract permits both, and the write deletes every existing `Point` first, which is what makes it worth guarding.

src/api/utils/inference.py
```diff
@@ -0,0 +1,342 @@
...
+            # bump has a redelivery window before both sides agree; a genuine permanent
+            # mismatch retries to the DLQ and alarms instead of failing silently.
+            raise InferenceError(
+                f"Classifier version drift: Lambda served {response.classifier_version!r}, "
+                f"expected {expected!r}",
+                retryable=True,
+            )
+
+        installed = CONTRACT_VERSION
+        if response.contract_version and response.contract_version != installed:
```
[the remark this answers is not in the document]
gnieuwenhuis: Fixed in 2c076141 — the comparison is direct now, so a response carrying no stamp no longer passes the pin. `test_classify_via_lambda_missing_contract_version_tolerated` is `..._raises` and expects `InferenceError`.

- Nothing deployed relied on the tolerance: every image tag in mermaid-inference — `v1-1`, `v1-2`, `v2-2`, `v2-3` — stamps `contract_version` on all three response paths, and both environments run `v1-2`.
- The `ErrorEnvelope` branch above keeps its truthiness test. It refines the diagnosis on an envelope that already failed validation, where an absent version should stay "malformed envelope" rather than becoming a contract mismatch.

src/simpleq/workers.py
```diff
@@ -47,7 +47,12 @@ def work(self, burst=False, wait_seconds=5):
             for queue in self.queues:
                 for job in queue.jobs:
                     if job.visibility_timeout:
-                        queue.extend_job_visibility(job, job.visibility_timeout)
+                        try:
+                            queue.extend_job_visibility(job, job.visibility_timeout)
+                        except Exception as e:
+                            logger.exception(
+                                f"[classify.processing_error] failed to extend visibility for job {job}, continuing without it: {e}"
+                            )
```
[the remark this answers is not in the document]
gnieuwenhuis: Fixed in aaa90334 — `Worker.work` continues to the next job after logging, so a job whose lease could not be extended never reaches `job.run()`.

- It skips `remove_job` with it, which is the point: the work was not done, so the message stays in flight and redelivers rather than being deleted.
- The test asserts the job neither ran nor was removed, alongside a job with no `visibility_timeout` that still runs in the same batch — a skip that also stopped the batch would be the worse of the two bugs and nothing else would tell them apart.

gnieuwenhuis:

Four comments, all four taken. The deploy doc's redelivery note carries its DLQ ceiling; `classify_via_lambda` rejects `valid_rowcol=False` and any response whose coordinates are not the ones requested; its contract-version check no longer accepts a missing stamp; and the queue worker skips a job whose visibility lease could not be extended rather than running it without one. e0352ffa...aaa90334 — full `api/tests` 602 passed, pre-commit clean.

docs/deploy-classifier-version.md
```diff
@@ -104,12 +149,13 @@ environment has its own settings file and its own stack
 
    Merging to `dev` triggers **[Deploy CDK](https://github.com/data-mermaid/mermaid-api/actions/workflows/deploy-cdk.yml)**,
    which updates `dev-mermaid-inference`'s `PyspacerInferenceFunction` to serve
-   the new image. Validate on dev.
+   the new image, and rolls `dev-mermaid-api-django`'s API/worker tasks to the
+   matching `INFERENCE_CLASSIFIER_VERSION`. Validate on dev.
 
 2. **Prod.** Make the same edit in
    [`iac/settings/prod.py`](../iac/settings/prod.py), merge it to `dev`, then cut
```
saanobhaai: I would make the same edit in both dev and prod settings files, then merge/deploy and test on dev, then merge to master / deploy to prod.

src/api/utils/classification.py
```diff
@@ -443,56 +409,47 @@ def _classify_image(image_record_id, profile_id=None):
...
-        score_sets = response_message.scores
-        _write_classification_results(image, score_sets, label_ids, classifer_record, profile)
-
-        with open(tmp_feat_vector_file_path, "rb") as tmp_feat_vector_file:
-            image.feature_vector_file.save(
-                f"{image.id}_featurevector", tmp_feat_vector_file, save=True
-            )
+        classifier_record = Classifier.active()
+        points = generate_points(image, num_points or settings.INFERENCE_DEFAULT_NUM_POINTS)
+        result = classify_via_lambda(image, points)
```
saanobhaai, line 407: Synchronous, up-to-~600s inference Lambda is invoked before _write_classification_results checks whether the image's points are already reviewed/confirmed.

Failure scenario: SQS redelivers a classify_image_job for an image that already completed successfully. The redelivered job re-invokes classify_via_lambda in full — a costly synchronous Lambda call plus an S3 feature-vector overwrite — only for _write_classification_results to silently discard the result once it finds the image already has reviewed points.

src/api/checks.py
```diff
@@ -0,0 +1,39 @@
...
+    errors = []
+    if not settings.INFERENCE_LAMBDA_PYSPACER:
+        errors.append(
+            Error(
+                "INFERENCE_LAMBDA_PYSPACER is not set.",
+                hint="Set INFERENCE_LAMBDA_PYSPACER to the pyspacer Lambda's function name.",
+                id="api.E001",
+            )
+        )
+    if not settings.INFERENCE_CLASSIFIER_VERSION:
```
saanobhaai, line 39: These checks verify only that the settings are non-empty, but not that a `Classifier` db row actually exists for that version (which would happen if we forget to run `register_classifier()` for that version).

requirements.txt
```diff
@@ -15,6 +15,7 @@ drf-nested-routers==0.95.0
 drf-recaptcha==4.0.3
 gunicorn==26.0.0
 mailchimp3==3.0.21
+mermaid-inference-contract @ git+https://github.com/data-mermaid/mermaid-inference.git@v0.5.0#subdirectory=packages/mermaid-inference-contract
```
saanobhaai, line 18: Shouldn't we make this release-level before referencing here?

src/api/models/classification.py
```diff
@@ -213,6 +225,24 @@ def clean(self):
     def latest(cls):
```
saanobhaai: Still used by  signals/classification.py -- and as far as I can tell, only there. Should this function be removed entirely?

src/api/tests/test_classify_lane_worker.py
```diff
@@ -0,0 +1,279 @@
...
+from api.models import Annotation, ClassificationStatus, Classifier, Image, Point
+from api.utils import classification, q
+from api.utils.classification import _classify_image, classify_image_job
+from api.utils.inference import InferenceError, LambdaClassificationResult
+
+PINNED = {"INFERENCE_CLASSIFIER_VERSION": "v2"}
+THRESHOLDS = {"CLASSIFIED_THRESHOLD": 0.5, "AUTOCONFIRM_THRESHOLD": 1.0}
+
+
+@pytest.fixture
```
saanobhaai: Duplicate fixtures not consolidated

Local image and classifier_v2 pytest fixtures duplicate the shared fixtures already added in src/api/tests/fixtures/classification.py, which are auto-imported for every test.

Failure scenario: A companion commit in this same branch (9ddb482f, "Drop tests that only an intentional decision can break") moved these exact duplicated fixtures into the shared fixtures module across several other test files but missed this one; a future change to the shared fixture (e.g. a required field) can be applied to the shared copy while this file's local copy silently drifts out of sync.

src/api/utils/classification.py
```diff
@@ -443,56 +409,47 @@ def _classify_image(image_record_id, profile_id=None):
...
+        _write_classification_results(image, result.point_predictions, classifier_record, profile)
+
+        if result.feature_vector_name:
+            # A queryset update records exactly the key the Lambda wrote: no re-upload,
+            # no get_available_name suffix, and no post_save re-checksum from S3.
+            Image.objects.filter(pk=image.pk).update(feature_vector_file=result.feature_vector_name)
+            image.feature_vector_file.name = result.feature_vector_name
```
saanobhaai: This assignment isn't doing anything, because the image isn't saved here or in `create_classification_status`.

src/api/utils/inference.py
```diff
@@ -0,0 +1,367 @@
...
+
+    Raises InferenceError on an ErrorEnvelope payload, a malformed envelope or
+    response, a classifier-version drift mismatch, a contract-version mismatch, a
+    response with valid_rowcol False, or a response whose points do not match the
+    requested coordinates. A feature vector the Lambda did not write where requested
+    does not fail the job: points and annotations are the product, and nothing in
+    src/ reads feature-vector bytes back.
+    """
+    tracer = otel_trace.get_tracer("api.inference")
+    with tracer.start_as_current_span("pyspacer.classify_via_lambda"):
```
saanobhaai, line 273: Duplicated contract-version-mismatch handling

Contract-version mismatch is checked and reported via two separate, near-identical logger.error + raise InferenceError blocks in classify_via_lambda — one for the malformed-error-envelope branch, one for the success-response branch.

Failure scenario: Both blocks guard the same root cause (contract version drift between independently-deployed services) with the same message shape; a future change to the message or mismatch condition must be updated in two places, risking silent divergence.

src/api/resources/classification/classifier.py
```diff
@@ -34,20 +36,45 @@ class Meta:
...
+            "is_default",
             "created_on",
             "created_by",
             "updated_on",
             "updated_by",
         ]
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._active_pk_cache = {}
```
saanobhaai: As far as I can tell this dict never has more than one entry, so could be a simple str/uuid no?

src/api/resources/classification/image.py
```diff
@@ -286,7 +286,12 @@ def create(self, request, *args, **kwargs):
 
         if trigger_classification:
             create_classification_status(image_record, status=ClassificationStatus.PENDING)
-            classify_image_job(image_record.pk)
+            # num_points is left unset here; classify_image_job falls back to the
+            # deployment-wide point count in effect when the job runs.
+            classify_image_job(
+                image_record.pk,
+                profile_id=profile.pk,
```
saanobhaai: Let's trace through and revisit the need for profile_id here -- I couldn't immediately see why this is being added.

src/api/utils/inference.py
```diff
@@ -0,0 +1,367 @@
...
+        f"pyspacer inference: {context} ({err.error_count()} validation error(s))"
+    )
+
+
+class LambdaClassificationResult(NamedTuple):
+    point_predictions: list
+    feature_vector_name: str | None
+
+
+def get_lambda_client():
```
saanobhaai, line 121: Global mutable state for the client cache — this already needed a workaround in this PR's own tests (`test_inference_invoke.py` has to `monkeypatch.setattr(inference, "_lambda_client", None)` before it can inject a fake client, otherwise state leaks across tests). `project.py`'s `_get_client` solves the same "cache a boto3 client" problem with an instance-scoped dict instead of a module global — no `global` keyword, no cross-test leakage. Consider `functools.lru_cache(maxsize=None)` here instead: same one-per-process memoization, thread-safe out of the box, and tests can reset it with `get_lambda_client.cache_clear()` rather than reaching into a "private" module var.

src/api/utils/inference.py
```diff
@@ -0,0 +1,367 @@
...
+        )
+        raw = response["Payload"].read()
+        if response.get("FunctionError"):
+            function_error = response["FunctionError"]
+            detail = raw.decode("utf-8", errors="replace")[:500]
+            raise InferenceError(
+                f"Lambda FunctionError ({function_error}): {detail}",
+                retryable=_function_error_is_retryable(function_error, detail),
+            )
+        return json.loads(raw)
```
saanobhaai: `json.loads(raw)` on the success path isn't covered by the `except (BotoCoreError, ClientError)` below it — a non-FunctionError response whose payload isn't valid JSON (truncated body, a proxy mangling it, a bad manual invoke) raises `json.JSONDecodeError` (a `ValueError`), which isn't wrapped into `InferenceError` and propagates raw. The outer `except Exception` in `_classify_image` still catches it and marks the image FAILED, but `retryable` defaults to `False` via `getattr(err, "retryable", False)`, and the stored status message is a bare `"Expecting value: ..."` instead of the informative messages every other failure path here produces. `Classifier.register()` already widens its except tuple to include `ValueError` for the identical reason (catching `json.JSONDecodeError` from `read_json_object`) — worth the same guard here for consistency. No test currently exercises a non-JSON success payload either.

saanobhaai (changes requested):

src/api/checks.py
gnieuwenhuis: It is possible to check the db in this verification, which would ensure that lamda inferences are set up for  success at worker startup, the big disadvantage I see is that it also ties deployment logistics with classifier registration. If the new classifier doesn't exist in the database before the classifier is updated through a deployment, the worker will fail to start, and CDK will roll the deployment back. This has consequences to running things locally too.

If we keep it as is, deployment will work, and images that get queued while waiting for classifier registration will fail (not automatically retryable). In this case an inference alarm will be triggered (meaning it is not an invisible error), and worker startup and deployment are not blocked.

requirements.txt
gnieuwenhuis: Do you mean we should release the mermaid-inference repo at >= 1.0.0? I guess I don't really have a concern with that if so, but given that it is still new and relatively untested I think keeping the version tag below 1.0.0 is also justifiable. Keeping it below 1.0.0 doesn't really block anything here, it's more of an admission about the maturity of the code in that repo (at least that's how I would read it - I don't feel very strongly about this though).

src/api/checks.py
gnieuwenhuis: One note here - I made some changes based on your later comment about removing `Classifier.latest()` which forces database compliance in this check by registering new models if they are not already registered in the db. This solves the problem you are noting here, as well as fixing potential problems at deploy time when the model doesn't exist in the db, but the environment has been updated with a new classifier version.

src/api/models/classification.py
gnieuwenhuis: Yes, I looked closer and I agree. This did make clearer some confusing semantics around when the active model is updated at deploy time, and I decided that it cleans things up a lot by registering the model automatically in the worker (similar to a migration I guess) to avoid differences between what is deployed and what is in teh database (which is part of why `latest()` was needed in that post save function). https://github.com/data-mermaid/mermaid-api/pull/768/commits/61cd4d4bc236fb309543b3ac871a32ae732cd27d

I also created a follow up issue for this, to make sure deploy failures do not break these semantics either: https://github.com/data-mermaid/mermaid-api/issues/772

src/api/tests/test_classify_lane_worker.py
gnieuwenhuis: Change made - copied the fixture into the shared fixtures: https://github.com/data-mermaid/mermaid-api/pull/768/commits/f4fef92058e50b5886c1750aab31ef099bd50b94

src/api/utils/classification.py
gnieuwenhuis: yep, copied over and missed - deleted that line: https://github.com/data-mermaid/mermaid-api/pull/768/commits/09ef3df85d8587c2503dfea4dadacb601728a202

src/api/utils/inference.py
gnieuwenhuis: Updated in https://github.com/data-mermaid/mermaid-api/pull/768/commits/3723611625aa57c0f50f1655ec690afaeb437a03. 

The message was deduplicated, but the condition stays separate on purpose - seperating the difference between a named version mismatch (raises) versus a missing version mismatch due to an older lambda (logs).

src/api/resources/classification/classifier.py
gnieuwenhuis: It has to tell 'not looked up yet' apart from 'looked up, no active classifier' (None), so a bare attribute would query again on every row when nothing is configured. I've switched it to a cached_property, which handles that without the dict. https://github.com/data-mermaid/mermaid-api/pull/768/commits/bcac80a968a7e5c9e1fad8a5954d408f3cb0e667

src/api/resources/classification/image.py
gnieuwenhuis: It was iniitally meant for allowing attribution on machine output - being able to say "uploaded by" for example. But it is never used, and it is both simpler as well as matches dev to remove it - done in: https://github.com/data-mermaid/mermaid-api/pull/768/commits/679935eedaf61e31b14478648cb90b4342f7138f

src/api/utils/inference.py
gnieuwenhuis: Agreed, switched to functools.cache, and the tests reset it with cache_clear() in a fixture that runs before and after each test. One note: the monkeypatch was restoring the old value, so nothing leaked. It was resetting the cache so Session gets called, which cache_clear() still has to do. https://github.com/data-mermaid/mermaid-api/pull/768/commits/7bc1561352790950e7c8faccf184eee0a3e490d2

src/api/utils/inference.py
gnieuwenhuis: Agreed — fixed. json.loads is now wrapped in its own narrow except ValueError (covers JSONDecodeError and UnicodeDecodeError) and raises a non-retryable InferenceError with a descriptive message. I kept it out of the boto except, since that clause derives a boto error code and also covers the json.dumps of the request. Added a parametrized test for a truncated body and invalid UTF-8. https://github.com/data-mermaid/mermaid-api/pull/768/commits/79b1a127434309cc478f3918d5049e8937a41f7e

docs/deploy-classifier-version.md
gnieuwenhuis: Done in https://github.com/data-mermaid/mermaid-api/pull/768/commits/e89297afecafd3af0aa421187095dfd6c429bea3

src/api/utils/classification.py
gnieuwenhuis: This is one I pondered for a bit, but I think it's safe to push back, largely because the 600s timeline is significantly overshooting the actual timeline. Most inferences should toak 2-4 seconds, regular cold starts might end up causing a single 10second inference once in a while, and at worst, the image gets dropped from the AWS fleet cache, resulting in a single 90s call on the front end (not repeated).

Also, this failure mode should be rare, and since remaking the inference isn't a big deal, I feel ok leaving this as is.

saanobhaai:
[Run classification on the inference Lambda, delete the in-process path](https://trello.com/c/H1XRq5sw/2109-run-classification-on-the-inference-lambda-delete-the-in-process-path)

saanobhaai:
@coderabbitai review

gridcell (approved):

saanobhaai (approved):

